PR #84, opened 2013-06-11
Relative Due Date for User Task

balsarori:
http://forums.activiti.org/content/time-duration-support-user-task-due-date


nivertius:
I know that this issue is more than two years old, but added line 31 should read:
`return new DateTime(clockReader.getCurrentTime()).plus(Period.parse(duedate)).toDate();`
Otherwise, relative due date is untestable with custom Clock. This bug is still present in master.

jbarrez:
Correct. Fixed : https://github.com/Activiti/Activiti/commit/e2f5d6c

PR #285, opened 2014-06-05
Allow 'public' processes to be selected when an authorised user is set.

brett-walker:


tijsrademakers:
This could be done as part of a native query. This implementation is not generic enough right now to merge it to master. Kept it open for a while (way too long) because I understand the use case, but a native query is probably a better solution.

PR #322, opened 2014-07-15
Task candidate and assignee

henryyan:


z772458549:
@henryyan How can I do it same with spring boot.

Best Wishes!

PR #346, opened 2014-07-31
Additional TaskQuery options

balsarori:
Added the following with test methods
taskAssigned() to select tasks that are assigned
taskOwned() and taskUnowned() to select tasks have or don't have an owner 
taskInvolvedGroup() and taskInvolvedGroupIn() to select tasks that a group or groups are involved in
taskInvolvementType() to select tasks based on identityLink type

cache candidate user groups


tijsrademakers:
Closing this because of complex SQL

PR #357, opened 2014-08-15
Generalise abstract layers

w32blaster:
Within this changeset we extracted new interface for ProjectEngineConfiguration. Now, all the abstract layers are communicated through an interface, but not by calling a particular class.

There is no new functionality within this pull-request. Thus, we didn't touch behaviour of the whole code at all. This changeset allows us to implement our own Process Engine Configurator and use our own persistance layer.  We require this feature in order to continue work on [our fork of activiti-neo4j](https://github.com/p-i/activiti-neo4j).


tijsrademakers:
Closing it, because of no response. Please reopen if there's more information.

PR #401, opened 2014-10-07
Oracle short string support (<4000 chars)

gro-mar:
http://forums.activiti.org/content/bug-oracle-ddls


fandji:
If I look into the DDLs of MSSQL, there is a nvarchar(4000):

create table ACT_RU_VARIABLE (
    ID_ nvarchar(64) not null,
    REV_ int,
    TYPE_ nvarchar(255) not null,
    NAME_ nvarchar(255) not null,
    EXECUTION_ID_ nvarchar(64),
    PROC_INST_ID_ nvarchar(64),
    TASK_ID_ nvarchar(64),
    BYTEARRAY_ID_ nvarchar(64),
    DOUBLE_ double precision,
    LONG_ numeric(19,0),
    TEXT_ nvarchar(4000),
    TEXT2_ nvarchar(4000),
    primary key (ID_)
);

and according to this documentation:

http://msdn.microsoft.com/de-de/library/ms186939.aspx

this means that you can put in 4000 Characters into such a column.
Which correspons to the initialization of the Variabletypes:
protected void initVariableTypes() {
    ---
      variableTypes.addType(new StringType(4000));
      variableTypes.addType(new LongStringType(4001));
   ---
  }

The remark, that in oracle a nvarchar(2000) means 4000 bytes is correct, so long as the encoding is AL16UTF16.
But that doesn't mean you can put in 4000 Characters, only 2000 characters. So on oracle string variables between 2001 and 4000 characters don't function.
As I'm unable to find out where the VariableInstanceEntity is really populated, I would propose that the LongString and String should be adjustet to 2000 and 2001.
As nvarchar2 on oracle is limited to 4000 bytes, safest would be a max of 1000 Characters, as Unicode Codepoints can need up to, need (UTF-32) 4 bytes.

martin-grofcik:
Hi fandji,

In case of oracle DB you can configure org.activiti.engine.impl.cfg.ProcessEngineConfigurationImpl#customPreVariableTypes.

I would prefer to change from nvarchar(4000) to nvarchar(2000) for the all DBs and in that case we could set LongStringType to 2001 but it is not possible without migration procedure (there could be data already in the column...)

Martin

fandji:
Hi Martin

How about something like this(Can't use git, srry)?
      /**
       * http://jira.codehaus.org/browse/ACT-2129
       */
      if (databaseType.equals("oracle")) {
        variableTypes.addType(new StringType(2000));
        variableTypes.addType(new LongStringType(2001));
      } else {
        variableTypes.addType(new StringType(4000));
        variableTypes.addType(new LongStringType(4001));
      }
I think this should be the only change, as the Historic thingies just copy the VariableEntityInstance. Changing the ddl's isn't possible for Oracle because it is already at max and shrinking a column on other databases is surely not viable.

fandji

PR #414, opened 2014-10-25
ReflectUtil.loadClass() and custom classloader

oefimov:
http://forums.activiti.org/content/reflectutilloadclass-and-custom-classloader


cyndi5:
The Activiti forums link no longer works. Here is the archived version: https://web.archive.org/web/20150920101714/http://forums.activiti.org/content/reflectutilloadclass-and-custom-classloader

PR #456, opened 2014-12-05
Fix missing StartNodeEvent and optimizing text displayed in workflow boxes 

laurents59:
Missing default case for start event node in if / else if statements
Computed BoxHeight for text is too small (height - 16 - 5 - 5 - 12 - 2 - 2) ==> deleting -16 to allow one more line to be drawn
On multilined text, use fixTextToWidth when text couldn't be fully displayed
On line text : use original if it already fits in original width (without -10 pixels for "..." statement).


tijsrademakers:
We've merged the none start event change, but  for the text change we need a good set of unit tests to show it's working in all cases.

PR #490, opened 2015-01-06
fix ACT-2066 add index on ACT_HI_TASKINST(PROC_INST_ID_)

smirzai:
Fixed ACT-2066.


tijsrademakers:
Hi Saeid, this patch will only create this index for new database because there are no upgrade scripts included. Is that with a reason?

smirzai:
Hi Tijs, 
Actually it is a performance improvement index, and might not be that critical.
I can add upgrade scripts if you think it is good to have them.

tijsrademakers:
Ok, yes if we add it to the create scripts then it should also be added to the upgrade scripts. If you could add the upgrade scripts, that would be great. Would be good to validate if this index doesn't cause an issue on DB2, since for standalone tasks the process instance id will be NULL.

smirzai:
Hi Tijs
I guess I should create upgrade new upgrade files.
For example fot mysql:
activiti.mysql.upgradestep.51702.to.51703.history.sql

Or shall I put it to existing
activiti.mysql.upgradestep.51701.to.51702.history.sql


I am not sure if I understand naming convention.

Saeid

tijsrademakers:
Please create a new one, activiti.mysql.upgradestep.51702.to.51800
I'll take care of the rest that's needed, if you can provide the upgrade script files.

smirzai:
Hi Tijs,
Done
Would you please check and merge ?

Saeid

tijsrademakers:
Thanks Saeid

PR #498, opened 2015-01-19
make context dynamic

smirzai:
Make context dynamic


tijsrademakers:
This 'generic' solution is no improvement to my opinion. Changing the context root property when necessary is a better solution for now.

PR #507, opened 2015-02-01
Batch fetching jpa-entity-list variables

eugenehr:


tijsrademakers:
Would it be possible to also add some unit tests to the pull request? Thanks in advance.

eugenehr:
Yes, i'm working on it

tijsrademakers:
Thanks for adding the test. When I run this test against the current Activiti master it already runs green. So this test doesn't show the reason for adding the changed logic of this pull request. Could you create a test showing cases where the current Activiti master fails? If that's not possible, could you explain the reasoning behind this pull request?

eugenehr:
I don't know how can I write such kind of  test. My pull request is fully covered with existing unit tests. The purpose of my pull request is just to reduce the amount of database read operations. For example:

    variables.put("entities", Arrays.asList(entity1, entity2, entity3, entity1000));
    runtimeService.startProcessInstanceByKey(.., variables);
    taskService.createTaskQuery().includeProcessVariables().singleResult(); 

Task query will perform a single read operation from the database instead of 1000 for every entity in the list.

Also I think that existing JPAVariableTest is not correct. It uses entities after only entityManager.persist() operation, but not after merge(), find() and createQuery().
In case with OpenJPA EntityManager after persist entities, flush changes into the database and select them back again, we will have enhanced versions of entities and, as a result, derived classes without @Entity and @Id annotations.
This will lead to the fact that JPAEntityMappings class will always return **false** from function isJPAEntity().

For example just add followed lines at the end of setupJPAVariables() method:

    manager = entityManagerFactory.createEntityManager();
    manager.getTransaction().begin();

    // just reload persisted entity
    simpleEntityFieldAccess = manager.find(simpleEntityFieldAccess.getClass(),
        simpleEntityFieldAccess.getId());

    manager.flush();
    manager.getTransaction().commit();
    manager.close();

    entitiesInitialized = true;

I think this simpleEntityFieldAccess variable won't be stored as **jpa-entity** process variable because it will be enhanced without @Entity and @Id annotations.
Also after that modifications some tests will fail on ReflectUtil.loadClass() method with:
**java.lang.ClassNotFoundException: org.apache.openjpa.enhance.org$activiti$standalone$jpa$FieldAccessJPAEntity$pcsubclass**

tijsrademakers:
Ok just to be clear on this. This patch will reduce the amount of database read operations a lot in case of JPA entity lists, right?

eugenehr:
I have to close this pull request for a while because it depends on correct storing of enhanced entity class names.
:disappointed:

PR #512, opened 2015-02-04
Updated com.oracle.ojdbc dependencies

jphuynh:


tijsrademakers:
Which version of Oracle are you testing with?

jphuynh:
The version is : 11.2.0.1.0

tijsrademakers:
Ok thanks, we are using that version as well with the current ojdbc configured in the pom.xml. Could that be solved in Bamboo instead? So adding the same jar to the group and artifact id we have defined in our pom.xml?

jphuynh:
I've seen with Samuel. We'll update the Maven repository to be compliant with your pom.xml.
Could you just send me the ojdbc jar file that you are using so we can make sure we use the same ?

tijsrademakers:
Ok thanks, it's fine to use the ojdbc jar you are using now, because the tests went ok.

PR #514, opened 2015-02-07
Support for enhanced jpa-entity and jpa-entity-list variables

eugenehr:


tijsrademakers:
Manually merged, thanks

PR #518, opened 2015-02-13
Added a method (processInstanceIdIn) in TaskInfoQuery

m4ng0:
See https://jira.codehaus.org/browse/ACT-2196.


jbarrez:
Thanks! Good addition to the query indeed!

PR #520, opened 2015-02-17
 activiti:failedJobRetryErrorCode, added

smirzai:
<activiti:failedJobRetryErrorCode> now makes the job executor to throw specified BPMN error after all tries are exhauseted

### DIFF
--- a/modules/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/constants/BpmnXMLConstants.java
+++ b/modules/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/constants/BpmnXMLConstants.java
@@ -284,4 +284,5 @@ public interface BpmnXMLConstants {
   public static final String ELEMENT_FORMAL_EXPRESSION = "formalExpression";
   
   public static final String FAILED_JOB_RETRY_TIME_CYCLE = "failedJobRetryTimeCycle";
+  public static final String FAILED_JOB_RETRY_EXCEPTION = "failedJobRetryErrorCode";
 }
--- a/modules/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/BaseBpmnXMLConverter.java
+++ b/modules/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/BaseBpmnXMLConverter.java
@@ -24,6 +24,7 @@ import org.activiti.bpmn.constants.BpmnXMLConstants;
 import org.activiti.bpmn.converter.child.BaseChildElementParser;
 import org.activiti.bpmn.converter.export.ActivitiListenerExport;
 import org.activiti.bpmn.converter.export.FailedJobRetryCountExport;
+import org.activiti.bpmn.converter.export.FailedJobRetryErrorCodeExport;
 import org.activiti.bpmn.converter.export.MultiInstanceExport;
 import org.activiti.bpmn.converter.util.BpmnXMLUtil;
 import org.activiti.bpmn.model.Activity;
@@ -204,6 +205,7 @@ public abstract class BaseBpmnXMLConverter implements BpmnXMLConstants {
     if (baseElement instanceof Activity) {
     	final Activity activity = (Activity) baseElement;
         FailedJobRetryCountExport.writeFailedJobRetryCount(activity, xtw);
+        FailedJobRetryErrorCodeExport.writeFailedJobException(activity, xtw);
         
      }
     
--- a/modules/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/child/ActivitiFailedJobErrorCodeParser.java
+++ b/modules/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/child/ActivitiFailedJobErrorCodeParser.java
@@ -1,0 +1,43 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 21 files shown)


tijsrademakers:
Hi Saeid, looks like a good addition to the Activiti Engine. Two questions:

1. I see you changed AtomicOperationTransitionDestroyScope, can you elaborate on this change?
2. Can you create a test with cancelActivity set to false and true?

Thanks!

smirzai:
Hi Tijs
Thank you for looking into the pull request.

1. Actually because the way I am calling the exception part, the scope is
created from scratch. So the get parent will be null. I assume it is safe,
as I am using the execution itself instead of its parent only if the parent
is null. In that case, it will throw npe any way. So one can be sure, there
will be no side effect.

2. sure. I will soon.

Regards
Saeid

smirzai:
Hi Tijs
I am creating the  testcase for cancelActivity. But I assume
ErrorBoundaryEvent by default stops the activitiy and actually
cancelActivity is ignored. is that the right thing to test ?

Regards
Saeid

tijsrademakers:
No, error boundary event uses the same BoundaryEventActivityBehavior to my knowledge. About my question 1, we don't think this is the right approach. No change on this level should be needed to get this working. To proof this would still work ok you would need to add a parallel gateway and nested sub process and multi instance to the testing as well. Because we expect some issues for these kind of processes.

smirzai:
Oh, I am surprised to hear that. It does not make much sense to continue a
process after exception. Just consider Java as an analogy.
It is strictly stated in standard, See table 10.91 "Boundary event
attributes", page 258 of official BPMN spec
http://www.omg.org/spec/BPMN/2.0/PDF
you can see it in comunda as well:
http://docs.camunda.org/latest/api-references/bpmn20/

I think it is worth a second check and maybe a patch. What do you think ?

Saeid

jbarrez:
An error event can indeed only be interrupting. However, in this specific use case for a custom extension it does mean the job is lost here.

>  It does not make much sense to continue a process after exception

How do you mean. Isn't that exactly what we're doing here: handling a failure and for example routing to an email to a sys admin or whatever.

smirzai:
it is related to our discussion above. Tijs proposes to repeat the test case, with cancelActivitiy=false. I argue that there should be no change in behavior and actually cancelActivitiy should be ignored by error boundary event. As far as I understand, Tijs believes Error boundary event uses the same BoundaryEventActivityBehavior  and is sensitive to cancelActivity.

smirzai:
Hi Tijs
Yeap, you are right. Now there is a test case to show the problem
JobExecutorFailRetryTest.testFailedServiceTaskExceptionInParallelBranch


Any guideline on how it should be implemented properly ?

tijsrademakers:
Hi Saeid,

Removing the change in AtomicOperationTransitionDestroyScope is a start. For implementing a new feature a change in this class should not be needed and results in problems with parallel behaviour like you experienced. So my advise would be to work on a solution where this change is not needed first.

smirzai:
Hi Tijs
I try

smirzai:
Hi Tijs, That part is clear to me.
But I am not sure what is the conclusion on cancelActivitiy. What should be
the behavior?

tijsrademakers:
Hi Saeid, 

For this extension it makes sense to have cancelActivity to use true and false. Because you might not want to cancel the job. I know this is not according to the spec, but this is an extension so it can behave differently.

smirzai:
Hi Tijs,
How cancelActivity works in a normal Error boundary exception ?
That part is not an extension and is better to comply with standard.
And if there cancelActiyity is neglected for error boundary events, it is
reasonable for this extension to behave similarly.

tijsrademakers:
Let's first get it working with cancelling the activity with support for parallel, multi instance and nest sub processes and we can decide on this later on.

smirzai:
Ok, I changed the implementation. Now AtomicOperationTransitionDestroyScope remains untouched. It passes the tests, including the parallel multi instance one.
If you can suggest any other complex test which may fail, I will try to implement it.
This way is more natural and I guess I should work for most complex cases.

tijsrademakers:
Hi Saeid,

I have a question about the following code (AsyncContinuationJobHandler):

if (job.getRetries() == 1) 
     execution.setAsyncRetryFailErrorCode(execution.getActivity().getFailedJobRetryErrorCode());

So the retry fail error code is only set when the retries value is 1?
Can you explain the reason for implementing it like this?

Also in JobRetryCmd:

if ((job.getRetries() == 0) && (activity.getFailedJobRetryErrorCode() != null)) {
+            // errorProgation is already executed, just delete the job
+            job.delete();
+        }

Maybe we could define this as an additional attribute, but by default we don't want to delete the job, because that would break the default behaviour.

smirzai:
Hi Tijs,

Let me explain it. We need to coordinate between asny Continuation which
knows how many retries are done and when the error should be propagated and
classDelegate which actually propagates. This is done
via asyncRetryFailErrorCode field of execution.
It is checked in ClassDelegate#139:
      } catch (Exception e) {
        String asyncFailRetryErrorCode = ((ExecutionEntity)
execution).getAsyncRetryFailErrorCode();
        // this was last try, if an errorcode is defined, follow it
        if (StringUtils.isNotEmpty(asyncFailRetryErrorCode))
          ErrorPropagation.propagateError(asyncFailRetryErrorCode,
execution);
        else
            throw e;


If it is set, it means that Async Continuation has signaled for
propagation. If it is not set, class delegate continues as usual. If it is
set, it means, if there is a failure, propogate to BPMN error is defined.

Is it clearer now ?


About deletion of job, I guess if the BPMN exception is caught, it means it
is dealt with and nothing remains for manual intervention. Just consider
the Java analogy. If you catch exception in any level, then you have dealt
with it. Otherwise system will catch it and show it on console.
To me it looks more natural.
If you are not still convinced, just let me know to add another option for
it.

Regards
Saeid

tijsrademakers:
Closing this PR, because it's not mergeable in this state. Please reopen when there's time to make changes and let's discuss them.

PR #521, opened 2015-02-18
ACT-2179 Date coerced to String should be in ISO8601 format

stephraleigh:
Pull request for ACT-2179. Updated the TypeConverterImpl so that a Date coerced into String will be in ISO8601 format.


jbarrez:
Thanks, nice addition!

I updated your test case slightly to not need the Thread.sleep.

PR #525, opened 2015-02-23
Jpa batch fetching

eugenehr:
Reduce multiple database reads when querying tasks JPA variables


tijsrademakers:
Can't merge the PR in this state, because it now always loads all variables for a task:

// ensure that all jpa-variables are loaded
 +    for (Task task : instanceList) {
 +      task.getProcessVariables();
 +    }

PR #527, opened 2015-02-26
Namespace Issue on Import/Export & Suspended Process Definition does not have effect on subprocesses.

vasiledirla:


tijsrademakers:
Thanks Vasile

PR #528, opened 2015-02-26
Map exception

smirzai:
This patch will add the exception mapping. 
one can have constructs like 
        <activiti:mapException  errorCode="myErrorCode1" withChildren="true">org.activiti.engine.test.bpmn.event.error.mapError.BoundaryErrorParentException</activiti:mapException>

if includeChildExceptions is set, the children of the mentioned class will also be routed.
The first map with no class mentioned, regardless of its includeChildExceptions will match every exception


tijsrademakers:
Thanks Saeid

jbarrez:
@smirzai : would you be able to add a small section of documentation for this?

smirzai:
Yes Joram. Just at the moment, I am trying to expand it for some other tasks, and writing some documentation. 
Stay tuned.

smirzai:
@jbarrez 
Done in https://github.com/Activiti/Activiti/pull/534. Waiting to be merged.

PR #542, opened 2015-03-11
Act 2135 Select property and header items from camel to activiti using regex

smirzai:
With this patch it is possible to select which camel properties or headers should be copied as activiti variables.
copyVariablesFromProperties and copyVariablesFromHeader can be true or false as before. True means all items and false means none
Now it is possible also to use simple or complex regex expressions. Expressions like
?copyVariablesFromHeader=(property1|property2)

expression can contain wildcards and can be as complex as required.

There is also a bug fix in this patch. If copyVariablesFromProperties were set to true, then camel body was not copied. There is a test case to show the problem and its fix


tijsrademakers:
Hi Saeid, I think this is a good addition to the Camel module.
Can you remove the changes in the camel-activiti-parent-context.xml file and remove the new BooleanStringUtils class (should be a method at most). Then I'll make sure it gets merged.

smirzai:
Done.

PR #547, opened 2015-03-16
MyBatis Multi-db vendor support + other simple tweaks

balsarori:
See 
http://forums.activiti.org/content/mybatis-multi-db-vendor-support


balsarori:
Will create a new PR for Activiti V6

PR #563, opened 2015-04-02
TerminateEndEvent extension elements

martin-grofcik:
added possibility to have EndEvent definition in TerminateEndEvent


tijsrademakers:
Hi Martin, What's the use case for having the end event in the behaviour class?

martin-grofcik:
Hi Tijs, 
Let's say we have terminateEndEvent in the process. Process instance terminates on this event. The fired PROCESS_CANCELLED event contains activity (but this activiti does not contain any extension element from definition) and handler should decide - based on the extension elements

Regards
Martin

PR #564, opened 2015-04-02
UserGuide: Fixed Call Activity super process example image

kberger:
Call Activity super process image for credit check example was not displaying correctly, nor was it linking to the correct image.  

Corrected link syntax and found correct image file.


jbarrez:
Indeed, thanks for fixing this!

PR #566, opened 2015-04-06
Fix for ACT-2236

gdharley:


tijsrademakers:
Hi Greg, I don't understand the change in BoundaryEventJsonConverter. The PROPERTY_CANCEL_ACTIVITY property is already set, so what does this change?

gdharley:
Hi Tijs,

The property is not set when you import a bpmn file into Activiti Explorer.
Release is 5.17.
Create a BPMN file in Eclipse modeler
Import into Activiti Explorer
Deploy immediately

The  CancelActivitiy flag is not promoted.
The change explicitly sets the flag.

tijsrademakers:
Ah ok, this was already fixed at the end of February, so it should already work on master.

gdharley:
Thanks Tijs, I missed that commit (would have saved me a lot of time :-) 
I am also chasing a namespace issue where the target namespace is overwritten (means reports cannot be created). Is this also already fixed?

PR #568, opened 2015-04-09
Update BpmnXMLUtil.java

bokyung:
So that callers of the parseChildElements() method can override the behavior of a child element parser. It is not possible with the current implementation because childParsers.putAll(genericChildParserMap) overwrites child parsers specified in childParsers with the same element name of any parser in genericChildParserMap. Also removes the unwanted side-effect of filling childParsers with all genericChildParserMap.


bokyung:
@Activiti:master, any comment/feedback/progress on this pull request? How do I proceed with the request?

PR #573, opened 2015-04-21
ACT-2173: Add support of import relative to the BPMN description

cdeneux:
see https://activiti.atlassian.net/browse/ACT-2173


cdeneux:
Caution: if the pull request [585](https://github.com/Activiti/Activiti/pull/585) is merged previously, the changes of the commit c76890d must be also applied.

PR #577, opened 2015-04-28
Oracle nvarchar2 >varchar2

martin-grofcik:
I would like to propose move from NVARCHAR2 to VARCHAR2 for oracle DB.
Oracle does not recommend to use these datatypes. There are restrictions in processing and performance is decreased.


tijsrademakers:
Hi Martin, did you do a test with a large dataset? These kind of changes are often killing for large datasets.

tijsrademakers:
Closing this for now, please reopen if works ok for large datasets.

PR #578, opened 2015-04-28
Conditional BoundaryTimer dueDate - based on NULL duration expression support.

skay2k:
This allows schedule *or not* a BoundaryTimer Job, depending on dueDate execution variable value.
Null value  expression (dueDate or ISO duration) is supported; if NULL - BoundaryTimer ignored.


skay2k:
Hi,
I had this feature to be bundled in our product.
First, a NULL dueDate/duration is throwing an Exception at runtime, so this is very annoying.
Secondly, we need to go back to our human task, and we need to check if timer has already expired once / or not. THe only way to do this is either a messy 'hack delegate/listener' code OR propose this simple pull request :)
best regards

jbarrez:
See http://forums.activiti.org/content/578-conditional-boundarytimer

skay2k:
Whats the status of the request? Should I add something?

tijsrademakers:
Sorry for the delay, we have been focusing on Activiti 6 and the Activiti community day of 10 June, so we are a bit behind in reviewing and merging pull request. We'll make sure to take time for this right after 10 June.

skay2k:
See you tomorrow 
Br
Envoyé de mon iPhone

PR #580, opened 2015-04-30
Fix read stencilset.json gibberish when in other languages

chenzhiguo:
Fix read stencilset.json gibberish when in other languages. (UTF-8)


kotslon:
I have had the same problem in my own application (i don't use Activiti REST, just Engine), and implemented the same solution. Don't you need to add the same changes in [ModelEditorJsonRestResource](https://github.com/Activiti/Activiti/blob/master/modules/activiti-modeler/src/main/java/org/activiti/rest/editor/model/ModelEditorJsonRestResource.java#L45) ?

```java
// ...
@RequestMapping(value="/model/{modelId}/json", method = RequestMethod.GET, produces = "application/json;charset=utf-8")
// ...
```

PR #581, opened 2015-04-30
Update LDAPUserManager.java

kgiannakakis:
- Remove '%' from fullNameLike. These characters would be inserted by Activiti Explorer (or similar tool), but they do not work with LDAP searching
- Avoid null pointer exception in case of a user with no email


jbarrez:
See http://forums.activiti.org/content/user-form-type-does-not-show-users-when-using-ldap

Issue #497, opened 2015-01-17
java.lang.NullPointerException when delete user picture


mengqhai:
The javadoc in IdentityService says that you can delete a user's picture by invoking setUserPicture(userId, null):
  /** Sets the picture for a given user.
   * @throws ActivitiObjectNotFoundException if the user doesn't exist.
   * @param picture can be null to delete the picture. */
  void setUserPicture(String userId, Picture picture);

But it's seems not true for 5.16.4, I actually got NullPointerException after doing that:
java.lang.NullPointerException
	at org.activiti.engine.impl.persistence.entity.UserEntity.setPicture(UserEntity.java:79)
	at org.activiti.engine.impl.persistence.entity.UserEntityManager.setUserPicture(UserEntityManager.java:168)
	at org.activiti.engine.impl.cmd.SetUserPictureCmd.execute(SetUserPictureCmd.java:51)
	at org.activiti.engine.impl.interceptor.CommandInvoker.execute(CommandInvoker.java:24)
	at org.activiti.engine.impl.interceptor.CommandContextInterceptor.execute(CommandContextInterceptor.java:57)
	at org.activiti.spring.SpringTransactionInterceptor$1.doInTransaction(SpringTransactionInterceptor.java:47)
	at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:133)
	at org.activiti.spring.SpringTransactionInterceptor.execute(SpringTransactionInterceptor.java:45)
	at org.activiti.engine.impl.interceptor.LogInterceptor.execute(LogInterceptor.java:31)
	at org.activiti.engine.impl.cfg.CommandExecutorImpl.execute(CommandExecutorImpl.java:40)
	at org.activiti.engine.impl.cfg.CommandExecutorImpl.execute(CommandExecutorImpl.java:35)
	at org.activiti.engine.impl.IdentityServiceImpl.setUserPicture(IdentityServiceImpl.java:107)

PR #584, opened 2015-05-05
Fixing java.lang.NullPointerException when attempting to delete user picture

agornovo:
This fixes https://github.com/Activiti/Activiti/issues/497


jbarrez:
Why the close?

agornovo:
Have discovered what may be a related bug, so will resubmit once I fix both.

PR #591, opened 2015-05-20
Json variable type

tstephen:
This pull request uses a heuristic to spot strings that are in fact JSON and then parses such variables as a javax.json.JsonStructures. 

This allows expressions to refer to child elements in conditions, when passing data to call activities etc. For example in a parent process that has a JSON variable named 'contact':  
```
    <callActivity calledElement="FollowUpContact" completionQuantity="1" id="_2" isForCompensation="false" name="Follow Up" startQuantity="1">
      <extensionElements>
        <activiti:in sourceExpression="${contact.getString('contactId')}" target="contactId"/>
        <activiti:in sourceExpression="${contact.getString('tenantId')}" target="tenantId"/>
      </extensionElements>
    </callActivity>
```

### DIFF
--- a/modules/activiti-engine/pom.xml
+++ b/modules/activiti-engine/pom.xml
@@ -154,7 +154,15 @@
       <version>2.1.3</version>
       <scope>test</scope>
     </dependency>
-    
+
+    <!-- Required for native JSON ssupport -->    
+    <dependency>
+      <groupId>org.glassfish</groupId>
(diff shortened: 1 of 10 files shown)


tijsrademakers:
Thanks Tim, we have merged the idea of this pull request into master. It's revised a bit, so would be good if you can have a look if it still matches your ideas.

PR #594, opened 2015-05-22
fix async jobs execution

martin-grofcik:


tijsrademakers:
Hi Martin, do you have a unit test that fails without this change?

martin-grofcik:
It is not possible to create jUnit test for it. The behaviour is exactly the same.

What I wanted to fix by this commit is (whenever it is possible) we should follow the same patterns as were used in DefaultJobExecutor implementation.
**AcquireJobsRunnableImpl** - 
1. fetches all runnable jobs by command AcquireJobsCmd.
2. executes job list from the fetch.

Current implementation of **AcquireAsyncJobsDueRunnable** 
1. fetches all jobs but does not executes them. They are already executed in the command.

I proposed the pull request to keep the same pattern in new async job executor impl.
(These two approaches should be synchronized.)

PR #599, opened 2015-06-04
Use OR instead of AND for test

dbmalkovsky:
The methods, `setBitOn()` and `setBitOff()`, are only used in the `org.activiti.engine.impl.util.BitMaskUtil.java` file in the `isBitOn()` method.  

The `isBitOn()` method is only used in the `getCachedEntityState()` method in `org.activiti.engine.impl.persistence.entity.ExecutionEntity.java` file.

And the `getCachedEntityState()` method doesn't seem to have any references to it at all in the code base.

So instead of making the change for `setBitOn()` and `setBitOff()`, those two methods could be deleted along with `isBitOn()` in `BitMaskUtil.java`.  Then `getCachedEntityState()` would need to be deleted in `ExecutionEntity.java`.

This pull request doesn't make the above change but makes the smallest change to fix the code assuming that it is acutally used.


jbarrez:
I'm going to leave the bitmask util in for a little longer, such that we don't forget about the concept. The original idea was an optimisation we still want to do in v6: keeping track of what 'dependencies' (tasks, jobs, ect) an execution has to optimize the number of deletes on cleanup. The main problem is that updating that data can lead to optimistic locking exceptions real easily.

Not sure if we want to go with a bitmap though in the future. It looks cool, but it's a pain to debug.

PR #601, opened 2015-06-04
Refactor ProcessInstanceUtil & StartProcessInstanceCmd for extensibility

roberthafner:
Our custom extensions (i.e attributes on process variables) require us to perform the following steps:
1) create a process instance
2) set variables 
3) set attributes (Hook needed for custom extensions)
3) start process.

This pull request refactors a few classes to allow us to extend the StartProcessInstanceCmd and still re-use the functionality provided in the ProcessInstanceUtil class. A similar PR was submitted to version 5 but was not brought forward with the Activiti 6 refactoring.

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/StartProcessInstanceCmd.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/StartProcessInstanceCmd.java
@@ -93,11 +93,15 @@ public class StartProcessInstanceCmd<T> implements Command<ProcessInstance>, Ser
       throw new ActivitiIllegalArgumentException("processDefinitionKey and processDefinitionId are null");
     }
 
-    ProcessInstance processInstance = ProcessInstanceUtil.createAndStartProcessInstance(processDefinition, businessKey, processInstanceName, variables);
+    ProcessInstance processInstance = createAndStartProcessInstance(processDefinition, businessKey, processInstanceName, variables);
 
     return processInstance;
   }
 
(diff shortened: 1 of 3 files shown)


jbarrez:
Makes sense, merged.

@roberthafner : I was wondering how you extend the StartProcessInstanceCmd and inject it in the RuntimeService in Actviiti v5 now? Do you extend the whole RuntimeServiceImpl to have your custom command?

One idea we're playing with is to keep a mapping of commands in the engine, which can be overridden on process engine config level. There would be no need for service extension at that point anymore then. Would something like that work for you? Or do you see another approach that is better?

roberthafner:
We do not extend the RuntimeServiceImpl class (or any Activiti *ServiceImpl class). Rather our service auto wires an instance of the ManagementService and invokes the  command using the executeCommand method.

In the past, I too thought it would be nice to have the ability to override an existing Activiti Command class with a custom one (similar to how the Activiti Behavior classes work). But now I don't think its sufficient for us in many cases. For example, we extend Activiti's StartProcessInstanceCmd, CompleteTaskCmd and SetExecutionVariablesCmd classes to support setting DataObjects in the engine when invoked. A DataObject in our implementation is a process variable with attributes (i.e Map<String,String>). Therefore simply replacing the Activiti command with a custom one is insufficient for us as we still need to pass in the attributes and persist them. 

FWIW, I'm not opposed to such a change as I could see that it might be beneficial for other use cases. It just doesn't seem helpful in our use case.

jbarrez:
Ok, that is clear and makes sense, thanks!

PR #606, opened 2015-06-05
[Fix] Resolve foreign key violation on Identity Link table 

roberthafner:
when process is started with an authenticated user and ends via a Terminate End Event.

A foreign key violation occurs when a process is started with an authenticated user and then is terminated via a Terminate End Event. The code path current fails to clean up the associated "starter" identity link.


tijsrademakers:
Merged manually, thanks

PR #609, opened 2015-06-08
[Fix] - DestroyScopeOperation does not remove variables in the scope …

roberthafner:
…being destroyed.

Problem:
Variables defined on a subprocess are not removed when the subprocess scope ends. 

Fix: 
Update DestroyScopeOperation to remove local variables from the execution scope being destroyed.


tijsrademakers:
Merged manually, thanks

PR #615, opened 2015-06-09
Subprocess start triggers Process execution listeners to run

roberthafner:
ContinueProcessOperation triggers execution listeners defined on the process to run when starting a subprocess. 

Fix: Update condition to check the currentFlowNode for a subprocess. 

Note: This pull request is dependent on my earlier pull request (https://github.com/Activiti/Activiti/commit/7330f77d87d9d41ef84f9a6713f1d00b34e7cca1) .


roberthafner:
Cancelling this pull request as I left a System.out.println() call in the test. I will submit a new pull request without it.

PR #616, opened 2015-06-09
[Fix] Subprocess start triggers Process execution listeners to run

roberthafner:
ContinueProcessOperation needs to check whether the currentFlowNode is defined in a subprocess and not trigger process execution listeners if it does.

Note, this pull request depends on one of my earlier pull requests.


tijsrademakers:
Merged manually, thanks

PR #617, opened 2015-06-10
Update references in readme.html to use new JIRA location.

dbmalkovsky:
Now that `codehaus.org` is no more, update the JIRA references from: `http://jira.codehaus.org` to `https://activiti.atlassian.net`.

There is still one `codehous.org` reference as I could not find the location of the Wiki.

### DIFF
--- a/distro/src/readme.html
+++ b/distro/src/readme.html
@@ -1,301 +1,301 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
-  <head>
+<head>
     <title>Activiti Readme</title>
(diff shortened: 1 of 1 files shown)


tijsrademakers:
The diff of this pull request is not showing the changes. Is it possible that you changed indentation? Is it possible to create a pull request showing the changes you made in the files changed overview?

dbmalkovsky:
I don't understand what is going on here because when select  "Files changed" above I see, "2,808 additions, 2,808 deletions not shown" and if I click on View I can see that references like https://activiti.atlassian.net/browse/ACT-1527 instead of the the old codehaus links.

Maybe there are too many changes or it doesn't like the html page?

I'm not sure; sorry.

PR #618, opened 2015-06-10
support for CancelEndEvent in angularJS modeler

esahekmat:
add TerminateEndEvent(EndCancelEvent) to the designer


tijsrademakers:
Thanks for the pull request, but I see some confusion between the cancel and terminate end event. What you implemented in the modeler is a cancel end event and not a terminate end event. The doc changes are also not valid, because it's correctly referring to cancel end events.

esahekmat:
are terminate end event and cancel end event seperate elements?

esahekmat:
I revert doc changes, and also correct my mistake, now it's just implementing cancel end event

esahekmat:
did you see my revise?

PR #621, opened 2015-06-15
exclude jcl over slf4j

martin-grofcik:


jbarrez:
@martin-grofcik : what's the reasoning for this? I believe this was needed for Spring. 
With jcl excluded, wouldn't this mean the initial bootup logs will be gone, too?

martin-grofcik:
Library should not have runtime dependency on jcl-over-slf4j jar.
I can add jcl-over-slf4j to spring examples - next commits.

tijsrademakers:
Hi Martin, we would like to keep the dependency in, because Spring still uses JCL.

martin-grofcik:
Hi Tijs,

I run mvn dependency:tree in the activiti-spring module. I found only test jcl-over-slf4j dependency there.

tijsrademakers:
Hi Martin, true but Spring still uses JCL. If there's a Log4j dependency that will be used for the JCL logging. So the SLF4j dependency is just there to make sure all our logging using SLF4j.

PR #626, opened 2015-06-18
Updated the JaCoCo Maven plugin to fix the Sonar build

jphuynh:
This fixes the build on Bamboo that is failing from build number 67


jbarrez:
What are the implications of this? Does this impact the normal buildcycle of maven and on others building Activiti?

jphuynh:
No it doesn't impact the normal buildcycle. It just points Maven to the proper JaCoCo maven plugin version. That will fix the [Activiti Community build on Bamboo] (https://bamboo.alfresco.com/bamboo/browse/ACT-ACTCOM).

jbarrez:
That url is Alfresco VPN only, right? (many people NOT from Alfresco build Activiti too, so they cannot acces this).
Also, what's the purpose of this build on Bamboo? Is it just a sanity check, seems to be doing an H2 test only? Cause we've got all builds running on http://ec2-54-217-5-246.eu-west-1.compute.amazonaws.com:8080/jenkins/

jphuynh:
Yes this fix is only to solve an internal use case and *will not* affect the build from the community anyway. The only reason why this scenario exist is that at one point I was asked to replicate all the builds from Jenkins to Bamboo and Unit Tests coverage was also added.

jphuynh:
Closing PR. Not needed for Community version.

PR #632, opened 2015-07-01
ACT-1999: BpmnError not managed on TaskActivityBehavior signal method

pmesmeur:
Throwing an exception from a signal() method

http://forums.activiti.org/comment/30149#comment-30149

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/ServiceTaskDelegateExpressionActivityBehavior.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/ServiceTaskDelegateExpressionActivityBehavior.java
@@ -55,8 +55,12 @@ public class ServiceTaskDelegateExpressionActivityBehavior extends TaskActivityB
   public void signal(ActivityExecution execution, String signalName, Object signalData) throws Exception {
     Object delegate = expression.getValue(execution);
     if( delegate instanceof SignallableActivityBehavior){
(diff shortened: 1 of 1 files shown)


tijsrademakers:
Hi, We've thought about this pull request quite a bit, and there's definitely a use case where you would like to throw a BpmnError in a custom signal method implementation. We however think, that it's not something we should add to the default behaviour implementation. It's already possible now to implement the logic when needed, there's only some code required. If you disagree, please let me know some use cases why you think it should be part of the default behaviour logic. Thanks.

pmesmeur:
For me it is not a strong problem if you do not integrate my pull-request as I have the possibility to overload the default behavior. That's what I am currently doing and I am fully satisiffied

For memory, the origin origin of the pull request was on the forum: http://forums.activiti.org/comment/30149#comment-30149: I need to be able to forward to activiti an exception that occurs in a service task executed in its own thread. Joram asked me for a pull request and I thought that it was a good idea in order other developers can take a benefit of it (e.g. the one that opened the discussion on the forum)

tijsrademakers:
Yes thanks for the effort for creating the pull request. After some more thought Joram and I think it's a good example on how to solve this use case, but it's not something we should include in the behaviour class by default.

pmesmeur:
OK.
I can understand that you don't want add this new behavior. However, I would appreciate if you could - at least - include in your code the function that I named propagateError(). It would be reusable by the ones that, such as me, wants to overload the default signal. It would avoid code duplication.

Thank you

PR #640, opened 2015-07-17
Improved logging at gateway transition

aegliv:
Having a workflow transitioning different paths after certain conditions are met is one of the most powerful features of any workflow engine. 
Unfortunately, Activiti's execution engine currently does not reveal any information about why a certain path was chosen. This might not be of much concern to savvy BPMN model experts, but a developer could save quite some time debugging erogenous gateway decisions (caused for example by wrong types passed to ``formVariables``) if Activiti would log them.

This small PR opens the previous "black box" of gateway execution and offers more insight for developers. I hope the community follows my reasoning and accepts my PR.


tijsrademakers:
Maybe we could have a engine configuration that allows more logging information when executing a process instance, but it should not be the default.

PR #651, opened 2015-08-05
Supports customize annotationFontName as same as labelFontName

quaff:
Default annotationFontName "Arial" can't display Chinese or Japanese or
Korean, Activiti should be able to customize annotationFontName like
labelFontName.


quaff:
It should be applied to 5.x too.

tijsrademakers:
That was done as well:

https://github.com/Activiti/Activiti/commit/2e04e8d197f408f9a052fac1bf7e512bfb6ad5e5

PR #670, opened 2015-09-08
Fix lane titles

davidmoser:
On MacOS the titles of lanes and pools are mangled:
image

It seems related to this issue:
http://stackoverflow.com/questions/14569475/java-rotated-text-has-reversed-characters-sequence

We tested the fix on MacOS and Windows, but please retest to be sure.


reva:
Can no one confirm this?

tijsrademakers:
I can't reproduce this. Closing it for now, but please reopen when there's a way to reproduce it on Windows or Mac

PR #682, opened 2015-10-03
This update concerns the image generator. Now unspecified tasks are r…

phalix:
…epresented as tasks with no symbol. Previously they were represented like a manual task.

Also I reindented BpmnXMLConverter.java, this was a mess.
Actual Changes are in 

TaskXMLConverter.java
Task.java


tijsrademakers:
Unspecified tasks are not supported by Activiti. The indentation should be ok already (2 spaces).

PR #685, opened 2015-10-07
Prototype for localization support of task name and description.

roberthafner:
Hi Tijs,

Here is a prototype of how I was thinking localization support for task name and description could be added to the engine using property files. I think this approach could be extended to other elements as well. 

(This is a work in progress so don't merge).

thanks,
Rob


tijsrademakers:
Implemented the dynamic BPMN feature with localisation support inspired by this pull request.

PR #689, opened 2015-10-12
fire process and task cancelled events for sub processes and tasks in…

martin-grofcik:
When process instance is deleted, process cancelled event is not fired for sub processes and ENTITY_DELETED event is not thrown for the tasks.


tijsrademakers:
Merged manually, thanks

PR #696, opened 2015-10-21
Add a ClassDelegateFactory interface.

caleblevine-google:
This change adds a ClassDelegateFactory interface along with an implementation that calls new ClassDelegate(...). DefaultActivityBehaviorFactory and TaskListenerFactory are updated to accept a ClassDelegateFactory and use it instead of calling new directly. By default they are constructed with the DefaultClassDelegateFactory so no additional configuration is required.

Also factored out getSkipExpressionFromServiceTask in DefaultActivityBehaviorFactory.

This change should be completely safe. It doesn't change any existing interfaces.


tijsrademakers:
I was missing the usage of the ClassDelegateFactory in the DefaultActivityBehaviorFactory class so I've added it. Let me know if you think it should be done in another way.

caleblevine-google:
Oops. No, that's perfect, thanks!

PR #697, opened 2015-10-22
Weblogic 12c -solve problem with ProcessEngineLookup 

charlesheidemann:
Activiti-cdi not work on Weblogic 12c because Weblogic ClassLoader on CDI Extension don´t localize META-INF/services necessary to load ProcessEngine.
Change on ActivitiExtension for not call ClassLoader and call ProgrammaticBeanLookup for find implementation of ProcessEngineLookup.
It´s work fine on time of deploy, but not localize activiti.cfg.xml because again ClassLoader not access resources at this time.
For create ProcessEngine i add a ServletContextListener and call ProgrammaticBeanLookup.lookup(ProcessEngineLookup.class).getProcessEngine();
At this time all resources are disponible by ClassLoader.
If need create another ProcessEngineLookup eg for TestCase, we have that extend LocalProcessEngineLookup and put @Specializes annotation.
This work well for me on Weblogic 12c

### DIFF
--- a/modules/activiti-cdi/pom.xml
+++ b/modules/activiti-cdi/pom.xml
@@ -10,8 +10,7 @@
   <parent>
     <groupId>org.activiti</groupId>
     <artifactId>activiti-root</artifactId>
-    <relativePath>../..</relativePath>
-    <version>5.18.1-SNAPSHOT</version>
+    <version>5.18.0</version>
   </parent>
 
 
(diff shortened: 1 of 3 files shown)


tijsrademakers:
This doesn't look like a general CDI change, but a Weblogic specific one. We would need additional logic to differentiate between Weblogic and other application servers.

PR #698, opened 2015-10-23
enum values in editor and converter: id and name instead of single value

kotslon:
Related thread on activiti forum: https://forums.activiti.org/content/loss-id-when-converting-bpmntojson


tijsrademakers:
Duplicate of #703

PR #702, opened 2015-10-26
Add support for localization fallback via ResourceBundle.Control.

roberthafner:
Tijs, Let me know if you would like me to create another pull request for this change on the activiti6 branch.


tijsrademakers:
Hi Rob,

This is a good addition to the localization feature, but I think it should be optional logic. Because you might not want to use Locale strings for some reason. So maybe we could add a withLocalizationFallback boolean to the different queries that support localization? By default, the fallback logic is not executed, but if you use the withLocalizationFallback option it will use it. WDYT?

roberthafner:
Seems reasonable. I'll update the code and submit another pull request. Thanks!

roberthafner:
Closing as we do not have a need for this functionality in version 5.

PR #703, opened 2015-10-27
fix enum values

kotslon:
1. Fix enum values in editor and converter:
    * `id` and `name` instead of single `value`.
2. Fix form properties test:
    * check the size of form properties list
    * use json instead of string for `formproperties` in test fixture file `test.formpropertiesmodel.json`
    * `id` and `name` of `enumValues` are changed to be different.

[related thread on activiti forum](https://forums.activiti.org/content/loss-id-when-converting-bpmntojson)


tijsrademakers:
Merged manually with some UI changes. Thanks for the PR.

PR #705, opened 2015-10-29
resolve use mybatis startup error

bubuhui:
add IbatisVariableTypeHandler to handler VariableType, some mapper.xml use VariableType , but don't use IbatisVariableTypeHandler  to mapper this type. So if you use mybatis, you'll startup with error.


bubuhui:
sorry, I make a mistake, this is right, we don't need configure the activiti conf in my mybatis-configure.xml, activiti will init org/activiti/db/mapping/mappings.xml after the project start up.

Issue #708, opened 2015-10-30
Activiti can't store string variables with length >2000 and <4000 characters in oracle db


vasiledirla:
https://activiti.atlassian.net/browse/ACT-2129

tijsrademakers:
Fixed

PR #711, opened 2015-11-02
Dev vasile 708 - (V6) - can't store string variables with length >2000 and <4000 characters in oracle db

vasiledirla:
fix for: #708


tijsrademakers:
Already merged

PR #716, opened 2015-11-04
Fixes ACT-3997 "Validate bpmn components for max length"

erikwinlof:


erikwinlof:
Activiti 5 engine tests now added

PR #720, opened 2015-11-05
Dev-rpopa-15062 - Amazon RDS 5.6 is not using anymore the defaults for timestamp

rpopa:
Fix for MNT-15062; modified columns timestamp default to null for ACT_RU_TASK and ACT_RE_DEPLOYMENT for mysql dialect


tijsrademakers:
Hi, why are you submitting this to the Activiti project? Making changes to old upgrade script files won't work.

rpopa:
Hi Tijs, In my case I am expecting that only for clean installation this scripts to be executed; for upgrade case I have a script for alter table with modify columns in alfresco side. Should this alter table script be moved in activiti project?

tijsrademakers:
The change to the create scripts are fine, but we can only do this if the alter table logic is added to the latest upgrade script (right now this is 5.18.0.0.to.5.18.0.1). I can also make this change if that's easier.

rpopa:
Yes please! 
<code>
alter table ACT_RE_DEPLOYMENT 
   MODIFY DEPLOY_TIME_ timestamp null;

alter table ACT_RU_TASK 
   MODIFY CREATE_TIME_ timestamp null;
</code>

Thank you!

tijsrademakers:
Merged manually

Issue #686, opened 2015-10-07
DefaultFlow of a exclusive gateway is lost after export/ import process.


vasiledirla:

tijsrademakers:
Merged the pull request

PR #723, opened 2015-11-06
dev vasile 686 (V5): -fixed the defaultFlow after importing a process

vasiledirla:
fix for: #686: -fixed the defaultFlow after importing a process


tijsrademakers:
Merged manually

PR #728, opened 2015-11-09
ACT-4051 - Spring Boot support does not discover processes in externa…

joepete04:
Fix for ACT-4051

Simply needs to getResources (instead of getResource) since `classpath*:/processes/` can result in an array of processes folders


tijsrademakers:
Already fixed in Spring boot update

PR #734, opened 2015-11-14
remove javaagent from start scripts

smirzai:
I think not every body has jrebels jar files on his machine.

### DIFF
--- a/scripts/start-dev-tomcat.sh
+++ b/scripts/start-dev-tomcat.sh
@@ -5,23 +5,21 @@ then
 	echo ""
 	echo "Cannot boot Activiti REST web application : TOMCAT_HOME is not set"
     echo ""
-	
+
 else
 
 	echo ""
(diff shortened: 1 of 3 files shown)


jbarrez:
Would it be possible to have two versions of the start scripts? So those people that do have can use it ;-)

Joram

smirzai:
hi Joram,
even in that case, do you think they should install it the same place as
that some others you  just mentioned.
How about have one script, default to  no agent, with an additional switch
for the agent owners (with the fix path for the agent file, of course).
Making the agent dynamic, is not much necessary I think. Because nobody has
complained so far.

Saeid

jbarrez:
Indeed, i think not many people run it, or we would have heard complaints. I prefer two different scripts, mainly because im lazy and want to use my terminal autocomplete rather than typing the extra param. 

This is for dev only, so having a bit more scripts doesnt hurt.

Joram

tijsrademakers:
Added scripts that don't require JRebel

PR #740, opened 2015-11-20
do not log the same exception twice

### LINKED ISSUE #739: exception handling anti-pattern in AtomicOperationProcessEnd
When several subprocesses are nested and an exception is thrown, it is logged at each subprocess level. This means a huge stack trace logged several times.

This is the responsible line : 
https://github.com/Activiti/Activiti/blob/master/modules/activiti-engine/src/main/java/org/activiti/engine/impl/pvm/runtime/AtomicOperationProcessEnd.java#L53

tyrcho:
fix exception handling anti-pattern : log and rethrow

would fix #739

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/pvm/runtime/AtomicOperationProcessEnd.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/pvm/runtime/AtomicOperationProcessEnd.java
@@ -17,16 +17,12 @@ import org.activiti.engine.ActivitiException;
 import org.activiti.engine.impl.pvm.delegate.SubProcessActivityBehavior;
 import org.activiti.engine.impl.pvm.process.ActivityImpl;
 import org.activiti.engine.impl.pvm.process.ScopeImpl;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 
 /**
  * @author Tom Baeyens
  */
 public class AtomicOperationProcessEnd extends AbstractEventAtomicOperation {
-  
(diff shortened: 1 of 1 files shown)


tijsrademakers:
Hi, 

Can you provide a test case showing the issue?
Is the exception logged somewhere else on "error" level, or is it on "debug" level?

Thanks.

tyrcho:
Hi @tijsrademakers 

Sorry I won't have the time to provide a test case, I'm only involved in projects using Activiti at the side of my daily activity at the moment.

The exception is logged each time on error level. The workaround is this log4j configuration :
`log4j.logger.org.activiti.engine.impl.pvm.runtime.AtomicOperationProcessEnd=OFF`

tijsrademakers:
Hi @tyrcho, Ok understood. I'm trying to reproduce the issue, but haven't succeeded yet. I was testing with a parent process with a call activity, and the sub process has another call activity and in the nested sub process an error was thrown because the service task class wasn't correct. But then this log message isn't created. So I was wondering which BPMN constructs and process definition you were testing with.

tyrcho:
I've tried also to reproduce the issue with a simple test but failed. It might be a side effect of our usage of camel.
I still think th PR is valid because it fixes a code anti pattern, but I can't prove when it will be relevant ...

Sorry, I won't have more time to help with that before at least 2 months. I still add the real log file where we detected the issue.
[log.txt](https://github.com/Activiti/Activiti/files/47517/log.txt)


Cheers

tijsrademakers:
Closing the PR because we don't want to change this for Activiti 5 without a unit test showing the issue.

PR #747, opened 2015-11-27
IllegalAccessException using the activiti-context.xml file inside a cdi context

flashboss:
I emule a bug using the activiti-context.xml file inside a cdi environmnent. Adding this file in the resources of the application I
have this exception:

Caused by: java.lang.IllegalAccessException: Class
org.activiti.engine.ProcessEngines can not access a member of class
org.activiti.spring.SpringConfigurationHelper with modifiers "public
static"

Infact we cannot use the reflection to call the
SpringConfigurationHelper because the class is protected through the
default access modifier. I see two solutions to resolve this problem:

1 - set to public the org.activiti.spring.SpringConfigurationHelper
class:
    public class SpringConfigurationHelper
    instead of:
    class SpringConfigurationHelper
2 - remove the managing of activiti-context.xml file in the
org.activiti.cdi.impl.ActivitiExtension of the activiti-cdi project so
it doesn't catch the IllegalAccessException

### DIFF
--- a/modules/activiti-cdi/pom.xml
+++ b/modules/activiti-cdi/pom.xml
@@ -118,6 +118,10 @@
       <artifactId>activiti-engine</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.activiti</groupId>
+      <artifactId>activiti-spring</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.jboss.arquillian.junit</groupId>
       <artifactId>arquillian-junit-container</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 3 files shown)


tijsrademakers:
Hi,

Would it be possible to create a unit test for this issue and make sure that only this unit test is using activiti-context.xml (possibly needs a rename of this file)?

flashboss:
I try.... I could create an alternative src/test/resources repository and move it when the test for activiti-context.xml is started

flashboss:
I've choosen to put the activiti-context.xml because all current tests are executed so the tests are more completed. Otherwise I can start an only new test to test the achievement of the activiti-context.xml file

tijsrademakers:
Changed the SpringConfigurationHelper modifier to public in a separate commit.

PR #751, opened 2015-12-04
ACT-1395: Allow to fetch tasks that are in a candidateGroup and that …

esarbanis:
…are already assigned


jbarrez:
This is breaking existing behaviour, which would break the code of people depending on it. 
A better way to do this is to introduce a new method specifically for this use case.

PR #752, opened 2015-12-08
factor deletion's exception byte array management and the execution/j…

stpm:
…ob relationship management out into separate methods.


jbarrez:
Makes sense, thanks.

PR #754, opened 2015-12-18
Initialize custom CommandContextFactories

marcus-nl:
An explicitly set (custom) CommandContextFactory should be initialized.


jbarrez:
Makes sense, thanks!

Issue #753, opened 2015-12-11
NullpointerException occured on referring Historic Variable(Serializable Type)


daisuke-yoshimoto:
Hi,

We found a bug, we report.
We are experiencing a NPE on referring Historic Variable(Serializable Type):

((HistoricVariableInstance) historicData).getValue();

# Occurrence condition
We are experiencing a NPE on referring HistoricVariableInstance(Serializable Type).
We are not experiencing a NPE on referring HistoricVariableInstance(String Type, etc.).

# Cause
Only Serializable Type, HistoricVariableInstance.class#getValue gets data from not Cache but RDB.
Therefore, HistoricVariableInstance.class#getValue try to get Context, but Context is null.

# Steps to Reproduce

0. We will build Postgres 9.4 environment.

1. We will build Activiti environment.
* Download https://github.com/Activiti/Activiti/releases/download/activiti-5.19.0/activiti-5.19.0.zip.
* Place "activiti-explorer.war" into in Tomcat's webapps directory, and then start the Tomcat.
* Place Postgres 9.4 JDBC Driver jar into %Tomcat_Root%/lib directory.
* Describe the connection setting in %Tomcat_Root%/webapps/activiti-explorer/WEB-INF/classes/db.properties.

2. We will install The project of the attached file.
https://drive.google.com/open?id=0ByO-BAU8ivtKcF80UUFUX0YtVU0

3. We will describe db.properties in %Project_Root%/src/main/resources/db.properties.

4. Execute jp.yoshimoto.historic_variable_npe.NullpointerExceptionReproduction.class, and then Check NullpointerException.

* Exception in thread "main" java.lang.NullPointerException
	at org.activiti.engine.impl.persistence.entity.ByteArrayRef.ensureInitialized(ByteArrayRef.java:96)
	at org.activiti.engine.impl.persistence.entity.ByteArrayRef.getBytes(ByteArrayRef.java:46)
	at org.activiti.engine.impl.persistence.entity.HistoricVariableInstanceEntity.getBytes(HistoricVariableInstanceEntity.java:138)
	at org.activiti.engine.impl.variable.ByteArrayType.getValue(ByteArrayType.java:32)
	at org.activiti.engine.impl.variable.SerializableType.getValue(SerializableType.java:49)
	at org.activiti.engine.impl.persistence.entity.HistoricVariableInstanceEntity.getValue(HistoricVariableInstanceEntity.java:129)
[...]

PR #755, opened 2015-12-19
Fix NPE occured on referring Historic Variable(Serializable Type)

daisuke-yoshimoto:
Fix NullpointerException occured on referring Historic Variable(Serializable Type) #753

TestCase
https://drive.google.com/open?id=0ByO-BAU8ivtKcF80UUFUX0YtVU0

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/HistoricVariableInstanceEntity.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/HistoricVariableInstanceEntity.java
@@ -112,8 +112,10 @@ public class HistoricVariableInstanceEntity implements ValueFields, HistoricVari
   	persistentState.put("textValue2", textValue2);
   	persistentState.put("doubleValue", doubleValue);
   	persistentState.put("longValue", longValue);
-  	persistentState.put("byteArrayRef", byteArrayRef.getId());
-  	
+  	if(byteArrayRef.getId() != null){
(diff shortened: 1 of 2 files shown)


daisuke-yoshimoto:
I added a test case of activiti-engine project style.
https://github.com/daisuke-yoshimoto/Activiti/commit/427d4ce2904670ce9fb1caeef43fdc025af59a7c

daisuke-yoshimoto:
If possible , please feedback on this pull request.
My correction method or test case was not appropriate?

jbarrez:
@daisuke-yoshimoto :  no problem at all, just the Christmas holidays that caused a bit of a delay in following up pull requests :-)

Thanks for the excellent test case (and finding of the bug). Indeed, the test was perfectly in line with the style of the Activiti code, many thanks!

I did do a small tweak to your fix, see https://github.com/Activiti/Activiti/commit/371fe190f16b6128822201004a21836105249e1f. Instead of doing it in the getPersistentState(), moved it to the query itself (this is the same as done in the HistoricVariableInstanceQueryImpl). Otherwise, if left in the getPersistentState() method, the value of a serializeable value would always be fetched eagerly (which would be different from the original behaviour)

daisuke-yoshimoto:
@jbarrez

Thank you for merging my pull request.

> Instead of doing it in the getPersistentState(), moved it to the query itself (this is the same as done in the HistoricVariableInstanceQueryImpl). Otherwise, if left in the getPersistentState() method, the value of a serializeable value would always be fetched eagerly (which would be different from the original behaviour)

Sorry, My consideration was not enough.
I will take care from the next pull request.

PR #765, opened 2016-01-06
exposes the methods to init the deployer dependencies and the parser ...

stpm:
…on ProcessEngineConfigurationImpl.

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java
@@ -1559,7 +1559,7 @@ public abstract class ProcessEngineConfigurationImpl extends ProcessEngineConfig
     }
   }
 
-  protected void initBpmnDeployerDependencies() {
+  public void initBpmnDeployerDependencies() {
     
     if (parsedDeploymentBuilderFactory == null) {
       parsedDeploymentBuilderFactory = new ParsedDeploymentBuilderFactory();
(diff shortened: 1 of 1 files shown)


jbarrez:
Hi @stpm, on itself I don't have anything exposing it, but I'd like to understand the use case for this (also to see which other methods would need a similar treatment).

stpm:
Sure!

We provide our own BpmnDeployer instead of using the standard one.
However, as a result of this, *initBpmnDeployerDependencies* doesn't get
called: if *initDeployers* finds a null value for *deployers*, then it
calls *getDefaultDeployers* which in turn calls
*initBpmnDeployerDependencies*.  In our case, *deployers* isn't null, but
we still need to call *initBpmnDeployerDependencies*.

I'd originally solved this by having a subclass of
ProcessEngineConfigurationImpl, which could then call the superclass's
method.  However, I realized that almost everything I was doing in the
subclass could be done more cleanly in a ProcessEngineConfigurator - except
for calling initBpmnDeployerDependencies.  If this were C++, we could make
it a "friend" class (but if this were C++... I wouldn't be nearly as happy
working with it.)  By making it public, we make it so
ProcessEngineConfigurator can still call it, even though they're in
different packages.

Turns out I don't actually need to call *initBpmnParser*, but I thought I
did when I made the modification.

It isn't ideal.  It might be cleaner to call *initBpmnDeployerDependencies*
regardless of whether there's a call to *getDefaultDeployers* - but I
wasn't sure how that would affect anyone else who was working with
non-standard deployers.

jbarrez:
@stpm : thanks for the thorough explanation, that makes a lot of sense. 

I was first going to say that doing a subclass would be the approach here (that was the idea when the class was first created here), but I can see that indeed using a configurator makes things cleaner.

I'm still in doubt whether this means we should make all init* methods public (as a similar reasoning could  probably be made by other use cases). Alternatively, we could make an init method for everything being instantiated in the *initBpmnDeployerDependencies* method (and they would only be really used by the default BpmnDeployer). Which would also make those instances available to your configurator. I'll have a chat with @tijsrademakers about it and update this issue (and the code).

stpm:
Thanks for the update.  Let me know what you decide.

jbarrez:
@stpm : exposed the other methods publicly, to be consistent.

PR #767, opened 2016-01-09
Make schema creation consistent with other schema operations

vpavic:
Checks for creating history and identity schemas are not consistent with other schema operations, which use ```isDbHistoryUsed``` and ```isDbIdentityUsed``` properties on ```DbSqlSessionFactory```.


tijsrademakers:
Merged manually, thanks

PR #768, opened 2016-01-09
Add history level and schema options properties for Spring Boot auto configuration

vpavic:


vpavic:
Any feedback on this PR?

I'm aware that release ```5.19.0.1``` introduced ```ProcessEngineConfigurationConfigurer``` to ease the customization of configuration, however history level should be considered a fairly important property of the ```ProcessEngineConfiguration``` and should be easier to configure.

PR #775, opened 2016-01-15
activiti-ui: translated Task App into Russian, added model sharing for Groups

sarthome:
Module activiti-ui:

Added runtime language switcher for activiti-ui (TaskApp only). Russian language included.
- added libs/angular-translate-storage-local/ to store current language in browser local storage
- updated libs/momentjs.min.js to the newer one
- added i18n/ru.json
- modified processes and tasks controllers to accept runtime language switching
- added new directive to show language bar in the header

Added model sharing for user groups.
Now it's possible to share the app definition with other groups (was only for users).
All users that are included there, will have the option to add the app to their dashboard.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ target
 /userguide/src/en/index.html
 migration.html
 /modules/**/bin
+
+modules/activiti-ui/activiti-app/src/main/webapp/workflow/i18n/ru.json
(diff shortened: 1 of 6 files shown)


Salaboy:
@balsarori Are we looking into this one? Can we make sure that is rebased to the correct branch? 
@sarthome can you please rebase this to point 6.x if that applies?

PR #779, opened 2016-01-18
Fix ClassNotFoundException when WSDL types with complexElement has mixed="true" attribute

bkmeneguello:


tijsrademakers:
Merged, thanks

PR #783, opened 2016-01-22
fix for https://forums.activiti.org/content/quering-ad-hoc-tasks-particular-task-and-process-params

jan-p:


jbarrez:
Makes sense. Can you also add the unit test you had in the forum post?

jan-p:
I added some test cases, but I was not 100% sure where to correctly put them. I hope this complies with your code standards.

jan-p:
Could you please review this again and merge for the next release version if possible?

PR #785, opened 2016-01-24
Fix NPE occured on referring Historic Variable(JPAEntityVariableType) 

daisuke-yoshimoto:
I solved the fix leaks of https://github.com/Activiti/Activiti/pull/755 .

NPE on referring Historic Variable(JPAEntityVariableType):

((HistoricVariableInstance) historicData).getValue();

java.lang.NullPointerException
	at org.activiti.engine.impl.variable.JPAEntityMappings.findEntity(JPAEntityMappings.java:127)
	at org.activiti.engine.impl.variable.JPAEntityMappings.getJPAEntity(JPAEntityMappings.java:121)
	at org.activiti.engine.impl.variable.JPAEntityVariableType.getValue(JPAEntityVariableType.java:79)
	at org.activiti.engine.impl.persistence.entity.HistoricVariableInstanceEntity.getValue(HistoricVariableInstanceEntity.java:129)

# Occurrence condition
We are experiencing a NPE on referring HistoricVariableInstance(JPAEntityVariableType).
We are not experiencing a NPE on referring HistoricVariableInstance(String Type, etc.).

# Cause
In the case of JPAEntityVariableType, HistoricVariableInstance.class#getValue gets data from not Cache but RDB.
Therefore, HistoricVariableInstance.class#getValue try to get Context, but Context is null.


daisuke-yoshimoto:
Because it was incorrectly correspondence of conflict , I cancel the pull request .

PR #788, opened 2016-01-27
Removal the restriction that latestVersion() can only be used in combination with key() or keyLike()

daisuke-yoshimoto:
# Overview

ProcessDefinitionQuery throws ActivitiIllegalArgumentException when I used latestVersion() with name() and nameLike, and so on.
Because I thought all other criteria (name) don't defeats the purpose of the "latest" concept, I removed the restriction and added test cases.

https://github.com/Activiti/Activiti/blob/master/modules/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java

// latest() makes only sense when used with key() or keyLike()
if (latest && ( (id != null) || (name != null) || (nameLike != null) || (version != null) || (deploymentId != null) ) ){
	throw new ActivitiIllegalArgumentException("Calling latest() can only be used in combination with key(String) and keyLike(String)");
}

# Detail

According to the following exchanges,I thought that ProcessDefinitionQuery 
restricts because using latestVersion() method with name() defeats the purpose of the "latest" concept.
https://forums.activiti.org/content/why-latest-can-only-be-used-combination-key

However , using latestVersion() method with name() do't defeat the purpose of the "latest" concept.

When I customized to remove the restriction , the following SQL was executed.
Since nameLike and latest has become a product set, the purpose of the "latest" concept cannot be defeated.

"select RES.* from ACT_RE_PROCDEF RES 
WHERE
RES.NAME_ like '%NAME%'
and
RES.VERSION_ = (select 
max(VERSION_) from ACT_RE_PROCDEF where KEY_ = RES.KEY_ and ( (TENANT_ID_ IS NOT NULL and TENANT_ID_ 
= RES.TENANT_ID_) or (TENANT_ID_ IS NULL and RES.TENANT_ID_ IS NULL) ) ) order by RES.NAME_ 
asc LIMIT 10 OFFSET 0"

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
@@ -309,11 +309,6 @@ public class ProcessDefinitionQueryImpl extends AbstractQuery<ProcessDefinitionQ
   
   public void checkQueryOk() {
     super.checkQueryOk();
-    
-    // latest() makes only sense when used with key() or keyLike()
-    if (latest && ( (id != null) || (name != null) || (nameLike != null) || (version != null) || (deploymentId != null) ) ){
-      throw new ActivitiIllegalArgumentException("Calling latest() can only be used in combination with key(String) and keyLike(String)");
-    }
   }
   
   //getters ////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


daisuke-yoshimoto:
If possible , please feedback on this pull request.
My correction method or test case was not appropriate?

jbarrez:
Merged, thanks. No problem with the pull request (just the usual business)!

daisuke-yoshimoto:
Thank you for merging my pull request.

PR #790, opened 2016-01-27
Activiti6

ihnaqvi:
Changes to initializeVariableInstanceBackPointer before inserting the variable scope entity. This change makes it easier for the datamanager to insert the entity and not rely on any caching.


jbarrez:
Makes sense, merged. Thanks!

PR #803, opened 2016-02-04
Fix bug to make it possible to query event subscriptions by id

reva:
Just a little bug fix


reva:
Could someone look at this easy PR?

jbarrez:
Thanks, this is indeed a very valid fix!

PR #805, opened 2016-02-09
ACT-4122 cannot start a process instance without history service

lsmall:
A SQL error occurs if a process instance is started without enabling the
history service.

Cause: org.h2.jdbc.JdbcSQLException: Table "ACT_HI_TASKINST" not found;
SQL statement:
select * from ACT_HI_TASKINST where ID_ = ?


jbarrez:
Hi @lsmall : Many thanks for catching this. You're absolutely right. It also showed us we had to add the none-history job on our QA bos as well.

I started merging this PR, but realized that the update method wasn't implemented in an optimal way when it comes to regular process user task execution (the extra query was not nice). So I've refactored the logic in the update (actually removed it) and moved it to the right places.

See commit https://github.com/Activiti/Activiti/commit/be1ce1e506dd3520f87300490994afcf6a7849a7

PR #806, opened 2016-02-09
Separate out TaskEntity creation and insertion.

ihnaqvi:
Instead of creating and inserting the task entity in one go, this changes creates the task entity with necessary values and then inserts it.


jbarrez:
Valid change. Thanks @ihnaqvi.

PR #809, opened 2016-02-10
Activiti 5.17.0.dorsale

Paueti:
There is a bug when we want to retrieve the task's local variables plus the process variables. The local variables of the current task are returned as well.

This problem presents itself when retrieving a historicTask. Along with the historicTask's local variables, the current task local variables are returned.

To fix this, the SQL in the HistoricTaskInstance.xml has been modified when includeTaskLocalVariables and includeProcessVariables are true, by adding the condition already present when only includeProcessVariables is true: `and VAR.TASK_ID_ is null`

### DIFF
--- a/modules/activiti-engine/pom.xml
+++ b/modules/activiti-engine/pom.xml
@@ -4,7 +4,32 @@
   <modelVersion>4.0.0</modelVersion>
 
   <name>Activiti - Engine</name>
+  <groupId>eu.europa.ec.digit.dorsale.third.org.activiti</groupId>
   <artifactId>activiti-engine</artifactId>
+  
+  <repositories>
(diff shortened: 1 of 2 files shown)


jbarrez:
Can you add a description of what you're fixing/doing here?

Also the changes in the pom.xml are not wanted.

Paueti:
Hi, I am not very familiar with git. I added the comment in the commit:
"There is a bug when a we want to retrieve the task's local variables plus the process variables. The local variables of the current task are returned as well." Where else should I add it?

Also, how to request the pull only of the file?

Thanks in advance.

tijsrademakers:
Merged, thanks

PR #815, opened 2016-02-19
Fixes activiti-explorer login form

ghnat:
When application is behind proxy (like nginx) over https - login form.action has http scheme.

The issue is that when I'm running activiti-explorer behind proxy and proxy exposes application over https, I cannot log in because of 'mixed-content'. This is because on my browser I have https:// url but inside the application thinks I'm running unsecured and provides login form action param like http://<host>/activiti-explorer/ui/1/loginHandler. After the fix the url here would simply be "1/loginHandler" so it will not try to guess the scheme. It seems to work. Please check the patch, I don't know activiti internals too good so there may be a reason why login form's action param was composed the way it was.


tijsrademakers:
Fixed in another PR by removing the http / https part of the url

PR #825, opened 2016-03-01
Add mapping for JobEntity.dueDate + fix resulting issues

marcus-nl:
(1) Add mapping for JobEntity.dueDate + fix resulting issues.


jbarrez:
@marcus-nl : we're re-read the emails we've been exchanging on this topic, and came to the conclusion there was indeed a window of opportunity to get a null duedate if a job was in the retry cmd and the server went down for some reason.

Instead of fixing the mappings, we decided the change the query that is used to fetch the jobs: https://github.com/Activiti/Activiti/commit/ef2c715da21082bb03d06b4e33a5ce9eb73e809c

This should also solve the problem you fixed in this commit. If not, please reopen or reach out to discuss.

PR #829, opened 2016-03-03
Add support for retrieving a VariableInstance and DataObject from the TaskService.

roberthafner:
This support is needed to retrieve localization values for Variables via the TaskService.

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/TaskService.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/TaskService.java
@@ -21,6 +21,7 @@ import java.util.Set;
 
 import org.activiti.engine.impl.persistence.entity.VariableInstance;
 import org.activiti.engine.query.NativeQuery;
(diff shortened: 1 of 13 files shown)


tijsrademakers:
Rob, do you think this is still needed or are the RuntimeService getDataObject methods sufficient?

roberthafner:
Yes, I think the this change balances out the api so that a variable can be retrieved in same formats (name/value, VariableInstance, DataObject) across the RuntimeService or the TaskService.

I plan to update the PullRequest to remove the localization support from the getVariableInstance methods and add the new getDataObject methods. So let’s hold off merging this PR until I include those updates.

thanks!

roberthafner:
Tijs, I've moved the localization support in this Pull Request from the getVariableInstance* methods to the  getDataObject* methods. 

thanks,
Rob

roberthafner:
Tijs, I've updated this Pull Request to also include the changes to support different localization values for DataObjects of the same name as we discussed. 

Thanks,
Rob

PR #830, opened 2016-03-03
Act-4116 Custom subprocess parser

lsmall:
We really need both createXML() method calls in BpmnXMLConverter to be protected. It looks like the second one has been updated already (since my last attempt at a pull request). But if you could also consider adding the new SubprocessXMLConverter that would be great. I tried to get it working with the existing Designer code, but got to where the next id was determined for each BPMN element within the UI and it was looping through and finding the next one. Wouldn't it be better to use a UUID or keep a key list somewhere with the next key value? Don't you use ids or sids for the web-based design tool already?


lsmall:
Separated out the basic BpmnXMLConverter change due to merge conflicts with other updates waiting for this issue to be merged.

PR #831, opened 2016-03-04
ACT-4103: Unused assignment and possible integer overflow

dbmalkovsky:
Fix for ACT-4103 at master


tijsrademakers:
Thanks, implemented a slightly different fix

PR #836, opened 2016-03-07
improved Time boundary events tests

vasiledirla:
added 2 more Boundary Timer Events tests to test the timer events after a fixed time without repetition

added these tests just to test this reported issue.
https://forums.activiti.org/content/boundary-timer-event-not-work-activiti519


tijsrademakers:
Merged, thanks

PR #844, opened 2016-03-16
ACT-4133(test): test for user task create listener in the subprocess

martin-grofcik:
Regression in 6.0beta2 regarding Listeners


jbarrez:
@martin-grofcik : is this to demonstrate the regression (cause I don't see a fix)?

jbarrez:
@martin-grofcik : apologies , just saw your comment on https://activiti.atlassian.net/browse/ACT-4133. Not able to reproduce with this test. Thanks!

PR #845, opened 2016-03-16
Fix - Include Process Instance execution in PROCESS_STARTED event.

roberthafner:
The entity included on the ActivityEvent with a type of PROCESS_STARTED is the execution for the start event and not the execution of the process instance. This change updates the entity on the event to the process instance execution.

In addition, I've added a new constructor to the StartProcessInstanceCmd to allow the process instance name to be passed in.


tijsrademakers:
There's now a ProcessInstanceHelper class that also allows process instance name to be passed. For the process started event, the start event execution is the correct entity to my opinion. You can always retrieve the process instance entity based on this execution entity.

PR #848, opened 2016-03-17
Timer Boundary Events must stop when end date is reached & Intermediate timer events should ignore endDate

vasiledirla:
Timer Boundary Events must stop when end date is reached.


tijsrademakers:
Can you add a unit test that shows that this is not working correctly now?

tijsrademakers:
Merged the PR with changes

PR #850, opened 2016-03-18
ACT-4116 Custom subprocess parser

lsmall:
Note that this is dependent upon ACT-4135 as I separated out the base BpmnXMLConverter update top avoid future merge conflicts.

I tried to get it working with the existing Designer code, but got to where the next id was determined for each BPMN element within the UI and it was looping through and finding the next one. Wouldn't it be better to use a UUID or keep a key list somewhere with the next key value? Don't you use ids or sids for the web-based design tool already?


lsmall:
Added update for files without existing BPMNDI data.

PR #833, opened 2016-03-06
Add the process definition search criteria.

daisuke-yoshimoto:
We added the search criteria related to process definition.

# ExecutionQuery

We added the following search criteria.
ExecutionQuery#processDefinitionCategory(String processDefinitionCategory)
ExecutionQuery#processDefinitionVersion(Integer processDefinitionVersion)

# ProcessInstanceQuery

We added the following search criteria.
ProcessInstanceQuery#processDefinitionCategory(String processDefinitionCategory)
ProcessInstanceQuery#processDefinitionVersion(Integer processDefinitionVersion)

# HistoricProcessInstanceQuery

1. In the same way as ProcessInstanceQuery, we became able to get processDefinitionName/ProcessDefinitionKey/processDefinitionVersion/deploymentId by using HistoricProcessInstanceQuery.

	In each case, we joined table ACT_RE_PROCDEF like ProcessInstanceQuery.

	In consideration of RepositoryService.deleteDeployment(String deploymentId, false), we selected outer join not inner join.After using this method, ProcessDefinition data was removed but Historic Process Instance data remains.
	http://activiti.org/javadocs/org/activiti/engine/RepositoryService.html#deleteDeployment(java.lang.String, boolean)

2. We added the following search criteria.

	HistoricProcessInstanceQuery#processDefinitionCategory(String processDefinitionCategory)
	HistoricProcessInstanceQuery#processDefinitionName(String processDefinitionName)
	HistoricProcessInstanceQuery#processDefinitionVersion(Integer processDefinitionVersion)

PR #807, opened 2016-02-09
Add the search criteria ProcessInstanceQuery#withException()

daisuke-yoshimoto:
Web added the search criteria(withException) to ProcessInstanceQuery.This criteria get process instances that job failed.

PR #852, opened 2016-03-23
exclude testcase from history level none mode

daisuke-yoshimoto:
We excluded testcase from history level none mode.
Because testcases that we added in this following pull request don't work in history level none mode, we excluded them.
https://github.com/Activiti/Activiti/pull/833
https://github.com/Activiti/Activiti/pull/807


jbarrez:
Many thanks. Awesome that you were able to deduce the solution for the 'none' test suite while we were sleeping !

daisuke-yoshimoto:
@jbarrez

Thank you for merging.

PR #853, opened 2016-03-24
Add support for DB2/LINUXPPC64LE

snaggen:
Activiti did not work when db2 was running on a linux ppc64 little endian. This patch simply add that to the lookup table used to identify the database type.


snaggen:
It would be great if this could be merged to the 5.x branch as well...

balsarori:
What 5.x branch? It's available for v5 and upcoming v6 version

snaggen:
Then I must have missed it when I read the v5 changelogs...

PR #854, opened 2016-03-25
Duplicate conditional in ActivitiListenerExport.java

kilida:
`if (ImplementationType.IMPLEMENTATION_TYPE_THROW_SIGNAL_EVENT.equals(eventListener.getImplementationType()))` is repeated. Looks like this should be `ImplementationType.IMPLEMENTATION_TYPE_THROW_GLOBAL_SIGNAL_EVENT.equals(eventListener.getImplementationType())`


tijsrademakers:
Fixed in a separate commit, thanks for the PR

PR #857, opened 2016-03-25
Fixed problem with process instances query

### LINKED ISSUE #856: SQL error when querying process instances with variables in 'or' and 'like ignore case'
I've added a test case that shows the error:
m4ng0/Activiti@eb96d47

m4ng0:
Fixed a bug that arises when querying process instances with process variables like 'ignore case' conditions in an 'or' clause.
Fixes #856.


jbarrez:
Good catch. Thanks!

Issue #865, opened 2016-03-31
Activiti6: Invalid reference in 'bpmnElement' attribute, sequenceFlow ... not found


reva:
Reopening https://github.com/Activiti/Activiti/issues/801. Could not find the unit tests @yvoswillens was referring. (Could not find ANY unit tests testing BpmnParse, really)

This is a bug in my opinion. Right now I get a log statement about missing sequence flows for every process that I deploy. Working on a pull request.

kartheepanmirra:
Yes the defect is due to the fact that the sequnceFlows maps in BpmnParse.java BpmnParse.sequenceFlows used to get added with values in the SequenceFlowParseHandler.executeParse at line number 55  bpmnParse.getSequenceFlows().put(sequenceFlow.getId(), transition); in the older version 5.19.2.


The current beta version code has the following syntax

protected void executeParse(BpmnParse bpmnParse, SequenceFlow sequenceFlow) {
    org.activiti.bpmn.model.Process process = bpmnParse.getCurrentProcess();
    sequenceFlow.setSourceFlowElement(process.getFlowElement(sequenceFlow.getSourceRef(), true));
    sequenceFlow.setTargetFlowElement(process.getFlowElement(sequenceFlow.getTargetRef(), true));
  }

reva:
The PR is already there. Another simple fix, really. Corresponding jira issue: https://activiti.atlassian.net/browse/ACT-4142

tijsrademakers:
Fixed

PR #866, opened 2016-03-31
Bugfix for #865 - Activiti6: Invalid reference in 'bpmnElement' attri…

reva:
Fix for https://github.com/Activiti/Activiti/issues/865


tijsrademakers:
@reva can you also add a unit test showing the "invalid reference in 'bpmnElement'" issue? Ideally this unit test should use only the default RepositoryService createDeployment logic. So not using BpmnParse directly in the unit test.

reva:
@tijsrademakers The 'issue' that I'm seeing here is really only a logging statement that is not necessary. Beyond that, I don't have any issues. Why not use the BpmnParse directly?

tijsrademakers:
@reva ok, but I still don't understand when this logging statement is shown. So can you add a unit test that would make this logging statement to appear? Using BpmnParse directly is not what most of our community developers will do, so would be better to use the public API service methods.

reva:
@tijsrademakers The logging statement is shown for any process that has any graphical information (di) and flows in it. Anyway, I looked into it and saw that the sequence flows aren't used anywhere. I changed my pull request to remove the map and all logic with it. Feel free to decline this one...

tijsrademakers:
Fixed it by changing the sequence flow lookup code.

PR #870, opened 2016-04-14
WORK-635 mleber processinstance query involved user like ignore case

MaggieLeber:
Time to land this guy.


MaggieLeber:
Oops....sorry. Never mind.

PR #873, opened 2016-04-15
The runAs token for Alfresco should be "runAs" not "runAS" as defined in

chrisrobbo:
SCRIPT_DELEGATE_RUN_AS_FIELD_NAME. This causes Eclipse plugin to output
wrong XML which causes Activiti run time error for at scripts defined
with Run As users


tijsrademakers:
Merged manually, thanks

PR #874, opened 2016-04-15
Update CallActivityBehavior to send PROCESS_STARTED event.

roberthafner:
This change is dependent on a prior pull request that refactored CallActivityBehavior to support initializing DataObjects in a subclass.


roberthafner:
Thanks for the update! We'll adjust our code accordingly.

PR #875, opened 2016-04-18
Update CallActivityBehavior to send PROCESS_STARTED event.

roberthafner:


tijsrademakers:
Merged the PROCESS_STARTED event change (but used the start event execution instead).

PR #876, opened 2016-04-19
Make explicit update() calls via dataManagers to activate/suspend a p…

bokyung:
…rocess instance

instead of relying on Entity cache's implicit update behavior. This will help for better readability, and allow special handling for entities in the overriding data managers.


jbarrez:
Thanks. Did a little tweak to not fire an event (see commit above) as to be consistent with the existing event firing.

bokyung:
Thanks a lot, jbarrez! This will help us to keep the activated/suspended states of multiple entities in sync.

PR #883, opened 2016-05-05
Refactor 'VariableInstance' terminology to 'DataObject'

roberthafner:
Tijs, here are the changes we discussed via email to use the term 'DataObject' instead of 'VariableInstance' in the API.


roberthafner:
Tijs, I'm not sure how you want to define what can be retrieved as a DataObject. For now the PullRequest only returns a variable as a DataObject if it is modeled in the BPMN file. We could loosen this up to return  all variables if you would prefer. One concern I have with the current approach is that if a DataObject X is defined on the process instance and then a variable X is defined on a subprocess, attempting to retrieve X as a DataObject from the subprocess execution will result in a null value being returned.

Also, regarding the issue with using the variable name as the key for the localization lookup. I think we could prefix the variable name with the id of the process or subprocess to support multiple values for variables of the same name since we already need to go to the BpmnModel for the description anyway. What do you think?

tijsrademakers:
Rob, I've committed the changes you did in this PR in the following commit:

https://github.com/Activiti/Activiti/commit/29f248791495bb2e080d437b4401e26667114e42

I added the getDataObject(s) methods, but kept the getVariableInstance(s) methods as well. The getVariableInstance methods are still there to have a lower level query API for variables, but localization is not part of that anymore. The getDataObject(s) methods have the localization logic. 

Yes, prefixing the variable name with the id of the process or sub process would be nice. But to make that required is maybe not so nice. Because then you would always need to use the dot notation when using the DynamicBpmnService to change the localization.

roberthafner:
Tijs, Does it make sense to use the id of the DataObject element in the BPMN document as the key in th e DynamicBpmnService cache now that localization has been limited to DataObjects (which must be modeled)?

tijsrademakers:
The DataObject id can be reused across different BPMN XMLs as well, so I don't think it solves the unique reference issue. Of course we could explain that the DataObject id needs to be unique to be able to localize the DataObject specific for a process definition. Maybe we could check if there's an entry with a processId.dataObjectName pattern or processId.subProcessId.dataObjectName and fallback to the dataObjectName when no entry could be found in the localization entries?

roberthafner:
Looking at the code again, it seems the ProcessDefinitionInfoCacheObject stored by the DynamicBpmnService is scoped based on the process definition id (and not the process definition key). I believe this means the re-use of the same id for a DataObject element across different BPMN definition files or even different versions of the same BPMN definition file should not conflict. Correct me if I am wrong here.

Assuming that's correct, I think we can return the value of the id attribute of the DataObject element defined in the BPMN file via the org.activiti.engine.runtime.DataObject class for the DynamicBpmnService to use when changing the localization values. 

Sound reasonable? If so, I can look into making these changes if you'd like.

tijsrademakers:
Yes, good point. Yes that's sounds reasonable. If you could look into making these changes that would be great, thanks.

PR #889, opened 2016-05-12
Refactor test to run under JDK7 and JDK8

dbmalkovsky:
Changed the script test so it works across JDK7 and JDK8 on OSX, Windows, and Linux.  This small change is "copied" from the method used in the previous test method (testAutoStoreVariables) at line 79.  This change "standardizes" the result so the assertEquals() is successful.

### DIFF
--- a/modules/activiti-engine/src/test/java/org/activiti/examples/bpmn/scripttask/ScriptTaskTest.java
+++ b/modules/activiti-engine/src/test/java/org/activiti/examples/bpmn/scripttask/ScriptTaskTest.java
@@ -86,20 +86,20 @@ public class ScriptTaskTest extends PluggableActivitiTestCase {
       assertTextPresent("No script provided", e.getMessage());
     }
   }
-  
+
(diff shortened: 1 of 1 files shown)


jbarrez:
@dbmalkovsky : what is failing on jdk7/jdk8? Is the assertEquals not executing correctly?

http://ec2-54-217-5-246.eu-west-1.compute.amazonaws.com:8080/jenkins/ runs everything on JDK8 and we've got a dedicated job for JDK 7. Would like to understand what's different in the setup if you see this behaviour.

roberthafner:
Yes, the test fails in the assert statement.  I am able to reproduce the issue when running JDK 1.8.0_25 on both a linux and OS X environment. Upgrading the JDK to 1.8.0_92 resolves the failure.

jbarrez:
@roberthafner : thanks. Merging it, doesn't seem to do harm. Just wanted to understand why it was failing.

PR #895, opened 2016-05-19
Activiti6 transaction dependent listeners

yvoswillens:
Added support for transaction dependent execution listeners. 
Support for 'committed' and 'rolled-back'.
During execution when the configured event occurs these will be put on a stack (which implements ContextCloseListener).
These listeners can be configured as a class delegate or as delegate expression (not an expression).
A custom properties resolver can be configured that will make the resolved properties available to the listener.

```
<activiti:executionListener delegateExpression="${myTransactionDependentExecutionListener}"
                            event="end"
                            onTransactionResult="committed"
                            customPropertiesResolverDelegateExpression="${myCustomPropertiesResolver}"/>
```


yvoswillens:
is part of PR #903

PR #897, opened 2016-05-22
if not set databaseType, before buildProcessEngine, should check if s…

yuanqixun:
…etCurrentTenantId.


yuanqixun:
if not set databaseType, when buildProcessEngine, there will be a exception.

jbarrez:
databaseType is mandatory when using msmt process engine config. I added an exception here: https://github.com/Activiti/Activiti/commit/192bc6bec0ad9f41fc81007744b8d066b0fcbf7d, to make sure users know it has to be set.

Closing this PR: what the code here does is looping over the tenants and as a side effect setting the 'currentTenantId' each iteration of the loop. The last tenantId remains in the 'currentTenantId' and this is why the exception goes away, as the TenantAwareDataSource has now a datasource it can find as the lingering tenantId is still there.

PR #902, opened 2016-05-27
* MNT-16040 : skipCount does not appear to work for Workflow tasks RE…

ramunteanu:
SHA-1: 8a074615958730f9c83c95fda099c342599dffae

* MNT-16040 : skipCount does not appear to work for Workflow tasks REST api

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/history/HistoricTaskInstanceQuery.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/history/HistoricTaskInstanceQuery.java
@@ -407,6 +407,11 @@ public interface HistoricTaskInstanceQuery  extends Query<HistoricTaskInstanceQu
    */
   HistoricTaskInstanceQuery includeProcessVariables();
   
+  /**
+   * Limit task variables
+   */
+  HistoricTaskInstanceQuery limitTaskVariables(int taskVariablesLimit);
+  
   /** Order by task id (needs to be followed by {@link #asc()} or {@link #desc()}). */
   HistoricTaskInstanceQuery orderByTaskId();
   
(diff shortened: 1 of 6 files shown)


daisuke-yoshimoto:
@ramunteanu

I also encountered the same problem that cannot use sql paging with includeTaskLocalVariables・includeProcessVariables.

I solved this problem by the following pull request.
https://github.com/Activiti/Activiti/pull/891

With includeTaskLocalVariables・includeProcessVariables false, I divided into two queries.
First query get tasks.
Second query get variables by taskIds.

ancutam:
@tijsrademakers Could you please port the fix to the 5.16.5 and 5.19.0.x versions?

tijsrademakers:
@ancutam released 5.16.6 with this PR change based on 5.16.5. Which version of 5.19.0.x do you need exactly? I would recommend to use 5.19.0.2 because of several bug fixes in that release. But it also includes an upgrade to MyBatis 3.3.0.

gavincornwell:
@tijsrademakers Is the upgrade to MyBatis 3.3.0 just a library upgrade or are there other changes that we'll need to make i.e. DB scripts etc? Just to be clear is you suggestion to apply the changes and create a 5.19.0.3 version of Activiti?

ancutam:
@tijsrademakers @gavincornwell we are using 5.19.0.1. 
I see 5.19.0.2 has also a security issue fixed.

gavincornwell:
@tijsrademakers @ancutam I spoke to Derek about the MyBatis upgrade, it's something we should try, if that's successful then we could take a 5.19.0.3, otherwise we'll need a hotfix version of 5.19.0.1.

paulhh:
@gavincornwell I need to understand the priority of this requirement.  I'll follow up by email.

ancutam:
@tijsrademakers the MyBatis upgrade to 3.3.0 went well. Can you please port the fix to 5.19.0.2? Thank you!
(cc: @gavincornwell)

gavincornwell:
@tijsrademakers @paulhh Just to be crystal clear, we're requesting a 5.19.0.3 release which just includes the fix from this pull request.

paulhh:
What sort of time frame?  We have a major release we're trying to get out with a very small team :)

gavincornwell:
This was for a hotfix which we've already gone way over our SLA for, so as always it's ASAP :-( 

However, realistically we have until the end of the sprint which is 1st August so before Thursday next week (28th) to give us time for builds and verification would probably work, would that work?

paulhh:
Hopefully, Tijs can get it done by then.

gavincornwell:
Thanks @paulhh, much appreciated.

PR #917, opened 2016-06-10
Fix - GetTaskDataObjectCmd should return a dataobject when locale is …

roberthafner:
…null.

Missed removing the if(locale != null) check when refactoring the GetTaskDataObjectCmd. Including the check results in a null value being returned when the locale is null. 

Added two additional tests to verify behavior retrieving a data object from a task without a locale works correctly.


roberthafner:
Tijs, can I get a status update on this pull request? I'm hoping this change will be included in the planned beta release. 

Thanks,
Rob

tijsrademakers:
Hi Rob, we are working on stabilizing the unit tests first now. But I'll make sure this PR will be included in the next beta release.

PR #943, opened 2016-07-03
update historic activity and historic task process definition id in SetProcessDefinitionVersionCmd

brig:


tijsrademakers:
Made some small changes to the PR and committed the following changes:

https://github.com/Activiti/Activiti/commit/07fba6052a1e28007cd47f84f0fed3d6fa9d9c5e

The historic activity instance already was updated as tested with the historic activity instance process definition test in the ProcessInstanceMigrationTest class.

PR #946, opened 2016-07-08
further changes after the removal of ActivityExecution

dbmalkovsky:
As part of the [19fad05](https://github.com/Activiti/Activiti/commit/19fad051eb349eee9abd570bd44c86c8473c4983#diff-7f31b80a406e836752411a3dcdfa6766) commit on 8/27/2015, when ActivityExecution was removed it appears that a change to the method parameters was made but there are some subsequent changes to the code that are applicable:

(1) fix the javadoc
(2) remove the ExecutionEntity casts as the parameter is already an ExecutionEintity
(3) remove the instanceOf test for the same reason; I replaced it with a null check just to be safe


jbarrez:
That's correct. Thanks!

PR #950, opened 2016-07-14
set retries makes job reexecutable again

martin-grofcik:
The jUnit test shows that after setting retries_ > 0 failed job is not fetched by the async job executor to be re-executed.
the fixed job.xml changes selectAsyncJobsDueToExecute to select all jobs without expiration time and due date null.

Another possibility is to set due date to now (there are still some test failing.). It could make sense too.
(?) which approach do you prefer?

### DIFF
--- a/modules/activiti-engine/src/main/resources/org/activiti/db/mapping/entity/Job.xml
+++ b/modules/activiti-engine/src/main/resources/org/activiti/db/mapping/entity/Job.xml
@@ -1,7 +1,7 @@
-<?xml version="1.0" encoding="UTF-8" ?> 
+<?xml version="1.0" encoding="UTF-8" ?>
 
-<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd"> 
-  
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
 <mapper namespace="org.activiti.engine.impl.persistence.entity.JobEntity">
 
   <!-- JOB DELETE STATEMENTS-->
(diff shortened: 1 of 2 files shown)


jbarrez:
@martin-grofcik : this is as designed: when retries is 0, the job is moved to the deadletterjob table. it needs to be moved to the regular table again for being picked up again (im actually writing that move logic today).

The idea in the new executor is that when its in ACT_RU_JOB, it's executable. I'll publish a blog soon about the new executor and how it works (there are not 4 tables).

The reason to keep the query as simple as possible, is because on huge installations, adding more stuff to the where clause is causing table scans.

martin-grofcik:
Is deadletterjob present in 5.22-SNAPSHOT branch too? ( I did not find any)

jbarrez:
No, the new AsyncExecutor design is v6 only.

jbarrez:
Sorry, my bad, thought this was v6 ... hence my confusion ;-)
Let me check the test.

jbarrez:
Ok, I see it now. Setting the retries to a number > 0 will make the job be picked up only after the lock has been expired. I think it would make sense to reset both the lock owner/time (and due date probably too) when calling that setRetries method. (I'm always weary of changing the query, the where clause has a big impact on performance)

martin-grofcik:
@jbarrez 
You are right.
But setRetries can be called for the timer (dueDate) and currently running job (expiration) too. -> We should not change these two values.

jbarrez:
That's true. I don't think it makes sense to add it to setRetries in that case, too many side effects. 

Eventually, the lock will expire and it will be picked up again after setting the retries > 0, for a failed job it's not expected that it needs to be executed immediately. If setRetries + clearing lock + duedate is needed, a custom command can always be used.

tijsrademakers:
Closing PR with Joram's comment

martin-grofcik:
When job is failed and retries are set to 0, lockExpirationTime ia already set to null. When we setRetries(X>0) the job is never reexecuted again with AsyncJobExecutor (see test). 

We have 3 possibilities to fix this issue:

1. pull request (with performance downgrade),
2. custom command,
3. use `managementService.executeJob()` instead (running job synchronously)

jbarrez:
I'd go for 3., as for me this is a manual 'admin' kind of operation.
(and actually the setRetries method doesn't make a lot of sense to off in the API)

PR #972, opened 2016-08-03
Error in DB2 create and update scripts

dbmalkovsky:
In create and update scripts for  the table ACT_RU_EXECUTION, the column  IS_COUNT_ENABLED_  specified the incorrect check() constraint variable.  Very simple change.


tijsrademakers:
Thanks

PR #976, opened 2016-08-03
make agenda configurable

gro-mar:
Making agenda configurable gives us opportunity to make engine more flexible.
(see    `org.activiti.examples.runtime.StandardAgendaFailingTest` and `org.activiti.examples.runtime.WatchDogAgendaTest`)

configurable agenda can improve crystalball implementation too


gro-mar:
Is there any issue with the pull request?

tijsrademakers:
Let's discuss this in a call Martin

gro-mar:
@tijsrademakers : This pull request is ready for the merge.

**Note:** We have discussed generic Agenda interface during code review session we had. I have reimplemented Agenda interface to be more generic.

**Proposal**: (not included in the pull request) From my point of view `Agenda` should be able to plan `Operation`. Agenda `Operation` has 2 features: 
1. Operation in agenda is `Runnable` (already included)
2. Operation in agenda is `Comparable`. `Comparable` gives the possibility to place the operation to the correct place in the agenda scheduling. 
I would create new interface Operation with both interfaces included

@tijsrademakers, @jbarrez : What do you think?

PR #980, opened 2016-08-06
Added missing class name in the exception message

Vadi:
Improved the exception message in case the businessRuleTask is configured with the invalid class name. Adding a class name to the exception messages makes the message more readable.


tijsrademakers:
Merged your change manually, thanks

PR #986, opened 2016-08-09
Fix problems found using terminateEndEvent in subprocess.

roberthafner:
A timer intermediate catch event within a subprocess is causing an
invalid process cancelled event to be sent. It should not be sent in this
case. Additionally, the Activity cancelled events are setting the wrong
activity Type.


tijsrademakers:
Thanks, merged the changes manually.

PR #993, opened 2016-08-11
ACT-4027 BaseBpmnXmlParser doesn't support endDate attribute

lsmall:
added support from master to activiti6 branch

also updated unit test to verify this value as well as the time duration
and cycle values


tijsrademakers:
Merged manually, thanks.

PR #1007, opened 2016-08-29
Add the inheritVariables option to callActivityBehavior task.

daisuke-yoshimoto:
We add the inheritVariables option to callActivityBehavior task.

If this option is true, sub process inherit all variables from parent process.


daisuke-yoshimoto:
@tijsrademakers

Thank you for fix and merging.

PR #1011, opened 2016-09-01
ACT-4202: Faults are not supported by the web-service implementation of service task

cdeneux:
This pull request propose to add fault management support on web-service implementation of service task. As mentioned in [ACT-4202](https://activiti.atlassian.net/browse/ACT-4212), a fault based on `org.apache.cxf.interceptor.Fault` that is a technical error is re-thrown as ActivitiException, other faults that are business error are re-thrown as BpmnError to be catchable in an error boundary event. The attributer `errorRef` of the event is initialized with the qualified name of the exception class associated to the fault.

Thread associated on Activiti forum: https://forums.activiti.org/content/catching-web-service-task-exceptions


jbarrez:
Thanks @cdeneux for this PR. I had one question: what's the reason for choosing the Fault as the technical exception? I would have assumed that the Fault, as it is defined in the .wsdl, maps to a business exception and the exception that don't have any match are converted to an ActivitiException,

Even better would be if the mapping business exception <-> webservice exception could be something that is defined in the process definition with some custom xml.

cdeneux:
Hi Joram,

In a technical point of view:
  - if a fault defined in WSDL is thrown on the web-service side, an exception is automatically generated by the Apache CXF stack. This exception is not inherited from `org.apache.cxf.interceptor.Fault`, it's an exception generated on the fly by XJC and inherited from `java.lang.Exception`.
  - in the web-service world, an unexpected error is always converted into a generic fault not defined in the WSDL. This generic fault is associated to the exception `org.apache.cxf.interceptor.Fault`. That's why I choose it as base exception to generate technical error converting it to an ActivitiException,
  - at implementation level, it seems to me that it is not possible to catch business error excepting using `java.lang.Exception`. That's why technical errors are firstly caught through `org.apache.cxf.interceptor.Fault`. And if the exception does not match `org.apache.cxf.interceptor.Fault`, I consider it's a business error.

IMO, in a conceptual or design point of view, business error must be defined at the service contract level, so in the WSDL. And we can consider that all others errors are technical errors. If a web-service is not correctly designed and returns business errors that are not defined in the WSDL, I think that this web-service should be fixed or facadized.

Can you explain me your idea about "_some custom XML for the mapping business exception <-> webservice exception_". I think that I had the same idea but I rejected it because:
  - I'm not be able to find where to put this custom XML: web-service reply is processed through an `<iospecification>/<dataOutputAssociation>` that does not support fault (except if I missed something).
  - I think that using such a mapping is not really business oriented because the process definition could be defined without taking into account these business errors,
  - and, moreover, how the web-service client will know the business errors available, asking developers of the web-service ? It's not really SOA.

jbarrez:
Ok, thanks for the detailed explanation. Makes sense now, thanks. PR merged.

I agree with your statements at the end about the mapping. Probably for webservices it indeed doesn't make sense to do it.

PR #1012, opened 2016-09-01
ACT-4206: Webservice endpoint address should be customizable

cdeneux:
See [ACT-4206](https://activiti.atlassian.net/browse/ACT-4206):
> When using a service task implemented by a web-service, the endpoint address of the web-service is extracted from the WSDL imported into the process definition.
> Once you have deployed your process definition, it is possible to move the web-service, and its endpoint address changes. And now all new process instances will fail.
> To avoid this problem, we should be able to override the endpoint addresses at Activiti engine level, in its configuration for example.

Endpoint addresses are configured through a map where the key is the endpoint qualified name available in the WSDL, and the value its location

### DIFF
--- a/modules/activiti-cxf/src/main/java/org/activiti/engine/impl/webservice/CxfWebServiceClient.java
+++ b/modules/activiti-cxf/src/main/java/org/activiti/engine/impl/webservice/CxfWebServiceClient.java
@@ -16,10 +16,16 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Enumeration;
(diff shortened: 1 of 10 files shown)


jbarrez:
Merged, thanks.

Applied one small change: moved the map to ProcessEngineConfigurationImpl instead of ProcessEngineConfiguration (as thats where the advanced settings are)

cdeneux:
Thanks Joram for the merge.

Perhaps can you update the associated issue in JIRA: #[4206](https://activiti.atlassian.net/browse/ACT-4206)

jbarrez:
Indeed. Done.

PR #1020, opened 2016-09-12
Add support for TIMER_SCHEDULED event.

roberthafner:
Neither the ACTIVITY_STARTED OR ENTITY_CREATED/ENTITY_INITIALIZED events can be used as an event type for a callback hook in a listener to know when a timer is started. ACTIVITY_STARTED does not work since a timer is not an 'activity'. The ENTITY_CREATED/ENTITY_INITIALIZED event types do not help since they are sent multiple times (i.e when the process has reached the timer element and when the timer is fired). This pulls request adds a new event (TIMER_STARTED) that is sent after a timer is scheduled which a listener can use.

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/delegate/event/ActivitiEventType.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/delegate/event/ActivitiEventType.java
@@ -59,6 +59,11 @@ public enum ActivitiEventType {
   ENTITY_ACTIVATED,
 
   /**
+   * A Timer has been scheduled.
+   */
+  TIMER_SCHEDULED,
+
+  /**
    * Timer has been fired successfully.
    */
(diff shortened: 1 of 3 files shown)


roberthafner:
Neither the ACTIVITY_STARTED OR ENTITY_CREATED/ENTITY_INITIALIZED events can be used as an event type for a callback hook in a listener to know when a timer is started. ACTIVITY_STARTED does not work since a timer is not an 'activity'. The ENTITY_CREATED/ENTITY_INITIALIZED event types do not help since they are sent multiple times (i.e when the process has reached the timer element and when the timer is fired). 

This pulls request adds a new event (TIMER_SCHEDULED). The TIMER_SCHEDULED event is sent when a process reaches a timer event definition as well as each time a repeat timer is created.

tijsrademakers:
Thanks

PR #1024, opened 2016-09-15
Byte Array to String conversion fixes

suttermichael:
Hi, I found some occurrences where the Byte Array to String may lead to data loss. Therefore I changed the occurrences to "more robust" version from my point of view. Please have a look.

Cheers,
Michael


tijsrademakers:
Hi Michael, do you have some references that explain why Arrays.toString would be more robust?

suttermichael:
Hi Tijs,
I found that possible issues from a static code check executed against the Activiti sources. After that I (sadly saying) made only a quick search and thought that it will solve the issue.
Now I took some more time to dig into and found some more detailed explanations for it. The main issue is that in case you call String.getBytes from the former constructed String the returned byte array can differ in case the CharacterSet used in the JVM does not have a mapping for a character from the byte of the array. There are some recommendations like e.g. use new String(array, CharacterSet) or Base64 Encoding but the initial problem will most likely still exists.
With this knowledge I have the assumption that my PR will not solve the issue and most likely will make it worse. Therefore I'm closing the PR.
Sorry for the inconvenience,
Michael

PR #1045, opened 2016-10-12
fix incorrect "dataUrl" response property in REST api doc

jtulley:
The REST api doc incorrectly had the a property, "dataUrl" coming
back from calls to GET repository/deployments/{deploymentId}/resources

The actual property is "contentUrl".


balsarori:
thanks!

PR #1048, opened 2016-10-14
identitylinks user & group docs did not match implementation

jtulley:
The RestIdentityLink class, which is being used for serialization
in identity link creation and retrieval, has the properties,
"group", and "user".  The doc used "groupId" and "userId" instead.
Creating identity links using the exact doc sample syntax resulted
in the error,
"A group or a user is required to create an identity link."

Note: I left "groupId" and "userId" when it was part of a URL, for example:
"GET identity/users/{userId}"  since this makes more sense.  I could be
wrong in having left these.

### DIFF
--- a/userguide/src/en/ch14-REST.adoc
+++ b/userguide/src/en/ch14-REST.adoc
@@ -911,7 +911,7 @@ POST repository/process-definitions/{processDefinitionId}/identitylinks
 [source,json,linenums]
 ----
 {
-  "groupId" : "sales"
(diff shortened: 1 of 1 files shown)


jtulley:
Note:  The correct thing here MIGHT be to change the RestIdentityLink class.  I just realized that the IdentityLink.java interface DOES have getUserId, and getGroupId.
I wasn't sure about changing RestIdentityLink, since that COULD be a breaking change for some.

PR #1053, opened 2016-10-23
Remaining optimizations for 'or queries' with multiple process variable values

m4ng0:
I addressed the same problem of #989 for Process Instance, Task and Historic Task queries.


balsarori:
thanks

PR #1062, opened 2016-10-27
add activiti-rest and dependent modules to main modules list

jtulley:
this allows things like 'mvn install' to work, pushing rest api libraries
to the local maven repo.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -634,6 +634,11 @@
     <module>modules/activiti-image-generator</module>
     <module>modules/activiti-bpmn-converter</module>
     <module>modules/activiti-engine</module>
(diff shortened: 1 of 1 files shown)


jtulley:
This may or may not be controversial.  I was going to post in the forum about it, but figured I could just suggest the change and discuss here.  Is there any reason these modules (or the rest that I did not include) are not in the main modules list in the root level pom.xml?

vasiledirla:
@jtulley 
you can obtain the same result by using "distro" profile. "mvn install -Pdistro".

balsarori:
@jtulley as @vasiledirla mentioned, all modules can be build using ```mvn install -Pdistro```

PR #1063, opened 2016-10-27
Set process definition version on process instance in rest api

jtulley:
Use the management api to set the process definition version on a process instance in the rest api


balsarori:
@jtulley changing the process definition version of a running instance isn't supported.

jtulley:
Excuse my ignorance, please -- if the process definition is changed, what is the typical process to update that and get all the current process instance put back into the process definition in the correct place?  Maybe we are just doing something very non-standard by allowing the process definition to change.
Are we abusing the SetProcessDefinitionVersionCmd by doing this?

jtulley:
Ah, I see the command came from this request: https://forums.activiti.org/content/proposal-runtimeservicesetprocessdefinitionversionpi-pd

And there are good reasons to not make it "part of the public API".  (though, one could argue that it is, given it is available in the ManagementService).

balsarori:
yes it's something not public. Only public APIs are supported

jtulley:
Closing this request

PR #1077, opened 2016-12-05
Restoring exception log level to debug.

daisuke-yoshimoto:
We restored CommandContext's exception log level to debug.

If an exception occur in Command, the exception is thrown to the caller and the caller handled the exception.
In addition, CommandContext output debug log of the exception when CommandContext is closing.

By the following change, CommandContext output not debug log but error log.
https://github.com/Activiti/Activiti/commit/1f8fb8ad37dbd4cb4be77d827ba19e857bd451d1

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/interceptor/CommandContext.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/interceptor/CommandContext.java
@@ -168,7 +168,7 @@ public class CommandContext {
               // reduce log level, as normally we're not interested in logging this exception
               log.debug("Optimistic locking exception : " + exception);
             } else {
-              log.error("Error while closing command context", exception);
+              log.debug("Error while closing command context", exception);
             }
 
             transactionContext.rollback();
--- a/modules/activiti-engine/src/test/java/org/activiti/engine/test/logging/mdc/CommandContextLoggingTest.java
+++ b/modules/activiti-engine/src/test/java/org/activiti/engine/test/logging/mdc/CommandContextLoggingTest.java
@@ -1,0 +1,159 @@
+package org.activiti.engine.test.logging.mdc;
+
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+
+import org.activiti.engine.ActivitiObjectNotFoundException;
+import org.activiti.engine.ActivitiOptimisticLockingException;
+import org.activiti.engine.ActivitiTaskAlreadyClaimedException;
+import org.activiti.engine.JobNotFoundException;
+import org.activiti.engine.identity.User;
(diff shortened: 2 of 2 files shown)


daisuke-yoshimoto:
@balsarori @vasiledirla
If possible , please feedback.

vasiledirla:
@daisuke-yoshimoto 
The caller could miss the exceptions since these are unchecked exceptions and a method is not obliged to establish a policy for the unchecked exceptions.

having it logged as error inside the CommandContext will make sure it will not be lost even the caller is catching it (or not).

daisuke-yoshimoto:
@vasiledirla
Since updating Activiti version to 5.22.0, we are in trouble because many error logs(like ActivitiObjectNotFoundException) are inadvertently outputted by CommandContext.
https://github.com/Activiti/Activiti/commit/1f8fb8ad37dbd4cb4be77d827ba19e857bd451d1

If this fix assume only runtime exceptions(unchecked exceptions), we can understand this fix.
But,there are a lot of Activiti Exception which is leaking in this branch.
(ex) ActivitiObjectNotFoundException

balsarori:
@daisuke-yoshimoto 
ActivitiObjectNotFoundException and all Activiti exceptions are runtime exceptions. Base class for all Activiti exceptions extends RuntimeException https://github.com/Activiti/Activiti/blob/master/modules/activiti-engine/src/main/java/org/activiti/engine/ActivitiException.java#L21

daisuke-yoshimoto:
@balsarori @vasiledirla

Thank you for your reply.
I understood a policy for the unchecked exceptions.

But, this fix was implemented in Activiti 5.22.0 version.
Since Activiti 5.22.0 version, was a policy for the unchecked exceptions changed?

If possible , please tell me the circumstances of this fix.This fix has no issue ticket.
https://github.com/Activiti/Activiti/commit/1f8fb8ad37dbd4cb4be77d827ba19e857bd451d1

balsarori:
@daisuke-yoshimoto 
Some customers were complaining that they don't see any logs of exceptions and that exceptions are being swallowed.

For ```ActivitiObjectNotFoundException``` exceptions it might make sense to reduce the level to warn.

daisuke-yoshimoto:
@balsarori

Thank you for your reply.

> Some customers were complaining that they don't see any logs of exceptions and that exceptions are being swallowed.

I understood the circumstances of this fix.
But,usually not CommandContext but the caller must catch exceptions  and log error.
I understood your comment as Alfresco special circumstance.

balsarori:
@daisuke-yoshimoto 
I don't see that this applies to Alfresco special circumstance. And I do understand you special circumstance.

balsarori:
@daisuke-yoshimoto 
I'll be happy to modify it if more users complain about it.

daisuke-yoshimoto:
@balsarori
Thank you for your polite response.
I'd like to wait and see.

PR #1081, opened 2016-12-13
act-4102 trigger all compensation event for process

motorina0:
ACT-4102 For multiple executions paths, only the compensation(s) on the current execution was executed.
Added unit test.

More details in the comments of the Jira bug:
https://activiti.atlassian.net/projects/ACT/issues/ACT-4102

Signed-off-by: Vlad Stan stan.v.vlad@gmail.com


motorina0:
bug does not reproduce on v6

balsarori:
Thanks!

PR #1085, opened 2016-12-15
Fix for process instance created via signal from a suspended process …

HamdanMsheik:
Process instance cannot be created from suspended process definition using message or runtime methods. However, this was possible using signal and therefore is not correct. The pull request here is to fix that unwanted behavior.


balsarori:
Thanks!

carwashi:
This change made it into the 5.x branch but I do not see any maven release with this change.  Can a release be performed with this change?

daisuke-yoshimoto:
@carwashi Yeah. Sorry, Activiti 5.23 hasn't been released yet.
If you want to use this change now, you should build 5.x branch by yourself.
https://github.com/Activiti/Activiti/blob/5.x
https://github.com/Activiti/Activiti/blob/5.x/modules/activiti-engine/src/main/java/org/activiti/engine/impl/event/SignalEventHandler.java#L58

carwashi:
@daisuke-yoshimoto  I don't really want to  build and release on my own here as this is a huge breakdown of a normal users process.  When can we expect to see the 5.23 release?  I assume that just a maven release is needed

Issue #791, opened 2016-01-30
Process instance not ended after End Terminate Event with terminateAll 


zanottifio:
Ending the process with terminateAll doesn't update act_hi_procinst row.

As END_TIME_ , DURATION_ , END_ACT_ID_ , are null the process is reported not ended.

eugeneroz:
Also occured

jbarrez:
Jira case: https://activiti.atlassian.net/browse/ACT-4137

jbarrez:
Indeed, thanks for reporting! Added tests and fixed.

PR #1097, opened 2017-01-06
Subprocess(call activity) instance not ended after terminate end event with terminateAll

daisuke-yoshimoto:
Ending the process with terminateAll doesn't update act_hi_procinst row of subprocess(call activity) instance and activities of subprocess(call activity).

## Case1 Ending the process by parent process's terminate end event with terminateAll

The historic process instance of subprocess(call activity) doesn't end when parent process reached terminate end event.
The values of END_TIME_ column(act_hi_procinst) and DURATION_ column(act_hi_procinst) are null.

The same goes for the historic activity instance of subprocess(call activity).

---

## Case2 Ending the process by sub process's terminate end event with terminateAll

The historic process instance of subprocess(call activity) doesn't end when subprocess(call activity) itself reached terminate end event.
The values of END_TIME_ column(act_hi_procinst) and DURATION_ column(act_hi_procinst) are null.

The same goes for the historic activity instance of subprocess(call activity).

---

## The similar issue that fixed in activiti 5.20.

#791

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/TerminateEndEventActivityBehavior.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/TerminateEndEventActivityBehavior.java
@@ -14,10 +14,14 @@ package org.activiti.engine.impl.bpmn.behavior;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


daisuke-yoshimoto:
@balsarori
If possible , please feedback.

daisuke-yoshimoto:
@balsarori
If possible , please feedback.

PR #1107, opened 2017-02-14
Added taskAssigneeIds method,

purvendra-k:
It takes a list of users as parameter and returns tasks with an assignee in the list. Currently, TaskInfoQuery interface provides methods for querying or filtering tasks assigned to a specific user. If one would like to query for tasks assigned to several users, he has to query for individual users and then merge them to create the complete list. This method solves this problem.
Similar method "taskCandidateGroupIn" for querying tasks of groups is already there in activiti. This will add similar functionality for querying tasks of users as well.


purvendra-k:
Similar method "taskCandidateGroupIn" for querying tasks of groups is already there in activiti. So, no need for a new implementation.

PR #1115, opened 2017-02-21
Add process variables in formService.submitStartFormData()

deanwong:
If use expression in start form , like this:
```
<startEvent id="start">
      <extensionElements>
        <activiti:formProperty id="name" expression="#{speaker.name}"/>
        <activiti:formProperty id="birthday" datePattern="yyyy-MM-dd" expression="#{speaker.birthday}" type="date"/>
      </extensionElements>
</startEvent>
```
The existing interface  
`formService.submitStartFormData(String processDefinitionId, String businessKey, Map<String, String> properties);` 
do not provide the parameter to pass initial process variables. Just enhance this point. Test case submitted as well.


balsarori:
Thanks for the pull request. I don't understand the use case here. Why are both form properties (which will be converted to variables) and process variables needed? If you want to start a process with process variables then use ```runtimeService.startProcessInstanceXXX``` methods.

deanwong:
Hi @balsarori 
If I use `runtimeService.startProcessInstanceXXX` to start the process instance,  the process will jump over the start form directly. How can I get user's form request and pass them into process variable to let next steps to use?

1289100276:
form  can solve complexity business? you can use jquery instead of activiti form.

balsarori:
@deanwong can you resolve the conflicts?

deanwong:
@balsarori  I think you have discarded the org.activiti.engine.FormService now ?

balsarori:
@deanwong yes it's been removed now. There is a new Form engine so it's no longer needed. Thanks!

PR #1133, opened 2017-03-09
Fix for ACT-4263

gdharley:


erdemedeiros:
Hi @gdharley ,

Thanks for providing this PR. May I ask you to provide a test for this fix as well?
In addition, could you, please, use the same indentation policy as the current code?

Best Regards,
Elias

modules/activiti-engine/src/main/java/org/activiti/engine/impl/pvm/runtime/AtomicOperationTransitionDestroyScope.java
```diff
@@ -89,6 +89,14 @@ public void execute(InterpretableExecution execution) {
 
         // TODO!
         execution.destroy();
+		
+		// GDH - fix for ACT-4263
+		List<JobEntity> jobs = (List) Context.getCommandContext()
```
erdemedeiros: You can use Lis<Job> here to avoid that cast.

balsarori:
Thanks

PR #1166, opened 2017-04-11
reduced db queries

mike-jordan:


Salaboy:
@balsarori are we looking into merging this? Should it be rebased to 6.x instead of activiti6? Is this something that we should take in develop as well?

PR #1183, opened 2017-04-26
Fix for NEP when process start_user_id field is null

ravitadela:
Hi, 

When a process is started by timer or an event, the instantiation user will be null and hence a null pointer exception is being thrown which is handled with this fix


baocaixue:
Hi, Any API for set "start_user_id"?

PR #1186, opened 2017-05-01
Removed duplicates in Base Element, having Extension Attribute and Ex…

fmcejudo:
…tension Element a common interface Extension to grab common properties has been created

Created a JUnit test for Base Element which tests adding extension elements and attributes, as well as cloning a Base Element

Extension attributes were not cloned in Extension Elements. Issue fixed up.


Salaboy:
@fmcejudo hi there. We appreciate your PR, but can you please rebase it branch develop? I will close this PR until we get the one which points to the develop we cannot review and merge.

PR #1192, opened 2017-05-09
added missing backslash to escape _ character

martinhanes:
Fixes issue#1191.


balsarori:
https://activiti.atlassian.net/browse/ACT-4287

PR #1215, opened 2017-06-07
(refactor) some opportunities to use the Diamond-operator

desmorto:


Salaboy:
@desmorto this is ok, but it is not adding much in my opinion. If you want us to include this in the latest code please rebase the PR to develop

PR #1225, opened 2017-06-30
Add tests for process definition and process instances rest endpoints

erdemedeiros:


activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/rest/SimpleRestTemplateBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.boot.web.client.RestTemplateBuilder;
+import org.springframework.hateoas.hal.Jackson2HalModule;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SimpleRestTemplateBuilder {
+
+    public TestRestTemplate build(String rootUri) {
```
Salaboy: Instead of creating a type here (a new class) I would rather do a @Bean injection like the example that I share previously. So you can auto wire the TestRestTemplate. We need to make sure that the RestTemplate itself is auto wired, because as we find out there is some extra magic injected to the object and the request headers for integrating with other frameworks.

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/runtime/ProcessInstanceRestTemplate.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.activiti.client.model.ExtendedProcessInstance;
+import org.activiti.client.model.ProcessInstance;
+import org.springframework.boot.test.web.client.TestRestTemplate;
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+
+public class ProcessInstanceRestTemplate {
```
Salaboy: Do we really need all these types? I know that the tests might look bigger, but what's the point of having all these wrappers if we can show simple tests doing pure requests to the service as any client will do. If we start adding all these types we will leak business logic to the clients and we should try to avoid doing that as much as possible. Wdyt?

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/BaseRestIT.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public abstract class BaseRestIT {
+
+    @Autowired
+    private SimpleRestTemplateBuilder simpleRestTemplateBuilder;
+
+    private TestRestTemplate restTemplate;
```
Salaboy: This is the one that needs to be Autowired.. so instead of having a type we just need a "Producer" @Bean method that uses the builder to provide a new instance and contains the configuration

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/Application.java
```diff
@@ -0,0 +1,40 @@
...
+
+@SpringBootApplication
+public class Application {
+
+	public static void main(String[] args) {
+		SpringApplication.run(Application.class, args);
+	}
+
+    @Bean
+    InitializingBean usersAndGroupsInitializer(final IdentityService identityService) {
```
Salaboy: I look forward to have these removed from here. We need to start thinking about how to remove the identityService and interact with an external module for having that information.

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/rest/SimpleRestTemplateBuilder.java
erdemedeiros: Done

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/runtime/ProcessInstanceRestTemplate.java
erdemedeiros: I see your point, but as the start process might be used in several tests (ProcessInstanceIT and ProcessInstanceVariablesIT, for the moment, but also tasks in the future), I prefere to centralize the REST call here, for the moment.

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/BaseRestIT.java
erdemedeiros: Updated. Moved the TestRestTemplate instantiation inside a configuration class.

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-hal-rest-api/src/test/java/org/activiti/Application.java
erdemedeiros: I removed it for the moment, but we might get blocked when writing task related tests.

PR #1228, opened 2017-07-03
Form service removal

balsarori:


activiti-rest/src/test/java/org/activiti/rest/service/api/history/HistoricDetailQueryResourceTest.java
```diff
@@ -75,11 +74,6 @@ public void testQueryDetail() throws Exception {
 
     requestNode = objectMapper.createObjectNode();
     requestNode.put("processInstanceId", processInstance2.getId());
-    requestNode.put("selectOnlyFormProperties", true);
-    assertResultsPresentInDataResponse(url, requestNode, 0, null, null);
-
-    requestNode = objectMapper.createObjectNode();
```
Salaboy: Does this need to be removed? Is that related to the form?

balsarori: yes, it used to test when only form properties results needs to be returned.

ClementWang:
@balsarori, hi, I want to use the form service in activiti 7.0, I noticed that it was removed in activiti 7.0. For what purpose, delete it? In previous versions, when we started the process instance, we could use the form service to save some form data. How should we save the form data now, is there some documentation or reference examples?

Elvynia:
My project has spring boot 2 so I'm forced to use Activiti 7, but the goal is to handle processes using UserTask with form data. I'm like you @ClementWang, did you find any other method to retrieve form properties ?
@balsarori @salaboy I can see FormProperty and FormValue classes but nothing to use it through UserTask or else. The task I get from the taskService gives me a TaskEntityImpl but no UserTask. I can't find any solution after hours of searching, please help. Is the only way getting back to spring boot 1.x and Activiti 6 ?

fastcodepig:
You could try Flowable. That what we did when we hit these problems.  Not saying its the answer but worked for us .

Elvynia:
I almost tried, but I got a working version using Spring Boot 1.x and Activiti 6 with FormService. From there I got it working with Spring Boot 2 dans more manual configuration (no activiti starter dependency). Thanks for answering.

mteodori:
@Elvynia if you are using activiti cloud and the runtime bundle service doing a POST /rb/v1/tasks/{taskId}/complete with variables in the payload would work or in activiti core calling TaskService.complete. The original design for removal was to focus activiti on just process and let developers choose what to do for application related concepts like forms. I see some form related classes for modeling have been left there though, so that a custom app can eventually use them, but probably this is even more confusing and they should rather be deleted as well. Not sure if this helps, if you can share more of your app and how you are deploying you can help us understand better what option can work for you and put this in a FAQ.

Elvynia:
@mteodori Thanks but I'm not using Activiti Cloud sadly. I don't need the task complete part though. I need to retrieve form properties in java from user tasks in the process to display generic forms. It seems like things are going all cloud and REST API, really different from BPMN before, so I'm not surprised there is no actual solution in Activiti 7 for that (form data). But it stunned me when I landed on the issue stating that's it definitely removed (specially because of the remaining classes like you said), I wasn't prepared for that ^^"
Anyway it works well with manual config, Spring Boot 2 and Activiti 6. I don't have control over the repository but if you guys need a working example with Spring Boot 2 I can upload the config here.

mteodori:
not really, feature of cloud are in core as well via Java API rather than REST/messaging - how do you display the form? are you using ADF, the webapp or a custom form display implementation? is there anything for which you need the old Form API that you cannot do via the Task API? if you can create a project in your github that would help

PR #1232, opened 2017-07-06
Removed distinct keyword from query

anoopvrma:
Tested query with and without distinct keyword with multiple scenarios. In both the cases response of query is same. So distinct keyword is not required, it will just create overhead of comparing rows.


balsarori:
without the distinct keyword duplicate results are returned.

PR #1246, opened 2017-07-14
Tasks specific to user

ryandawsonuk:
Tasks endpoint now gives tasks for which user is candidate or assignee or candidate via their groups. The user 'hruser' is a member of group 'hr' and so can see the task with this as candidateGroup. I breakpointed the KeycloakUserGroupProxy and checked to check that it was getting the groups correctly. I've a concern that when I invoke TasksKeyCloakIT it invokes the proxy 7-8 times. I am not sure why - could be something to do with paging or something at engine level. I'm also not sure whether keycloak does any caching. I suspect that keycloak only caches tokens and not queries so we'd probably have to cache ourselves if we wanted to. Perhaps we can discuss?


Salaboy:
Yes let's have a discussion about this

daisuke-yoshimoto:
@ryandawsonuk @Salaboy 

I think the cache of MyBatis was working within the Command once when IDM was on the engine table.
>  I am not sure why - could be something to do with paging or something at engine level.

Salaboy:
@daisuke-yoshimoto thanks for your comment, that might be one of the reasons why the interaction with IDM is heavily querying keycloak, we will take a look at it on monday.

ryandawsonuk:
Yes the reason is as @daisuke-yoshimoto says - it's because the engine looks up the groups from within a getter that is invoked at the MyBatis level and that getter gets called multiple times.

I've limited it to one call by using the alternative option of doing the proxying at the Controller level and passing the results down. The engine level supports both approaches but this way is faster. It also makes it possible to mock the group lookups from within the starter project so I've added some tests at starter level that make use of group lookup and user assignment.

PR #1263, opened 2017-07-20
Maven cleanup + automation of release tasks

slanglois:
Maven pom cleanup:
 - Removing the parent pom
 - Adding / upgrading project metadata
 - Upgrading Maven plugins
 - Removing useless `distro` profile

Automating releases:
 - include distribution URLs (snapshot and release)
 - configuration of the maven-release-plugin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: java 
 jdk:
   - oraclejdk8
-env:
-  - MAVEN_OPTS=-Xmx1536m
 notifications:
   webhooks:
     urls:
@@ -11,6 +9,8 @@ notifications:
     on_failure: always  # options: [always|never|change] default: always
     on_start: never     # options: [always|never|change] default: always
 
+before_script:
+  - echo MAVEN_OPTS=-Xmx1536m > ~/.mavenrc
 # Enable integration tests
 script:
   - mvn verify -B
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: java 
 jdk:
   - oraclejdk8
+env:
+  - MAVEN_OPTS=-Xmx1536m
 notifications:
   webhooks:
     urls:
--- a/activiti-bpmn-converter/pom.xml
+++ b/activiti-bpmn-converter/pom.xml
@@ -65,28 +65,4 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
-  <profiles>
-    <profile>
-      <id>distro</id>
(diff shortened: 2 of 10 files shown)


slanglois:
The build keeps failing in OutOfMemory on Travis, always on the same test:
```
[INFO] Running org.activiti.engine.test.bpmn.gateway.InclusiveGatewayDefaultFlowTest
Exception in thread "Thread-8" java.lang.OutOfMemoryError: GC overhead limit exceeded
	at java.util.Arrays.copyOf(Arrays.java:3332)
	at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:124)
	at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:596)
	at java.lang.StringBuffer.append(StringBuffer.java:367)
	at java.io.BufferedReader.readLine(BufferedReader.java:370)
	at java.io.BufferedReader.readLine(BufferedReader.java:389)
	at org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.StreamPumper.run(StreamPumper.java:76)
Exception in thread "ThreadedStreamConsumer" java.lang.OutOfMemoryError: GC overhead limit exceeded
```
I've tried setting `-Xmx256m` on the forked JVM, but I get the same error. 
I updated the `maven-surefire-plugin` so it may behave a bit differently, but hopefully not that badly...
I may need your help to understand what's going on there.

daisuke-yoshimoto:
@slanglois 
I think you may be able to set JVM parameter of Junit process in .travis.yml.
``
-DsurefireArgLine="-Xmx256m"
``

But, you need to restore the following change.
https://github.com/Activiti/Activiti/blob/master/pom.xml#L557
https://github.com/Activiti/Activiti/pull/1263/files#diff-600376dffeb79835ede4a0b285078036L557

slanglois:
Thanks @daisuke-yoshimoto 
I'd rather not bring this change back... It's not standard and I find it clumsy.
The standard way to specify this is through the `argLine` property, like in d5322d01a21217d1802a6ae06c97c76d5dc4aacb
Moreover, I'd rather do this change in the POM rather than `.travis.yml`, so that everyone benefits from it.

Anyway, the memory problem was actually not in the forked JVM, which runs the tests, but in the one running Maven itself.
I fixed this by setting the `MAVEN_OPTS` env variable on Travis.
It's all good now :-)

erdemedeiros:
Great job! It seems we were missing some coverage reports before (5.75% more).

Issue #1245, opened 2017-07-14
Run external depedency (e.g. keycloak) for tests inside docker


ryandawsonuk:
Starting point - https://docs.travis-ci.com/user/docker/

Next step after proving it is to provide a reference a docker image with the keycloak json files applied (the json files here - https://github.com/Activiti/Activiti/tree/9308ca29fb2c7033a945f03ba961165bb9f6f62a/activiti-spring-boot/spring-boot-samples/spring-boot-sample-hal-rest-api)... See https://github.com/dfranssen/docker-keycloak-import-realm and https://github.com/keycloak/keycloak-test-docker-images/tree/master/keycloak-sssd-integration-tests for info on how to do this.

ryandawsonuk:
Enzo found this - https://github.com/spotify/dockerfile-maven

ryandawsonuk:
We now do have a reference docker file that has the json configuration applied to it and which the tests pass against - https://github.com/Activiti/Activiti/tree/tasks-specific-to-user/activiti-spring-boot/spring-boot-samples/spring-boot-sample-hal-rest-api  (note this isn't in master branch at time of writing). 

So the outstanding bit is this bit to initiate it from maven and invoke the tests against it from maven (currently maven set to exclude any tests with 'Keycloak' in the name).

Salaboy:
@ryandawsonuk there are 3 options for maven plugins the one provided by spotify, the one provided by fabric8 and another one provided by an individual. There is also a matrix provided by fabric8 comparing the 3 options. I should go with the one provided by spotify unless the fabric8 one provides too many extra advantages.

PR #1286, opened 2017-07-25
include keycloak-enabled tests in travis build

ryandawsonuk:
For #1245
Have moved keycloak-enabled tests out of activiti-services-identity-keycloak as they are integration tests that need a full application running. They are now in the starter. This is ok as starter dependencies can be overridden with maven if alternative IDM is used. Have accordingly updated the starter tests to be keycloak-enabled. Also updated the README files to fit changes to project structure.


activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/pom.xml
```diff
@@ -131,5 +136,25 @@
...
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>${maven-surefire-plugin.version}</version>
+        <configuration>
+          <redirectTestOutputToFile>true</redirectTestOutputToFile>
+          <failIfNoTests>false</failIfNoTests>
+          <trimStackTrace>false</trimStackTrace>
+          <excludes>
+            <exclude>**/*TestCase.java</exclude>
+            <exclude>**/*IT.java</exclude>
+            <!--no longer exclude keycloak as travis should be running it
```
Salaboy: @ryandawsonuk what's the point of having the comments here?

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/src/test/java/org/activiti/cmdendpoint/CommandEndpointKeycloakIT.java
```diff
@@ -48,9 +50,9 @@
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @TestPropertySource("classpath:application-test.properties")
-@DirtiesContext
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_CLASS)
 @EnableBinding(MessageClientStream.class)
-public class CommandEndpointTest {
+public class CommandEndpointKeycloakIT extends KeycloakEnabledBaseTestIT {
```
Salaboy: @ryandawsonuk I;m worried about having all the tests called KeycloakIT now... there is no need if we assume that keycloak will be there always right?

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/src/test/java/org/activiti/definition/ProcessDefinitionKeycloakIT.java
```diff
@@ -35,7 +37,8 @@
 
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-public class ProcessDefinitionIT {
+@TestPropertySource("classpath:application-test.properties")
+public class ProcessDefinitionKeycloakIT extends KeycloakEnabledBaseTestIT{
```
Salaboy: same here

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/src/test/java/org/activiti/keycloak/KeycloakEnabledBaseTestIT.java
```diff
@@ -55,11 +55,11 @@ protected AccessTokenResponse authenticateUser() throws IOException {
 
     protected HttpHeaders getHeaders(String token) {
         HttpHeaders headers = new HttpHeaders();
-        headers.add("Authorization", "Bearer " + new String(token));
+        headers.add("Authorization", "Bearer " + token);
```
Salaboy: @ryandawsonuk just curious... is this a normal practice... I would assume that spring will inject this for us.

pom.xml
```diff
@@ -631,7 +631,8 @@
           <excludes>
             <exclude>**/*TestCase.java</exclude>
             <exclude>**/*IT.java</exclude>
-            <exclude>**/*Keycloak*.java</exclude>
+            <!--no longer exclude keycloak as travis should be running it
```
Salaboy: @ryandawsonuk same.. remove the comments.. they are dead code that we don't want to maintain

pom.xml
```diff
@@ -869,11 +870,12 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-failsafe-plugin</artifactId>
           <version>${maven-failsafe-plugin.version}</version>
+          <!--no longer exclude keycloak as travis should be running it
```
Salaboy: @ryandawsonuk same here

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/src/test/java/org/activiti/keycloak/KeycloakEnabledBaseTestIT.java
ryandawsonuk: If it were always the same token then I could add it to the template but when we pass keycloak auth we then get a temporary token which we use for the life of that particular test. In theory if we had a really long-running test we'd have to call authenticateUser() again and get a new token (which would be simple to do as the methods are there in the base test).

ryandawsonuk: Have added comments to the base test to explain this now

PR #1289, opened 2017-07-26
Salaboy 1284 api versioning

Salaboy:
Just adding simple versioning to HAL API endpoints (v1)


Salaboy:
@erdemedeiros @ryandawsonuk please check.. not a big change but something that we need now.

Issue #1308, opened 2017-07-29
JSON/XML endpoints for Process Definitions


daisuke-yoshimoto:
Provide the following endpoints

/process-definitions/{id}/json
/process-definitions/{id}/xml

These endpoints should return the corresponding model in each of these formats.

PR #1319, opened 2017-08-01
JSON/XML endpoints for Process Definitions #1308

daisuke-yoshimoto:
Refer to https://github.com/Activiti/Activiti/issues/1308


activiti-services/activiti-services-core-model/src/main/java/org/activiti/services/core/model/MetaBpmnModel.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.activiti.bpmn.model.BpmnModel;
+import org.activiti.services.core.model.json.converter.BpmnModelDeserializer;
+import org.activiti.services.core.model.json.converter.BpmnModelSerializer;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MetaBpmnModel {
```
Salaboy: @daisuke-yoshimoto this name is confusing... We have MetaBpmnModel, BpmnModel and ProcessModel. We need to unify all under just one model. BpmnModel make sense, but I don't know if we need that type anyway. It sounds to me that if we have something called BpmnModel just to represent the content (xml or json) as a string will be enough. So instead of having Serializers and Deserializers for our type we just have serialization and deserialization to a string.

activiti-services/activiti-services-rest/src/main/java/org/activiti/services/rest/controllers/ProcessDefinitionController.java
```diff
@@ -73,4 +95,23 @@ public ProcessDefinitionResource getProcessDefinition(@PathVariable String id) {
         }
         return resourceAssembler.toResource(processDefinitionConverter.from(processDefinition));
     }
+
+    @RequestMapping(value = "/{id}/xml",
+            method = RequestMethod.GET)
+    public ProcessModelResource getProcessModel(@PathVariable String id) {
+        try (final InputStream resourceStream = repositoryService.getProcessModel(id)) {
+            String xml = new String(IoUtil.readInputStream(resourceStream, null), StandardCharsets.UTF_8);
```
Salaboy: @daisuke-yoshimoto The endpoint should return the XML string, at least that is what I would expect as a client of this service. We don't need to return a resource. @erdemedeiros what do you think about that?

activiti-services/activiti-services-rest/src/main/java/org/activiti/services/rest/controllers/ProcessDefinitionController.java
```diff
@@ -73,4 +95,23 @@ public ProcessDefinitionResource getProcessDefinition(@PathVariable String id) {
...
+            throw new ActivitiException("Error occured while getting process model '" + id + "' : " + e.getMessage(),
+                                        e);
+        }
+    }
+
+    @RequestMapping(value = "/{id}/json",
+            method = RequestMethod.GET)
+    public BpmnModelResource getBpmnModel(@PathVariable String id) {
+        BpmnModel bpmnModel = repositoryService.getBpmnModel(id);
+        return bpmnModelResourceAssembler.toResource(new MetaBpmnModel(id, bpmnModel));
```
Salaboy: @daisuke-yoshimoto same here, just return the json content of the process definition. In the same way that we are going to return the svg content when we have it.

activiti-services/activiti-services-rest/src/main/java/org/activiti/services/rest/controllers/ProcessDefinitionController.java
Salaboy: We need to change the produces property of @RequestMapping to xml

activiti-services/activiti-services-rest/src/main/java/org/activiti/services/rest/resources/BpmnModelResource.java
```diff
@@ -0,0 +1,15 @@
+package org.activiti.services.rest.resources;
+
+import org.activiti.services.core.model.MetaBpmnModel;
+import org.springframework.hateoas.Link;
+import org.springframework.hateoas.Resource;
+
+public class BpmnModelResource extends Resource<MetaBpmnModel> {
```
Salaboy: @daisuke-yoshimoto I think that we don't need these resources. Unless you find a clear advantage of this approach I would recommend to not have this extra layer of complexity here.

activiti-services/activiti-services-rest/src/main/java/org/activiti/services/rest/resources/assembler/BpmnModelResourceAssembler.java
```diff
@@ -0,0 +1,30 @@
...
+@Component
+public class BpmnModelResourceAssembler extends ResourceAssemblerSupport<MetaBpmnModel, BpmnModelResource> {
+
+    public BpmnModelResourceAssembler() {
+        super(ProcessDefinitionController.class, BpmnModelResource.class);
+    }
+
+    @Override
+    public BpmnModelResource toResource(MetaBpmnModel model) {
+        Link selfRel = linkTo(methodOn(ProcessDefinitionController.class).getBpmnModel(model.getId())).withSelfRel();
```
Salaboy: @daisuke-yoshimoto same here.. no need for the assemblers if we don't have the resources. We should keep it as simple as possible.

daisuke-yoshimoto:
@Salaboy 
I fixed for pointed out issues to return directly JSON/XML string instead of resources.
Can you make a review again?

Issue #1299, opened 2017-07-28
Provide Docker compose to start dev environment.


Salaboy:
Provide docker-compose.yml so developers can start all the necessary third party services to run against the new process engine APIs

PR #1322, opened 2017-08-01
docker-compose to run application #1299

ryandawsonuk:
Actually there's a start.sh that builds the images through maven and then starts using docker-compose. The README explains.

Note I have changed the sample app to run on 8090 in order to free up 8080 for keycloak.


Salaboy:
@ryandawsonuk please create an issue to define the port strategy

Issue #1260, opened 2017-07-19
Migration tool to export users and groups from tables and put in the keycloak


ryandawsonuk:
In v7 we've removed User, Group, Membership and IdentityInfo from the core Engine and instead provided an example of integrating with an external identity provider (keycloak - https://github.com/Activiti/Activiti/pull/1246). We're seeing that as the default path now.

This means that any data that Activiti6 users have in their identity tables should be migrated. For the table definitions refer to the identity table creation scripts in https://github.com/Activiti/Activiti/tree/6.0-release/modules/activiti-engine/src/main/resources/org/activiti/db/create. For the entities see also https://github.com/Activiti/Activiti/tree/6.0-release/modules/activiti-engine/src/main/java/org/activiti/engine/identity and https://github.com/Activiti/Activiti/tree/6.0-release/modules/activiti-engine/src/main/resources/org/activiti/db/mapping/entity .

Keycloak provides a way to migrate users and an example of doing so - https://github.com/Smartling/keycloak-user-migration-provider . But that's not necessary for our purposes - I think there is an easier way.

What we could do is to add a migration tool to the keycloak integration sample. What it would need to do would be to read the identity information from the legacy tables (they're not referenced anywhere in that project right now) - https://github.com/Activiti/Activiti/tree/master/activiti-spring-boot/spring-boot-samples/spring-boot-sample-hal-rest-api . I see this as a class with a main method that would connect to the engine DB, read from the legacy tables if they exist, connect to keycloak using the admin API (which we do already here - https://github.com/Activiti/Activiti/blob/master/activiti-spring-boot/spring-boot-samples/spring-boot-sample-hal-rest-api/src/main/java/org/activiti/keycloak/KeycloakUserGroupLookupProxy.java#L51) and add Users, Groups, user info and memberships that are not already in the keycloak system. See https://technology.first8.nl/programmatically-adding-users-in-keycloak/

After the tool runs you can verify that the users and groups have been added by checking in the keycloak admin console (or by querying using the API).

PR #1323, opened 2017-08-01
Migration tool:-

abdotalaat:
to export users and groups from tables and put in the keycloak #1260


Salaboy:
@abdotalaat hi there, great stuff. This is definitly the first step on getting this done, but I'm wondering if this is going to really work with Activiti 6 and Activiti 5. Have you tested it against an existing DB? I notice that you are using JPA to access those entities that you have defined, are those correctly mapped to the existing tables of the old versions? 

In any case we need tests that demonstrate that the code does what it is supposed to do. I would recommend you to add the activiti6 engine as a dependency to your project and use the Identity Service to create new users, groups and memberships and then run your code to see if the correct keycloak information is properly migrated. 

What do you think?

ryandawsonuk:
I notice you've used Roles instead of Groups in what you put into Keycloak in MigrateUsersGroupsToKeycloackApplication. I think this would create a mismatch as KeycloakUserGroupLookupProxy uses Groups (https://github.com/Activiti/Activiti/blob/master/activiti-services/activiti-services-identity-keycloak/src/main/java/org/activiti/services/identity/keycloak/KeycloakUserGroupLookupProxy.java). Groups and Roles are very similar so I expect it will be easy to change.

As Salaboy suggests, you could add the v5 or v6 engine as a dependency on the project (just as per the quickstart guide https://www.activiti.org/quick-start). Actually you could do that and then read the users and groups straight out of the old engine's identityService. Then there's no need even to define the classes and map them to tables as the old engine already does that and has a way to query them. Sorry we didn't think of this suggestion before as actually it should be a lot simpler.

abdotalaat:
@Salaboy  one of my todo to add test case for this.

@Salaboy  i tried it using activiti 6 i create simple application and i used the identityservice to create group and user .`RepositoryService repositoryService = processEngine.getRepositoryService();
		IdentityService identityService = processEngine.getIdentityService();
		
		String userName = "employeeName"+(new Date().getTime());
		User user = identityService.newUser(userName);
		user.setEmail(userName+"@gmail.com");
		user.setFirstName("21212");
		user.setLastName("1212");
		user.setPassword("1234");
		identityService.saveUser(user);
		
		
		String userGroup = "group";
		Group group = identityService.newGroup(userGroup);
		group.setName("test");
		group.setId("id"+(new Date().getTime()));
		identityService.saveGroup(group);
		
		
		identityService.createMembership(userName, group.getId());`



@Salaboy @ryandawsonuk  okay i will add the activiti6 as dependencies and use it directly to access users and groups.


 @ryandawsonuk i get confused yasterday about this "Roles and Groups" . but from my understand i said that group in Activiti it's like Role in keyClock but no issue i will change it.

ryandawsonuk:
I am sure this would work and allow users and groups to be migrated. Could you add a junit test for the logic inside the execute method of MigrateUsersAndGroupsService? I am guessing from your comment above that you did already test this as you clearly wrote code to create users and groups in the v6 engine. We'd just like to have that test as part of the pull request. You will need to have a keycloak instance available for the test. You can do that by adding the keycloak image using the docker maven plugin - https://github.com/Activiti/Activiti/blob/master/activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/pom.xml#L155 . I don't think you'll need any further configuration but you will want to use a different keycloak user (see - https://github.com/Activiti/Activiti/blob/master/activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/src/test/resources/application.properties). I'm guessing you tested with a standalone keycloak?

abdotalaat:
@ryandawsonuk  yes i am using standalone keycloack, okay let me write the junit test and use the keyclock docker image.

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/docker/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM jboss/keycloak:3.2.0.Final
+MAINTAINER https://gitter.im/Activiti/Activiti7
```
ryandawsonuk: Don't think this file or directory is needed. The main.basedir is located using the directory maven plugin which should find the root level. So the test can just refer to the docker directory under root. See https://github.com/Activiti/Activiti/blob/master/activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/pom.xml#L184 - that project doesn't have a docker subfolder, it refers to the one under root.

It looks like you've created a separate docker image just for these tests. But it's ok to re-use the existing image.

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/docker/README.md
```diff
@@ -0,0 +1,7 @@
+# Docker containers for Activiti
+
```
ryandawsonuk: See above - this directory not needed.

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/pom.xml
```diff
@@ -0,0 +1,195 @@
...
+						<configuration>
+							<property>main.basedir</property>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
+				<groupId>io.fabric8</groupId>
+				<artifactId>docker-maven-plugin</artifactId>
+				<version>${docker.maven.plugin.version}</version>
```
ryandawsonuk: You don't need to specify a version as it should be inherited.

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/pom.xml
```diff
@@ -0,0 +1,195 @@
...
+				<configuration>
+					<images>
+						<image>
+							<name>testkeycloack</name>
```
ryandawsonuk: You can use the same image name in https://github.com/Activiti/Activiti/blob/master/activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/pom.xml#L181 . That project also runs tests against a docker keycloak container. You should be able to just copy-paste the config from that project.

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/pom.xml
```diff
@@ -0,0 +1,195 @@
...
+							</build>
+							<run>
+								<namingStrategy>none</namingStrategy>
+								<ports>
+									<port>8180:8080</port>
+									<port>9090:9090</port>
+								</ports>
+								<wait>
+									<http>
+										<url>http://localhost:8180/auth</url>
```
ryandawsonuk: The other project does this using a wait for a logging line rather than a url but both work https://github.com/Activiti/Activiti/blob/master/activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/pom.xml#L195

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/src/main/java/org/activiti/tasks/MigrateUsersAndGroupsService.java
```diff
@@ -0,0 +1,15 @@
+package org.activiti.tasks;
+
+import org.activiti.engine.delegate.DelegateExecution;
+import org.activiti.engine.delegate.JavaDelegate;
+import org.springframework.stereotype.Component;
+
+@Component
```
ryandawsonuk: Presumably this class can be removed? Is it just a hangup from the previous implementation or is it necessary to use the engine in the test?

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/src/main/resources/processes/MigrateUserAndGroups.bpmn
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<definitions xmlns="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:activiti="http://activiti.org/bpmn" xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI" xmlns:omgdc="http://www.omg.org/spec/DD/20100524/DC" xmlns:omgdi="http://www.omg.org/spec/DD/20100524/DI" typeLanguage="http://www.w3.org/2001/XMLSchema" expressionLanguage="http://www.w3.org/1999/XPath" targetNamespace="http://www.activiti.org/test">
+  <process id="migration" name="My process" isExecutable="true">
```
ryandawsonuk: Presumably this can be removed? Is it just a hangup from the previous implementation or is it necessary to use the engine in the test?

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/src/test/java/org/activiti/tasks/MigrateUsersAndGroupsServiceTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Test
+    public void duplicatedUse() {
+
+        Response response = keycloakUtil.createUser("developer");
+        assertThat(409,is(response.getStatus()));
+        response.close();
+
+    }
```
ryandawsonuk: Presumably the method getAllGroupsSize in KeycloakUtil was written so that it could be used here?

abdotalaat: done

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/src/main/java/org/activiti/tasks/MigrateUsersAndGroupsService.java
abdotalaat: removed

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/src/main/resources/processes/MigrateUserAndGroups.bpmn
abdotalaat: it's mandatory to all spring boot initialize the activiti engine. i removed it and reuse "SimpleProcess.bpmn20.xml" .

activiti-spring-boot/spring-boot-samples/spring-boot-sample-migrate-users-groups-to-keycloack/pom.xml
abdotalaat: i changed it to follow same technique

PR #1360, opened 2017-08-07
zuul and eureka in docker-compose 1339

ryandawsonuk:
When reviewing this PR please also look at https://github.com/Activiti/activiti-examples/pull/4/ which updates the sample app


activiti-services/activiti-services-registry/pom.xml
```diff
@@ -33,6 +33,16 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-web</artifactId>
+      <exclusions>
```
Salaboy: @ryandawsonuk why do we need jetty here? I mean, if we don't have the keycloack adapter

activiti-services/activiti-services-audit/activiti-services-audit-jpa/src/main/resources/application.properties
```diff
@@ -22,4 +22,7 @@ keycloakclientpassword=client
 # this user needs to have the realm management roles assigned
 # one option to deal with password being here would be spring-boot-jasypt, keycloak may support others (can it be set as env variable?)
 spring.rabbitmq.host=rabbitmq
-spring.rabbitmq.port=5672
\ No newline at end of file
+spring.rabbitmq.port=5672
```
Salaboy: @ryandawsonuk as I mention today, do we really need this? I would try to avoid it.

activiti-services/activiti-services-gateway/src/main/resources/application.properties
```diff
@@ -1,3 +1,23 @@
 server.port=8080
 spring.application.name=gateway
-management.security.enabled=false
\ No newline at end of file
+management.security.enabled=false
+
+spring.rabbitmq.host=rabbitmq
+spring.rabbitmq.port=5672
+eureka.client.serviceUrl.defaultZone=http://activiti-services-registry:8761/eureka/
+hystrix.command.default.execution.timeout.enabled=false
```
Salaboy: @ryandawsonuk I would avoid tuning the hystrix configuration here.

activiti-services/activiti-services-identity-keycloak/src/main/java/org/activiti/services/identity/keycloak/SecurityConfig.java
```diff
@@ -74,6 +74,6 @@ protected void configure(HttpSecurity http) throws Exception {
         super.configure(http);
         http.authorizeRequests()
                 .antMatchers("/*").hasRole("user") //could optionally set these by autowiring and parsing keycloakSpringBootProperties
```
Salaboy: @ryandawsonuk here we are adding another place to configure roles and what they are allowed to do?

docker/docker-compose.yml
```diff
@@ -1,21 +1,32 @@
...
-    - "8082:8082"
-  links:
-    - activiti-keycloak:activiti-keycloak
-    - rabbitmq:rabbitmq
-#elasticsearch:
-#  image: docker.elastic.co/elasticsearch/elasticsearch:5.4.1
-#  ports:
-#    - "9200:9200"
-#    - "9300:9300"
+version: '2'
```
Salaboy: @ryandawsonuk :+1

docker/keycloak/springboot-realm.json
```diff
@@ -381,7 +381,7 @@
     "enabled" : true,
     "clientAuthenticatorType" : "client-secret",
     "secret" : "f80d93b4-9039-4012-99c8-a45663485862",
-    "redirectUris" : [ "http://localhost:8090/*", "http://localhost:8082/*"],
+    "redirectUris" : [ "http://localhost:*", "http://activiti-audit:*", "http://activiti-sample:*", "http://activiti-services-gateway:*", "http://activiti-services-registry:*"],
```
Salaboy: @ryandawsonuk this looks sooooo much better!

activiti-services/activiti-services-gateway/src/main/resources/application.properties
ryandawsonuk: I found I had to do this as it runs a bit slower in docker, at least on the first routing action. Otherwise it times out.

activiti-services/activiti-services-registry/pom.xml
ryandawsonuk: Yeah we don't, I just changed it to be consistent but could change it back.

activiti-services/activiti-services-identity-keycloak/src/main/java/org/activiti/services/identity/keycloak/SecurityConfig.java
ryandawsonuk: Could do with looking at this a bit more but it's not actually something changing in this PR. In this file the PR just turns off basic auth.

PR #1378, opened 2017-08-09
BDE-1243 Fix dependency of assertj-core in generator plugin

slanglois:
Specifying the proper version of assertj-core in the Maven generator plugin.
Side effect: fixes the build on Bamboo...


ryandawsonuk:
Indeed - the passing build on bamboo - https://bamboo.alfresco.com/bamboo/browse/ACT-ACTIVITI2 . Thanks @slanglois !

PR #1385, opened 2017-08-10
move starters

ryandawsonuk:
These three PRs all linked:
https://github.com/Activiti/activiti-examples/pull/8
https://github.com/Activiti/activiti-spring-boot-starters/pull/1
https://github.com/Activiti/Activiti/pull/1385


activiti-spring-boot/spring-boot-starters/activiti-spring-boot-audit-starter/src/main/java/org/activiti/configuration/SwaggerConfig.java
```diff
@@ -0,0 +1,39 @@
...
+import springfox.documentation.swagger2.annotations.EnableSwagger2;
+
+@Configuration
+@EnableSwagger2
+public class SwaggerConfig {
+
+    @Bean
+    public Docket api() {
+        return new Docket(DocumentationType.SWAGGER_2).select()
+                                                      .apis(RequestHandlerSelectors.basePackage("org.activiti.services"))
```
Salaboy: @ryandawsonuk check with @erdemedeiros but here we should use org.activiti.services.audit right?

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-audit-starter/.gitignore
```diff
@@ -0,0 +1,13 @@
+/target
```
erdemedeiros: Let's delete this file as it's not adding any new exclusion that's not in the root .gitignore file.

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-audit-starter/src/main/java/org/activiti/configuration/SwaggerConfig.java
erdemedeiros: @Salaboy I think that would work using `org.activiti.services` but you're right maybe it's better to restrict to `org.activiti.services.audit`

PR #1388, opened 2017-08-11
retrofitting to main repo change done under starters

ryandawsonuk:
Change in starter repo -  Activiti/activiti-spring-boot-starters#1 in this commit - Activiti/activiti-spring-boot-starters@bc042f9

Just including for main repo to keep in sync until we remove. Alternatively we could just remove the activiti-spring-boot directory. But note that we should have a bamboo job set up for the new repo first as examples repo depends on starter artefacts getting published.


Salaboy:
@ryandawsonuk I would vote for removing this to avoid wasting time to keep them in sync. Then we will need to set up all the builds in bamboo. Removing this from master will force us to do that.

Salaboy:
@ryandawsonuk can we close this right?

PR #1454, opened 2017-09-04
Gmalanga 1288

gmalanga:
Implemented the ElasticSearch repository for the query service.


activiti-services/activiti-services-query/activiti-services-query-model/src/main/java/org/activiti/services/query/app/model/Task.java
```diff
@@ -0,0 +1,233 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.format.annotation.DateTimeFormat;
+
+
+@Entity
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Task {
```
Salaboy: @gmalanga so you move this to a new place? or why is this changed?

activiti-services/activiti-services-query/activiti-services-query-rest/src/main/java/org/activiti/services/query/events/TaskAssignedEvent.java
```diff
@@ -16,30 +16,22 @@
 
 package org.activiti.services.query.events;
 
-import org.activiti.services.query.model.Task;
+import org.activiti.services.query.es.model.TaskES;
 
 public class TaskAssignedEvent extends AbstractProcessEngineEvent {
 
-    private Task task;
+	private TaskES task;
```
Salaboy: @gmalanga not sure to understand why we need this in the normal query-rest service? Are we mixing stuff here? 
I will be happy with having a shared model or two separate models between ES and JPA, but not a mixed one. How is TaskES different from Task and why we replaced Task for TaskES in the JPA option?

activiti-services/activiti-services-query/activiti-services-query-model/src/main/java/org/activiti/services/query/app/model/ProcessInstance.java
```diff
@@ -0,0 +1,126 @@
+/*
```
Salaboy: @gmalanga Can you check that this class exists in master? because it looks like you are adding this class, but it should exist already.. Maybe it was moved to a different package?

activiti-services/activiti-services-query/activiti-services-query-model/src/main/java/org/activiti/services/query/app/model/Task.java
```diff
@@ -0,0 +1,233 @@
+/*
```
Salaboy: @gmalanga Can you check that this class exists in master? because it looks like you are adding this class, but it should exist already.. Maybe it was moved to a different package?

activiti-services/activiti-services-query/activiti-services-query-model/src/main/java/org/activiti/services/query/app/model/Variable.java
```diff
@@ -0,0 +1,121 @@
+/*
```
Salaboy: @gmalanga Can you check that this class exists in master? because it looks like you are adding this class, but it should exist already.. Maybe it was moved to a different package?

activiti-services/activiti-services-query/activiti-services-query-repo/pom.xml
```diff
@@ -1,57 +0,0 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
```
Salaboy: @gmalanga are you deleting this on purpose? can you check the status of master? because this might be a problem.

activiti-services/activiti-services-query/activiti-services-query-rest/src/main/java/org/activiti/services/query/events/ProcessCompletedEvent.java
```diff
@@ -16,30 +16,22 @@
 
 package org.activiti.services.query.events;
 
-import org.activiti.services.query.model.ProcessInstance;
+import org.activiti.services.query.es.model.ProcessInstanceES;
```
Salaboy: @gmalanga can you explain me why we need to use the *ES classes here? not sure to understand the logic of these changes here in this -rest module which should use the generic model.

activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter/src/test/java/org/activiti/starter/tests/runtime/ProcessInstanceIT.java
```diff
@@ -0,0 +1,262 @@
+/*
```
Salaboy: @gmalanga starters are not longer here in this repository, so you will need to apply your changes in activiti-cloud-starters repository for the relevant starters that you have changed. I imagine, that you are going to provide a new activiti-cloud-query-elastic-starter or something along those lines right?

salaboy (changes requested):
@gmalanga I've added some comments, some of them needs to be fixed and some of them are more clarifications for me to understand what is the scope of the work that you have provided. If you have doubts about my comments please write back. I will also send an email to @erdemedeiros so he check this while he is online.

Salaboy:
@erdemedeiros if you have a couple of days to work remotely you have two options:
1) just keep relaxing on the beach 
2) check this PR and give @gmalanga some assistance so we can move this forward

:)

erdemedeiros (changes requested):
@gmalanga I've added some comments on your changes. My main points are:
- looking at your changes it looks like that you replaced the JPA implementation the Elastic one, but as far as I know we want to keep both
- a lot of noise (merges, commits that are not directly related to your changes) was introduced  to the PR because the branch is not based on the master. Do you think that it would be possible to create a new branch from the master and apply your changes on it?

Salaboy:
Guys I will close this PR. Please provide a new PR. I think that @willkzhou is working on it. But please confirm

PR #1456, opened 2017-09-06
6.0 release zh cn

qijunbo:
I'm  happy to commit a  better  Chinese version  for  activiti-app,   And I'll continuously revise this i18n document.


ryandawsonuk:
Could the target of this PR be swtiched to the 6.x branch and would it be possible to put here a few screenshots to show it working in both Chinese and English?

PR #1476, opened 2017-09-30
Expose Read-only Rest Query API with Spring Data Rest

igdianov:
Here is the solution to expose read-only queries on Query repositories for dynamic run-time Spring Data Rest Query Api that supports schema documentation, sorting, paging queries with links to associated entities with the help of Spring Data Rest with Hateoas. It disables all entity mutation methods with HTTP 405 "Method not allowed" as follows:

1. Created @NoRepositoryBean annotated RestResourceQueryRepository interface that extends PagingAndSortingRepository and overrides mutating methods with @RestResource(exported=true).  

2. Extended existing query repository interface for each entity type instead of CrudRepository. 

3. Customized Spring Data Rest Configuration with base path '/v1' and annotation discovery strategy


igdianov:
Closing to resubmit

PR #1479, opened 2017-10-03
Add a default ServiceTask behavior

### LINKED ISSUE #1474: PoC about default Spring Enabled default Java Delegate for Service Tasks
In Activiti-Examples create a process that defines a Service task without specifying any JavaDelegates and if you are running in a Spring enabled context by default a JavaDelegate will be provided out of the box.

erdemedeiros:
Closes #1474
A new message queue based becomes the default behavior for services tasks:
- services tasks which not specify any class or delegate or type or operation will have this new behavior
- it will send a message in a MessageChannel and wait on trigger
- the stream handling this message should call the RuntimeService.trigger method once it has finished


activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskChannels.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.annotation.Output;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.SubscribableChannel;
+
+public interface ServiceTaskChannels {
+
+    String SERVICE_TASK_RESULT_CONSUMER = "serviceTaskResultConsumer";
+
+    @Input(SERVICE_TASK_RESULT_CONSUMER)
```
Salaboy: @erdemedeiros good stuff.. but let's rename this to integrationEventsProducer and integrationResultsConsumer.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskResultHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private final RuntimeService runtimeService;
+
+    @Autowired
+    public ServiceTaskResultHandler(RuntimeService runtimeService) {
+        this.runtimeService = runtimeService;
+    }
+
+    @StreamListener(ServiceTaskChannels.SERVICE_TASK_RESULT_CONSUMER)
+    public synchronized void receive(ServiceTaskResult serviceTaskResult) {
```
Salaboy: @erdemedeiros same here.. the types should be modelled as events and I think that IntegrationEvent and IntegrationResult makes more sense as names.. Because this should not be related with the BPMN Service Task Concept. We might be able to generate integration events from other places.. not only ServiceTasks

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/model/AsyncContext.java
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.activiti.services.connectors.model;
+
+public class AsyncContext {
+
+    private String processInstanceId;
```
Salaboy: @erdemedeiros as I mention before we might need here a correlationId. This structure is the one that we need to store, are you keeping these instances somewhere? Not about the name yet.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/model/ServiceTaskEvent.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.activiti.services.connectors.model;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ServiceTaskEvent {
```
Salaboy: @erdemedeiros not sure about this structure. I think that we need just one structure. IntegrationEvent which will contain the content of the AsyncContext plus the correlationId plus the variables. is there a reason to have all those types ?

activiti-services/activiti-services-connectors/src/test/java/org/activiti/services/connectors/behavior/MQServiceTaskBehaviorTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Test
+    public void triggerShouldCallLeave() throws Exception {
+        //given
+        DelegateExecution execution = mock(DelegateExecution.class);
+        doNothing().when(behavior).leave(execution);
+
+        //when
+        behavior.trigger(execution, null, null);
```
Salaboy: @erdemedeiros what are those 2 nulls in that trigger method?

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskResultHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+
+        runtimeService.trigger(serviceTaskResult.getExecutionId(),
```
Salaboy: @erdemedeiros before calling runtimeService.trigger() we need to do several checks and that's why I think that we need that persistent correlationId that stores the executionId. So the checks should be something like: 
0) a new IntegrationEvent is created with a correlationId, at this stage we need to store that we will be waiting for that correlationId to be completed when an IntegrationResultEvent arrives.
1) a new IntegrationResultEvent arrives we need to check in the DB that the correlationId that the event contains has a matching in the DB
2) load the entry using the correlationId, load the executionId from the DB and then call the runtimeService.trigger() method
3) delete the correlationId entry from the DB because the Service Task (or integration) was completed. 

Does this make sense? 
Again, sending the execution ID out of the Runtime Bundle scope is not a good idea, process id and task id are fine because they are accesible via our public API anyways.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/behavior/MQServiceTaskBehavior.java
```diff
@@ -0,0 +1,58 @@
...
+    private final ServiceTaskChannels channels;
+
+    @Autowired
+    public MQServiceTaskBehavior(ServiceTaskChannels channels) {
+        this.channels = channels;
+    }
+
+    @Override
+    public void execute(DelegateExecution execution) {
+        AsyncContext context = new AsyncContext(execution.getProcessInstanceId(),
```
Salaboy: @erdemedeiros look at the other comment with the steps to store these contexts in the DB. Here we should create the correlation ID so we don't send the executionId into the IntegrationEvent. 
It sounds like we will need a new engine internal service that will store this info to the DB.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskResultHandler.java
erdemedeiros: ok

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskResultHandler.java
erdemedeiros: It makes sense. I'm going to update that.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/model/AsyncContext.java
erdemedeiros: @Salaboy I'm not sure that we need to store all this information. IMO processInstanceId and TaskId are not necessary.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/model/ServiceTaskEvent.java
erdemedeiros: Well, the SeviceTaskEvent is going to be renamed to IntegrationEvent. I initially grouped executionId, processInstanceId and taskId inside AsyncContext because they are all related to the engine core execution and this avoid having an ugly constructor with too many parameters.
I'm not sure that taskId is actually necessary and we can rename AsyncContext to something else.

activiti-services/activiti-services-connectors/src/test/java/org/activiti/services/connectors/behavior/MQServiceTaskBehaviorTest.java
erdemedeiros: They are `String signalEvent` and `Object signalData`. They are defined in the interface `TriggerableActivityBehavior` but don't seem to be necessary here.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/behavior/MQServiceTaskBehavior.java
erdemedeiros: Ok. I'll work on that.

activiti-services/activiti-services-connectors/src/main/java/org/activiti/services/connectors/behavior/MQServiceTaskBehavior.java
```diff
@@ -0,0 +1,58 @@
...
+                                                     execution.getCurrentActivityId(),
```
balsarori: Note that execution.getCurrentActivityId() is not same as taskId. execution.getCurrentActivityId() is the id element in the BPMN model of the process definition.

erdemedeiros: Thanks @balsarori. Yeah, I realised that after creating the PR. Is there any easy way to get the taskId from the execution? Not sure that this information will be necessary, maybe not.

balsarori: The taskId is useful in case of user tasks only (here you are only working with service tasks), so I don't think its needed here. But I do think the activity is necessary since it can be used to identify the activity that is currently being executed (or was executed).

activiti-engine/src/main/java/org/activiti/engine/impl/cmd/integration/RetrieveIntegrationContextCmd.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.activiti.engine.impl.cmd.integration;
+
+import org.activiti.engine.impl.interceptor.Command;
+import org.activiti.engine.impl.interceptor.CommandContext;
+import org.activiti.engine.impl.persistence.entity.integration.IntegrationContextEntity;
+
+public class RetrieveIntegrationContextCmd implements Command<IntegrationContextEntity> {
+
+    private String correlationId;
```
Salaboy: @erdemedeiros ok let's remove the correlation ID and use the UUID provided by the executionId, that will be secure enough.

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/integration/IntegrationContextEntity.java
```diff
@@ -0,0 +1,45 @@
...
+
+public interface IntegrationContextEntity extends Entity {
+
+    String getExecutionId();
+
+    void setExecutionId(String executionId);
+
+    String getCorrelationId();
+
+    void setCorrelationId(String correlationId);
```
Salaboy: @erdemedeiros ok so let's remove it from here.

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/integration/IntegrationContextEntity.java
```diff
@@ -0,0 +1,45 @@
...
+
+    void setProcessDefinitionId(String processDefinitionId);
+
+    String getProcessInstanceId();
+
+    void setProcessInstanceId(String processInstanceId);
+
+    Date getCreatedDate();
+
+    void setCreatedDate(Date createdDate);
```
Salaboy: @erdemedeiros I don't see the data being sent away in here. We need to make sure that looking at this entity we can retrigger the IntegrationRequestEvent

activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/parser/handler/ServiceTaskParseHandler.java
```diff
@@ -75,7 +75,7 @@ protected void executeParse(BpmnParse bpmnParse, ServiceTask serviceTask) {
       serviceTask.setBehavior(webServiceActivityBehavior);
       
     } else {
-      logger.warn("One of the attributes 'class', 'delegateExpression', 'type', 'operation', or 'expression' is mandatory on serviceTask " + serviceTask.getId());
+      serviceTask.setBehavior(bpmnParse.getActivityBehaviorFactory().createDefaultServiceTaskBehavior(serviceTask));
```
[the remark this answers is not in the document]
balsarori: that might cause a lot of info being logged IMO, maybe debug?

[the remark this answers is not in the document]
erdemedeiros: @Salaboy I agree with @balsarori: this would cause a lot of info being logged. In addition, previously there were a check during the validation phase triggering an error if the information was missing. So I'm not convinced about logging this information here.

activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java
```diff
@@ -49,21 +137,176 @@
...
+import org.activiti.engine.impl.variable.LongType;
+import org.activiti.engine.impl.variable.NullType;
+import org.activiti.engine.impl.variable.SerializableType;
+import org.activiti.engine.impl.variable.ShortType;
+import org.activiti.engine.impl.variable.StringType;
+import org.activiti.engine.impl.variable.UUIDType;
+import org.activiti.engine.impl.variable.VariableType;
+import org.activiti.engine.impl.variable.VariableTypes;
+import org.activiti.engine.integration.IntegrationContextService;
+import org.activiti.engine.integration.IntegrationContextServiceImpl;
```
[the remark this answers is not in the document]
erdemedeiros: Totally agree! I was really upset when I discovered that I'd need to add new lines to these massive class!

activiti-engine/src/main/java/org/activiti/engine/impl/cmd/integration/RetrieveIntegrationContextCmd.java
erdemedeiros: Ok. But should we send the `executionId` as `executionId` or should we hide this behind another name? 
- exposing it as `executionId` would simplify things for people who are already familiar with the concept
- exposing it with a different name gives the flexibility to rely on other information later.
In any case, the `executionId` is already exposed through the `RuntimeService`, so it's not really an internal concept and we can probably expose it in the message.

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/integration/IntegrationContextEntity.java
erdemedeiros: The data is not directly represented in this class, they are hold the the execution. Here we only have the executionId that will allow you two retrieve this data. See the [second pull request for more information](https://github.com/Activiti/activiti-cloud-runtime-bundle-service/pull/1/files#diff-4d3ed153efda1fa500c39d3e8c25a63fR65).

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/integration/IntegrationContextEntityImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    public Date getCreatedDate() {
+        return createdDate;
+    }
+
+    public void setCreatedDate(Date createdDate) {
+        this.createdDate = createdDate;
+    }
+
+    @Override
+    public Object getPersistentState() {
```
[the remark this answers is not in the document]
erdemedeiros: This is used for internal cache purposes. It's here because we are relying on the core engine persistence mechanism that requires this information. This does not refer to the data that is send out. See my previous comment.

erdemedeiros:
@Salaboy the test you mentioned concerning the data transfer can be found in the second part of this pull request: https://github.com/Activiti/activiti-cloud-runtime-bundle-service/pull/1/files#diff-2ad746e9bd827829fb8a3418c06d85f7R48.

activiti-engine/src/main/java/org/activiti/engine/impl/cmd/integration/RetrieveIntegrationContextCmd.java
Salaboy: @erdemedeiros let's use the UUID executionId

activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java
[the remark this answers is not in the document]
Salaboy: 👍

activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/parser/handler/ServiceTaskParseHandler.java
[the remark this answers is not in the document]
Salaboy: @erdemedeiros debug maybe is the best one.. for me it will make some sense to be able to check when the default behavior was selected.

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/integration/IntegrationContextEntityImpl.java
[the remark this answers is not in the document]
Salaboy: @erdemedeiros understood.. so this is a requirement for all entities?

activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/parser/handler/ServiceTaskParseHandler.java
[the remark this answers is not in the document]
erdemedeiros: done.

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/integration/IntegrationContextEntityImpl.java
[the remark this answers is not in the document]
erdemedeiros: Exactly

PR #1483, opened 2017-10-04
Fix Query Model ProcessInstance @Id attribute to use String data type

igdianov:
This PR fixes attribute data type inconsistencies in query entities. For example, @Id field processInstanceId in ProcessInstance is declared as Long type, and  as String in Task and Variable entities. It creates inconsistency in GraphQL and Rest API schemas visible to end-user.  

I have also fixed tests to verify correct data type are used in generated Rest and GraphQL schemas


igdianov:
@erdemedeiros I have not changed the data type semantics of Variable @Id attribute because it is auto-generated internal id and it is not provided by Activiti events. My main concern was about having consistent data type for  processInstanceId used in entity references and query schemas.  I think we don't have to change Variable @Id type right now.

Issue #1493, opened 2017-10-11
security at process definition level


ryandawsonuk:
We want a new set of /admin endpoints for the runtime bundle for read/list tasks which allow an admin user to see everything (which will also mean an update to the keycloak config).

We also want to be able to define some policies such as:

activiti.cloud.group.hr.policy.read=SimpleProcess1,SimpleProcess2
activiti.cloud.user.bob.policy.write=SimpleProcess

We can read this and get it refreshed from the config server by adding the spring-boot-starter-actuator dependency in activiti-services-rest-impl together with:

@ConfigurationProperties("activiti.cloud")
@RefreshScope
public class SecurityProperties  {
    private Map<String, String> group = new HashMap<String, String>();
    private Map<String, String> user = new HashMap<String,String>();
    public Map<String, String> getGroup() {
        return this.group;
    }
    public Map<String, String> getUser() {
        return this.user;
    }
}

ryandawsonuk:
New idea is that we could use a schema like:

activiti.cloud.processDefA.visibleFor = group1,group2,user1
activiti.cloud.processDefA.writeableFor = group1,group2,user1

Or

activiti.cloud.group1.canRead = processDef1,processDef2
activiti.cloud.group2.canWrite = processDef1,processDef2

The reason to prefer this approach is that it is a set of rules that can be overlaid on existing security. Removing all of the rules does not mean that there will be no restrictions or that everything will be restricted - it means that the default engine behaviour is used with whatever identity links are in place in the database.

ryandawsonuk:
In addition to above PRs also an in progress branch at https://github.com/Activiti/activiti-cloud-runtime-bundle-service/tree/feature/ryan-1493-applysecurity

PR #1517, opened 2017-10-24
Feature/ryan 1493 inrestriction

ryandawsonuk:
This doesn't implement #1493 - it is a preliminary step.

### DIFF
--- a/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
+++ b/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
@@ -34,35 +34,35 @@ public class ProcessDefinitionQueryImpl extends AbstractQuery<ProcessDefinitionQ
 
     private static final Logger log = LoggerFactory.getLogger(ProcessDefinitionQueryImpl.class);
 
-    protected String id;
-    protected Set<String> ids;
-    protected String category;
-    protected String categoryLike;
-    protected String categoryNotEquals;
-    protected String name;
-    protected String nameLike;
-    protected String deploymentId;
-    protected Set<String> deploymentIds;
-    protected String key;
-    protected String keyLike;
-    protected Set<String> keys;
-    protected String resourceName;
-    protected String resourceNameLike;
-    protected Integer version;
-    protected Integer versionGt;
-    protected Integer versionGte;
-    protected Integer versionLt;
-    protected Integer versionLte;
-    protected boolean latest;
-    protected SuspensionState suspensionState;
-    protected String authorizationUserId;
-    protected String procDefId;
-    protected String tenantId;
-    protected String tenantIdLike;
-    protected boolean withoutTenantId;
+    private String id;
+    private Set<String> ids;
+    private String category;
+    private String categoryLike;
+    private String categoryNotEquals;
+    private String name;
+    private String nameLike;
+    private String deploymentId;
+    private Set<String> deploymentIds;
+    private String key;
+    private String keyLike;
+    private Set<String> keys;
+    private String resourceName;
+    private String resourceNameLike;
+    private Integer version;
+    private Integer versionGt;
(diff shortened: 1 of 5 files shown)


activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
```diff
@@ -149,6 +150,14 @@ public ProcessDefinitionQueryImpl processDefinitionKey(String key) {
         return this;
     }
 
+    public ProcessDefinitionQueryImpl processDefinitionKeys(Set<String> keys) {
+        if (keys == null) {
```
erdemedeiros: I'm wondering if this `if` is necessary. When you create a new instance of ProcessDefinitionQueryImpl the default value is `null` and that's a valid state. In addition, you have a check in the query itself. What do you think?

activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
```diff
@@ -149,6 +150,14 @@ public ProcessDefinitionQueryImpl processDefinitionKey(String key) {
         return this;
     }
 
+    public ProcessDefinitionQueryImpl processDefinitionKeys(Set<String> keys) {
```
erdemedeiros: IMO the return type should be the interface instead of the implementation.

activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
ryandawsonuk: Yeah it's not necessary, I think the idea is to reveal to the consumer code using the API that there is a mistake in the consumer code. If you've called this method and passed in null then presumably you've made a slip and intended to pass in a value as passing null would be pointless.

activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
```diff
@@ -45,6 +45,7 @@
     protected Set<String> deploymentIds;
     protected String key;
     protected String keyLike;
+    protected Set<String> keys;
```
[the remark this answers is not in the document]
ryandawsonuk: I would say we could make them all private but am not sure if people out there might be extending this class in their own implementations and if so what our view on that should be. Will speak to you further about this and next point.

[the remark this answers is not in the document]
erdemedeiros: I think that make them all private it's fine. Accessing directly the field is usually perceived as a bad practice end getters and setters will be available anyway.

activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
erdemedeiros: Ignore my previous comment on this. It's ok to have the implementation here (`ProcessDefinitionQueryImpl`), the important is to have the interface in `ProcessDefinitionQuery`. For testing purposes it might be interesting to have the implementation type here.

activiti-engine/src/main/java/org/activiti/engine/impl/ProcessDefinitionQueryImpl.java
erdemedeiros: Yeah, it makes sense. But in this case it would be nice to have unit tests covering that.

ryandawsonuk: Have added a test

Salaboy:
@ryandawsonuk what is this exactly doing? not sure to follow what is this doing. Let's catch up tomorrow.

Issue #1553, opened 2017-11-09
Deploy via Rest-API always fails with "Multipart request is required"


hoax:
I have trouble to deploy a process definition using the REST API. First I thought it is because of my Javascript code but even with curl it won't work:

```
$ curl -v --basic --user admin:admin -F file=@assignTaskToUser.bpmn20.xml http://localhost:8080/activiti-app/api/repository/deployments`
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
* Server auth using Basic with user 'admin'
> POST /activiti-app/api/repository/deployments HTTP/1.1
> Host: localhost:8080
> Authorization: Basic YWRtaW46YWRtaW4=
> User-Agent: curl/7.52.1
> Accept: */*
> Content-Length: 3357
> Expect: 100-continue
> Content-Type: multipart/form-data; boundary=------------------------3a84c849f7353c4d
> 
< HTTP/1.1 100 Continue
< HTTP/1.1 400 Bad Request
< Server: Apache-Coyote/1.1
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< Cache-Control: no-cache, no-store, max-age=0, must-revalidate
< Pragma: no-cache
< Expires: 0
< X-Frame-Options: DENY
< Content-Type: application/json
< Transfer-Encoding: chunked
< Date: Thu, 09 Nov 2017 10:23:35 GMT
< Connection: close
< 
* Curl_http_done: called premature == 0
* Closing connection 0
{"message":"Bad request","exception":"Multipart request is required"}
```

What I found out is, that the exception gets thrown in _activiti-rest/src/main/java/org/activiti/rest/service/api/repository/DeploymentCollectionResource.java#uploadDeployment_. Here the request is expected to be of MultipartHttpServletRequest but actually is of type org.springframework.security.web.servletapi.HttpServlet3RequestFactory$Servlet3SecurityContextHolderAwareRequestWrapper.

I'm using Activiti 6.0.0 with tomcat 7.075.

Does anyone know this problem and/or how to resolve it?

hoax:
Found the bug. Fix: copy method _multipartResolver()_ from _org.activiti.app.servlet.AppDispatcherServletConfiguration_ to _org.activiti.app.servlet.ApiDispatcherServletConfiguration_.

hoax:
[...]

PR #1562, opened 2017-11-15
provide MulipartResolver bean for REST API

hoax:
fixes broken multipart file uploads, see issue
https://github.com/Activiti/Activiti/issues/1553


Salaboy:
@balsarori this looks like a good contribution can we decide if we will merge it or ask for tests agains the changes? This is pointed to the correct branch and besides not having tests the change seems to solve the problem mentioned in the issue linked here.

hoax:
Unfortunately I don't know Spring good enough to be able to write a related test.

balsarori:
I agree. It's a good one indeed, thanks @hoax. Can you set a default max size or maybe just set the max size if the property exists, so that it doesn't  break if property isn't set?

hoax:
set to 1mb as this is the current default in activiti-app.properties

Salaboy:
@hoax thanks a lot for the contribution!
@balsarori please merge this one in 6.x branch.

balsarori:
thanks @hoax

PR #1131, opened 2017-03-09
fix Chinese Disorderly Code when import/edit/deploy model .#818

MicyToy:
fix Chinese Disorderly Code when import/edit/deploy model #818  .Based on **5.21.0-release**

reslove import/edit messy-code
add configuration in `activiti-custom-context.xml` under `processEngineConfiguration`

`<property name="activityFontName" value="\u5fae\u8f6f\u96c5\u9ed1" />
<property name="labelFontName" value="\u5fae\u8f6f\u96c5\u9ed1"/>`

PR #1568, opened 2017-11-23
fix Chinese Disorderly Code when import/edit/deploy model on branch 5.x

### LINKED ISSUE #818: 5.19.0 Deployment causes Chinese messy code
If there are Chinese characters in deployed process, it will cause Chinese messy code in the database and acitiviti explorer deployed process definition page:
I found the reason is in the class `org.activiti.editor.ui.EditorProcessDefinitionDetailPanel` and in the function `deployModelerModel`

```
deployment = this.repositoryService.createDeployment().name(this.modelData.getName()).addString(processName, new String(bpmnBytes)).deploy();
```

`bpmnBytes` should encode with "uft-8" and this line shoud change to:

```
deployment = this.repositoryService.createDeployment().name(this.modelData.getName()).addString(processName, new String(bpmnBytes, "utf-8")).deploy();
```

MicyToy:
fixed #818 
see also PR #1131

### DIFF
--- a/modules/activiti-explorer/src/main/java/org/activiti/editor/ui/EditorProcessDefinitionDetailPanel.java
+++ b/modules/activiti-explorer/src/main/java/org/activiti/editor/ui/EditorProcessDefinitionDetailPanel.java
@@ -12,10 +12,18 @@
  */
 package org.activiti.editor.ui;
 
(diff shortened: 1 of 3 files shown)


balsarori:
Thanks @MicyToy

Issue #1610, opened 2017-12-11
Possible bug in editor-controller.js - Canvas resizers getting positioned incorrectly


piyushshri:
# HOW TO REPRODUCE
1. Open a new model
2. Click on the default Start Event and end an End Event by clicking in the quick menu icon (Do not click anywhere else, keep the new quick menu of the End Event visible)
3. Click on the left or right canvas resizers (canvas_resize_indicator)

The canvas resize indicators get repositioned to the center of the canvas.

# THE CAUSE
This is happening because in editor-controller.js at line [225](https://github.com/piyushshri/Activiti/blob/de8e347b012829bfc811cfbee47b5b700799797c/modules/activiti-ui/activiti-app/src/main/webapp/editor/editor-app/editor-controller.js#L225) , `canvas[0].children[1]` is not giving the full canvas, instead it is giving the delete button from the quick menu of the End Event. Therefore, `actualWidth` is getting incorrectly calculated which leads to incorrect value of `canvasLeft`. This pushes the resize indicators to the center.

The actual canvas is at `canvas[0].children[11]` . 

Please confirm if this is a bug.

Salaboy:
It seems that this was merged and the PR closed as well. So I will close the issue, please write a comment if this is not correct.

piyushshri:
@Salaboy No, I closed the pull request as my solution didn't completely solve the problem. I can work on the issue once a master/owner confirms that this is an actual bug.

Salaboy:
@balsarori can you confirm that? if this is reproducible we will be happy to receive a fix from @piyushshri

PR #1611, opened 2017-12-11
Resolving resize indicator position issue

piyushshri:
This pull request resolved a possible bug as listed here https://github.com/Activiti/Activiti/issues/1610 .


piyushshri:
Closing this pull request as this solution does not work for slow networks. The window resize methods get called before the canvas is loaded and result in the same problem again. Will propose a better solution once the issue is addressed by maintainers.

PR #1618, opened 2017-12-14
Fix signal events throw/catch race conditions in transaction scope

igdianov:
Currently, throw/catch signal event handling in Activiti engine causes race conditions behavior to occur when throw/catch signals are used to synchronize parallel activities in the current transaction scope.

For example, after a parallel gateway, due to various task execution timings or non-deterministic execution path by Activiti engine, a signal on one path is emitted before subscribers on parallel paths can subscribe to it, causing subscribers to block and wait indefinitely:

image

To fix this problem, the PR introduces throw signal event registration in the transaction execution context. This will allow subscribers to receive events after the signal has been published, so the execution can proceed if the signal has been registered  for any global or process scoped events in the current  transaction scope.

I have investigated a fix which deferred signal throw until next transaction. Unfortunately, it breaks a lot existing tests which expect Activiti signal catch behavior to receive signals in the same transaction. 

The only viable approach is to fix race conditions with ThrowSignalEventSessionManager in transaction scope in order to maintain consistent signal catch behavior between process and global scoped events.

The following tests are covered

1. Throw Signal in Process Instance and Global scopes with Sync and Async Catch behaviors:
image

2. Throw signal in the same transaction with Event Based Gateway:
image


.gitignore
```diff
@@ -9,3 +9,9 @@ target
 .DS_Store
 *.bpmn/
 /userguide/src/en/index.html
+
+\.metadata/
+
+\.recommenders/
+
+modules/activiti-spring-boot/spring-boot-starters/activiti-spring-boot-starter-basic/\.factorypath
```
daisuke-yoshimoto: It's a very trivial problem, but is it correct to specify only the files of this project?

igdianov: I am using GitKraken to deal with Git. By default it puts directives to ignore specific files. I fixed it.

Salaboy:
@ryandawsonuk @erdemedeiros are we ok with this one? So @balsarori can merge?

daisuke-yoshimoto:
@Salaboy 
I'm sorry  for the delay of the review.
However, because the modification scale is large, it can not be easily reviewed.

Please wait this whole week.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ThrowSignalEventSessionManager.java
```diff
@@ -0,0 +1,54 @@
+package org.activiti.engine.impl.persistence.entity;
+
+
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.function.Function;
```
balsarori: This is a java 8 feature. 5.x supports java 1.6.

erdemedeiros:
@igdianov sorry, I'm late on my reviews. I try to take look on that today or tomorrow.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/IntermediateThrowSignalEventActivityBehavior.java
```diff
@@ -62,5 +68,10 @@ public void execute(ActivityExecution execution) throws Exception {
       leave(execution);
     }
   }
- 
+
+  @Override
```
daisuke-yoshimoto: @igdianov 
I'm sorry for pointing at a minor error, but what is the purpose of this override?

modules/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class ThrowSignalEventSessionManagerTest extends PluggableActivitiTestCase {
+
+  @Deployment
+  public void testSignalRaceConditionsAfterParallelGateway() {
+
+    runtimeService.startProcessInstanceByKey("testSignalRaceConditionsAfterParallelGateway");
+    
+    // No tasks should be open then and process should have ended
+    assertEquals(0, taskService.createTaskQuery().count());
```
daisuke-yoshimoto: @igdianov I'm sorry that I cannot understand the intention of checking task count.
Because there is no user task setting in the process definition of the test material.

modules/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.java
```diff
@@ -0,0 +1,58 @@
+package org.activiti.engine.test.bpmn.event.signal;
+
+import org.activiti.engine.impl.test.PluggableActivitiTestCase;
+import org.activiti.engine.test.Deployment;
+
+public class ThrowSignalEventSessionManagerTest extends PluggableActivitiTestCase {
```
daisuke-yoshimoto: I would like to add the test case that check the case behaviors not supported by this fix.
Specifically, it is a test case for patterns that set the following for each signal event.
1. activiti:async="true"
2. activiti:async="true" AND activiti:exclusive="false"

modules/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ThrowSignalEventSessionManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.activiti.engine.impl.persistence.AbstractManager;
+import org.activiti.engine.impl.pvm.delegate.ActivityExecution;
+
+/**
+ * Keep track of fired events for process instance execution 
+ * in the current transaction context 
+ *
+ */
+public class ThrowSignalEventSessionManager extends AbstractManager {
```
daisuke-yoshimoto: Is it necessary to implement AbstractManager.class?
It does not seem to be persisting the state of the signal in the database.
AbstractManager.class is the abstract class for managing the entity (although there are some exceptions such as LDAPGroupManager.class and LDAPUserManager.class).

If you just want to keep the state of the signal in CommandContext, will you be able to use the attributes of CommandContext?
https://github.com/Activiti/Activiti/blob/5.x/modules/activiti-engine/src/main/java/org/activiti/engine/impl/interceptor/CommandContext.java#L83
https://github.com/Activiti/Activiti/blob/5.x/modules/activiti-engine/src/main/java/org/activiti/engine/impl/interceptor/CommandContext.java#L238
https://github.com/Activiti/Activiti/blob/5.x/modules/activiti-engine/src/main/java/org/activiti/engine/impl/interceptor/CommandContext.java#L245

modules/activiti-engine/src/test/resources/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.testSignalRaceConditionsAfterParallelGateway.bpmn20.xml
```diff
@@ -0,0 +1,276 @@
...
+  <signal id="signal1" activiti:scope="processInstance" name="Signal1"/>
+  <process id="testSignalRaceConditionsAfterParallelGateway" name="Racing Event Signals After Parallel Gateway" isExecutable="true">
+    <startEvent id="StartEvent_1">
+      <extensionElements>
+        <modeler:editor-resource-id>StartEvent_1</modeler:editor-resource-id>
+      </extensionElements>
+      <outgoing>sid-0A4A04CE-9653-4B46-B81C-FE8BC3B2E2AA</outgoing>
+    </startEvent>
+    <parallelGateway id="ExclusiveGateway_1">
```
daisuke-yoshimoto: I'm sorry for pointing at a minor error, but why is it giving the label 'ExclusiveGateway' even though it is a parallelGateway?

daisuke-yoshimoto:
@Salaboy @igdianov 
I'm sorry for keeping you waiting for such a long time.
My review is complete.

There is no objection to the rough modification policy, but I commented on some implementation methods.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/IntermediateThrowSignalEventActivityBehavior.java
igdianov: Removed

modules/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.java
igdianov: Corrected

modules/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.java
igdianov: @daisuke-yoshimoto I am not sure this will have any effect on throw, catch and event gateway behaviors, as I am already testing for all async throw events. Any async catch behavior should work out of the box in a separate transaction. I will be happy to execute your test case...

modules/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ThrowSignalEventSessionManager.java
igdianov: It is actually a good suggestion. I have removed ThrowSignalEventSessionManager in favor of CommandContext attributes. It makes the fix less invasive. :)

modules/activiti-engine/src/test/resources/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.testSignalRaceConditionsAfterParallelGateway.bpmn20.xml
igdianov: Fixed

igdianov:
@Salaboy

I have implemented @daisuke-yoshimoto suggestions. Also, I have optimized catch behaviors to reuse existing event subscriptions in the execution scope to mimic existing event received Activity handling.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ThrowSignalEventSessionManager.java
igdianov: Fixed. ThrowSignalEventSessionManager is removed in favor of CommandContext session attributed

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/IntermediateCatchEventActivityBehavior.java
```diff
@@ -12,13 +12,31 @@
...
 
 
 public class IntermediateCatchEventActivityBehavior extends AbstractBpmnActivityBehavior {
 
-
+  @Override
   public void execute(ActivityExecution execution) throws Exception {
+    
+    // Find active subscriptions
+    List<EventSubscriptionEntity> subscriptions = ((ExecutionEntity)execution).getEventSubscriptions();
```
erdemedeiros: The code block added here (lines 29 to 38) is identical to the code added for `EventBasedGatewayActivityBehavior`, excepted by the fact that the first one has a `break` clause. Is this difference intentional? If not, this code could be extracted into a separated class which is injected where necessary to avoid logic duplication. 
In the case where injecting a class is too intrusive in the current code structure you can imagine to have a method in a common super class as well.

What do you think?

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/EventBasedGatewayActivityBehavior.java
```diff
@@ -13,18 +13,42 @@
...
   public void execute(ActivityExecution execution) throws Exception {
-    // the event based gateway doesn't really do anything
-    // ignoring outgoing sequence flows (they're only parsed for the diagram)
+      
+      // Continue with signal catch event activities for signal events
+      // already fired and registered in process instance execution scope
+      List<EventSubscriptionEntity> subscriptions = ((ExecutionEntity)execution).getEventSubscriptions();
+
+      for(EventSubscriptionEntity subscription: subscriptions) {
+          // If event already fired matches existing subscription
```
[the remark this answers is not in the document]
igdianov: @erdemedeiros Sorry, for slow reply - I am away on vacation right now. I have extracted the method and moved it to super class.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/IntermediateCatchEventActivityBehavior.java
igdianov: @erdemedeiros The break statement in `EventBasedGatewayActivityBehavior` is intentional to handle only one out of many signals after the gateway. I have changed the corresponding code in `IntermediateCatchEventActivityBehavior` to distinguish for the difference in handling. I also moved common signal event registration and checking logic to FlowNodeActivityBehavior super class as well.

ryandawsonuk:
@igdianov I see you've made some more changes since @balsarori 's review. Is this ready to merge from your perspective?

igdianov:
@ryandawsonuk I have made changes as suggested by @erdemedeiros. It is ready to merge from my side if there are no other suggestions for improvements...

modules/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/ThrowSignalEventSessionManagerTest.java
daisuke-yoshimoto: @igdianov I'm sorry for the delay.

> as I am already testing for all async throw events. Any async catch behavior should work out of the box in a separate transaction. I will be happy to execute your test case...

I made test cases  not supported by this fix.
Please check them.
https://drive.google.com/file/d/1rKMNohUpNIAqgCCHql2t6P7M_fVCryla/view?usp=sharing
https://drive.google.com/open?id=1wsK60AXiUmhw0rGN-3jAfrnxAscTnjuT
https://drive.google.com/open?id=17KnP88mlbceHceZxE8RgpTzyPSEzwAAi

daisuke-yoshimoto:
@igdianov I'm sorry for the delay.

> as I am already testing for all async throw events. Any async catch behavior should work out of the box in a separate transaction. I will be happy to execute your test case...

I made test cases  not supported by this fix.
Please check them.
https://drive.google.com/file/d/1rKMNohUpNIAqgCCHql2t6P7M_fVCryla/view?usp=sharing
https://drive.google.com/open?id=1wsK60AXiUmhw0rGN-3jAfrnxAscTnjuT
https://drive.google.com/open?id=17KnP88mlbceHceZxE8RgpTzyPSEzwAAi

igdianov:
@daisuke-yoshimoto I have reviewed your test cases. The engine executes each signal behavior in a separate isolated Command Context. Therefore, it is not possible to resolve thrown signal events in the same transaction with async:true behavior, because it requires common shared memory signal registry with persistence.

daisuke-yoshimoto:
@igdianov 

Thank you for reply.

Yes, this case is not supported.

So, my intention is to leave the unsupported case as a test case so that the specification of this modification will become clearer later.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/FlowNodeActivityBehavior.java
```diff
@@ -52,5 +53,20 @@ public void signal(ActivityExecution execution, String signalName, Object signal
     // concrete activity behaviours that do accept signals should override this method;
     throw new ActivitiException("this activity doesn't accept signals");
   }
+
+  /**
+   * Register fired signals to handle race conditions within current transaction scope
+   */
+  protected void registerFiredSignalEvent(ActivityExecution execution, String eventName) {
```
daisuke-yoshimoto: If the command is executed  in a nested state, it may result in unintended movement. The signal of the process instance processing within the parent scope command may also execute other process instance's signal subscriptions within the child scope command.

daisuke-yoshimoto: ## Test case

https://files.gitter.im/Activiti/Activiti7/s2fL/SignalEventTest.java
https://files.gitter.im/Activiti/Activiti7/b8EB/SignalEventTests.nestCommandParent.bpmn20.xml
https://files.gitter.im/Activiti/Activiti7/4Ujj/SignalEventTests.nestCommandSub.bpmn20.xml

balsarori: Signal names should be unique

daisuke-yoshimoto: No, if the scope of the signal is a process instance, signal catch events should not be handled cross different process instances, even if the signal names are the same.

balsarori: @igdianov what are your thoughts?

daisuke-yoshimoto: @igdianov @balsarori 
I made the pull request https://github.com/Activiti/Activiti/pull/1699 for this problem.

balsarori: @daisuke-yoshimoto
how about just adding extra info to the attribute to identify it from any other nested commands? for example adding the process instance id
```Context.getCommandContext().addAttribute(eventName + ":" + execution.getProcessInstanceId())```

igdianov: @balsarori @daisuke-yoshimoto I had reviewed #1699 by @daisuke-yoshimoto. I think it is a generally good idea to register signals and then match event subscribers by signal scope definition in the command context to avoid corner cases such as #1699.

balsarori:
thanks @igdianov & @daisuke-yoshimoto

PR #1659, opened 2018-01-05
add .editorconfig matching existing IntelliJ/Eclipse settings

mteodori:
should this PR also fix the existing formatting mismatch issues?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


ryandawsonuk:
After discussion with @Salaboy seems the best way forward on this is to put it in https://github.com/Activiti/activiti-build/tree/develop/ide-configuration and maybe also update the README here to say that exists

PR #1684, opened 2018-01-18
Task and SequenceFlow names do not display

wangjunyuan89:
Resolve the problem that Activiti workflow images do not display Task names and do not display SequenceFlow names.


daisuke-yoshimoto:
@wangjunyuan89 
Is the target to be merged for this pull request develop branch?
In 6.x or 5.x?

Salaboy:
@wangjunyuan89 adding on top @daisuke-yoshimoto comment. This looks like a good contribution, but there are several things that needs to happen for us in order to accept this. First of all we need to understand the intention and where are those two classes going to be used. Second, we need to have test coverage for the functionality. Third Codacy and CodeCov needs to be happy about the changes and now they are not. Unfortunately we are not getting any information back from you, which makes me think that this PR is abandoned.

PR #1699, opened 2018-01-25
Fix for nested command case

daisuke-yoshimoto:


igdianov:
@daisuke-yoshimoto It is a good use case to match signals using their scope definition. I had made some minor changes to embed all signal registration and event subscription resolution in one place to make throw and catch signal behaviors clean.

modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/FlowNodeActivityBehavior.java
```diff
@@ -57,16 +64,57 @@ public void signal(ActivityExecution execution, String signalName, Object signal
...
+
+  protected boolean isSignalEventType(EventSubscriptionEntity subscription) {
+    return "signal".equals(subscription.getEventType());
+  }
+
+  protected String getEventSubscriptionScope(ActivityExecution execution, EventSubscriptionEntity subscription) {
+    String subscriptionScope = null;
+    
+    if (subscription.getConfiguration() != null) {
+      Pattern pattern = Pattern.compile("\"scope\"\\s*:\\s*\"([^\"]+)\",?");
```
balsarori: Can this be a static field?

daisuke-yoshimoto:
@igdianov 
Thanks a lot for your improvement.
But, the balsarori's idea that your improvement depends on has two concerns.

## 1. The letter ':' is inappropriate as a delimiter.

We can use the letter ':' as a signal name.
In addition, arbitrary characters can be set as processInstanceId , so any character may not be used as delimiter. processInstanceId is numbered by IdGenerator implementation. Developer can implement IdGenerator interface and can plug own numbering logic for processInstanceId.

## 2. Existing coding habit

Any developer can use CommandContext's attribute. So, we need to be aware of attribute key conflicts.

In the existing implementation, instead of directly attaching data to the key, the fixed type is stored in the key, and the collection that manage data is stored in the value.

https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/event/logger/EventLogger.java#L101

https://github.com/Activiti/activiti-cloud-runtime-bundle-service/blob/develop/activiti-cloud-services-runtime-bundle/activiti-cloud-services-events/src/main/java/org/activiti/cloud/services/events/listeners/BaseCommandContextEventsAggregator.java#L33
https://github.com/Activiti/activiti-cloud-runtime-bundle-service/blob/develop/activiti-cloud-services-runtime-bundle/activiti-cloud-services-events/src/main/java/org/activiti/cloud/services/events/listeners/ProcessEngineEventsAggregator.java#L45

igdianov:
@daisuke-yoshimoto I agree it is better to use better naming to avoid collisions. I think managing fired signal attributes as a collection is too verbose - better keep it simple and explicit. I made improvements to signal event attribute name execution scope generation to address that.

@balsarori I moved regex pattern to static field

daisuke-yoshimoto:
@igdianov Thanks.

balsarori:
thanks @daisuke-yoshimoto & @igdianov 
I'm merging this and will do a final review on the main PR

PR #1612, opened 2017-12-12
Update the persisted historic task instances that are open

lourdupranay:
Update the persisted historic task instances that are open

PR #1732, opened 2018-02-05
fix task local variables execution id replacement

balsarori:
Following discussion in PR #1612

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
@@ -394,6 +394,9 @@ public void recordActivityStart(ExecutionEntity executionEntity) {
         historicActivityInstance.setExecutionId(replacedBy.getId());
       }
 
+      // keep record of unfinished tasks to only update related variables later
+      Set<String> unfinishedTasks = new HashSet<String>();
+      
       // Update the cached historic task instances that are open
       List<HistoricTaskInstanceEntity> cachedHistoricTaskInstances = getDbSqlSession().findInCache(HistoricTaskInstanceEntity.class);
       for (HistoricTaskInstanceEntity cachedHistoricTaskInstance: cachedHistoricTaskInstances) {
@@ -401,6 +404,7 @@ public void recordActivityStart(ExecutionEntity executionEntity) {
                && (execution.getId().equals(cachedHistoricTaskInstance.getExecutionId()))
              ) {
               cachedHistoricTaskInstance.setExecutionId(replacedBy.getId());
+              unfinishedTasks.add(cachedHistoricTaskInstance.getId());
           }
         }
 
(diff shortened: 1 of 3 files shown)


modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
```diff
@@ -407,8 +409,12 @@ public void recordExecutionReplacedBy(ExecutionEntity execution, InterpretableEx
         .executionId(execution.getId())
         .unfinished()
         .list();
+      
+      // keep record of unfinished tasks to only update related variables later
+      Set<String> unfinishedTasks = new HashSet<String>();
       for (HistoricTaskInstanceEntity historicTaskInstance: historicTaskInstances) {
         historicTaskInstance.setExecutionId(replacedBy.getId());
+        unfinishedTasks.add(historicTaskInstance.getId());
```
daisuke-yoshimoto: We need to consider unfinished entities as well, so we need to add the cached entities to unfinishedTasks. Specifically, it is the following part.

https://github.com/Activiti/Activiti/blob/a43ff77ed55f881abeb7da87125b29f0bbd30df6/modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java#L403

modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
```diff
@@ -423,7 +429,11 @@ public void recordExecutionReplacedBy(ExecutionEntity execution, InterpretableEx
         .executionId(execution.getId())
         .list();
       for (HistoricVariableInstanceEntity historicVariableInstance: historicVariableInstances) {
-          historicVariableInstance.setExecutionId(replacedBy.getId());
+    	  // only variables of unfinished tasks
+    	  if (unfinishedTasks.contains(historicVariableInstance.getTaskId())) {
```
daisuke-yoshimoto: We need to consider the cached entities that is not flushed as well, so we need to add this filter to the cached entities as well. Specifically, it is the following part.
https://github.com/Activiti/Activiti/blob/a43ff77ed55f881abeb7da87125b29f0bbd30df6/modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java#L421

modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
balsarori: do you have a case when this is needed?

modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
balsarori: I've also included cached tasks to the check

modules/activiti-engine/src/main/java/org/activiti/engine/impl/history/DefaultHistoryManager.java
daisuke-yoshimoto: @balsarori 
I am sorry to have kept you waiting.

I added the case when this is need in the following pull request.
https://github.com/Activiti/Activiti/pull/1739

PR #1736, opened 2018-02-06
feat: add support for localization information conversion

alirezamirian:
Activiti currently supports localized **name** and **documentation** for a couple of things such as `UserTask` and `SubProcess`. This support for specifying name and documentation in different locales is done via a special **localization** extension element, and is handled inside [`BpmnDeployer#createLocalizationValues()`][1]. 
However `ActivitiJsonConverter` doesn't care about this localization information and it's not retained when converting to/from json.

This pull request adds support for converting localization information when converting `BpmnModel` to/from json via `ActivitiJsonConverter`.

[1]: https://github.com/Activiti/Activiti/blob/f1526dba92702aa07a0dcffb0144436d45074c7a/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/deployer/BpmnDeployer.java#L261

### DIFF
--- a/activiti-json-converter/src/main/java/org/activiti/editor/constants/StencilConstants.java
+++ b/activiti-json-converter/src/main/java/org/activiti/editor/constants/StencilConstants.java
@@ -241,4 +241,6 @@ public interface StencilConstants {
   final String PROPERTY_DECISIONTABLE_REFERENCE_ID = "decisiontablereferenceid";
   final String PROPERTY_DECISIONTABLE_REFERENCE_NAME = "decisiontablereferencename";
   final String PROPERTY_DECISIONTABLE_REFERENCE_KEY = "decisionTableReferenceKey";
+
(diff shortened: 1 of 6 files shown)


Salaboy:
@alirezamirian this is a very good contribution, and sorry for keep it open for so long, but I guess that what Codacy is reporting is quite important. This is a very good opportunity to make those methods less complex and also to fix the other issues reported in the Codacy report. Let us know if you can tackle those and we will merge this PR immediately after those are fixed.

alirezamirian:
Thank you, and yes, Codacy is doing a nice job. I should have fix them
earlier. Anyways, I think I can fix the reported issues of this and the
other PR in next couple of days.
Thank for following up.

Salaboy:
@alirezamirian awesome! let's get this done for Beta1!

alirezamirian:
Issues are fixed and it's also rebased onto develop branch.
@Salaboy

PR #1739, opened 2018-02-07
Fix for cached entity case

daisuke-yoshimoto:


balsarori:
thanks @daisuke-yoshimoto

PR #1740, opened 2018-02-08
bug fix: Could not create tables automatically

shenyefeng:
Could not create tables automatically when change to a new database in MySQL.


shenyefeng:
Could not create tables automatically when change to a new database in MySQL.

Salaboy:
@balsarori can you please check this. It is pointing to 6.x

balsarori:
@shenyefeng you should set the catalog using ```ProcessEngineConfiguration.setDatabaseCatalog```

PR #1783, opened 2018-02-19
Signals between process definitions in cloud

daisuke-yoshimoto:


daisuke-yoshimoto:
With @Salaboy's feedback, I found that extending process engine itself is unnecessary.
So, I close this pull request.

PR #1798, opened 2018-02-28
Rename MQServiceTaskBehavior bean

balsarori:
Need to create a PR for RB that set MQServiceTaskBehavior as the default behavior service tasks


erdemedeiros:
Refers to https://github.com/Activiti/Activiti/issues/1886

PR #1818, opened 2018-03-13
ACTIVITI-937 set user task end time when it is completed

cristianflorescu:


cristianflorescu:
As discussed with Bassam pr should be made against 5.x

PR #1874, opened 2018-04-13
#1854 correct task event order

### LINKED ISSUE #1854: Engine is returning reversed order of events for creating a new standalone task
If a new standalone task is created with an assignee using the `TaskService` the engine triggers the **TASK_ASSIGNED** event before **TASK_CREATED**. 
Investigate the impact of changing this inside the engine. 
@balsarori provided some context:

> - new task is created in UserTaskActivityBehavior.java, insert method is called here
> https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java#L196
> 
> - which fires TASK_ASSIGNED is here
> https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManagerImpl.java#L99
> - later in UserTaskActivityBehavior.java the TASK_CREATED event is triggered
> https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java#L216
> - should be possible to do an insert without triggering TASK_ASSIGNED and later fire TASK_CREATED & TASK_ASSIGNED events

balsarori:
fixes #1854

### DIFF
--- a/activiti-engine/src/test/java/org/activiti/engine/test/api/event/DatabaseEventLoggerTest.java
+++ b/activiti-engine/src/test/java/org/activiti/engine/test/api/event/DatabaseEventLoggerTest.java
@@ -208,7 +208,7 @@ public class DatabaseEventLoggerTest extends PluggableActivitiTestCase {
       if (i == 11 || i == 14) {
 
         assertNotNull(entry.getType());
-        assertEquals(ActivitiEventType.TASK_CREATED.name(), entry.getType());
+        assertEquals(ActivitiEventType.TASK_ASSIGNED.name(), entry.getType());
         assertNotNull(entry.getTimeStamp());
         assertNotNull(entry.getProcessDefinitionId());
         assertNotNull(entry.getProcessInstanceId());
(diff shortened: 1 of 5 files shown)


activiti-engine/src/test/java/org/activiti/engine/test/api/event/DatabaseEventLoggerTest.java
```diff
@@ -521,11 +521,19 @@ public void testStandaloneTaskEvents() throws JsonParseException, JsonMappingExc
...
+    task.setAssignee("kermit");
+    taskService.saveTask(task);
 
     List<EventLogEntry> events = managementService.getEventLogEntries(null, null);
-    assertEquals(2, events.size());
-    assertEquals("TASK_ASSIGNED", events.get(0).getType());
-    assertEquals("TASK_CREATED", events.get(1).getType());
+    assertEquals(4, events.size());
+    assertEquals("TASK_CREATED", events.get(0).getType());
+    assertEquals("TASK_ASSIGNED", events.get(1).getType());
```
Salaboy: @balsarori why so many??

balsarori: Each time assignee changes an event should be triggered. Here I want to make sure that all cases are covered and that my changes didn't miss a case :).

Issue #1905, opened 2018-06-05
ReflectUtil.loadClass(String) throws `java.lang.ClassNotFoundException: int` on primitive types


raja-anbazhagan:
We are using activiti version 5.18.0 and it was working fine as a spring mvc application. But since the application is converted to Spring boot 2, the executions fails with a weird ClassNotFoundException for primitive types.

Here is a sample exception.

Caused by: org.activiti.engine.ActivitiException: Couldn't deserialize object in variable 'executionObject'
        at org.activiti.engine.impl.variable.SerializableType.deserialize(SerializableType.java:112)
        at org.activiti.engine.impl.variable.DeserializedObject.flush(DeserializedObject.java:43)
        at org.activiti.engine.impl.db.DbSqlSession.flushDeserializedObjects(DbSqlSession.java:700)
        at org.activiti.engine.impl.db.DbSqlSession.flush(DbSqlSession.java:576)
        at org.activiti.engine.impl.interceptor.CommandContext.flushSessions(CommandContext.java:211)
        at org.activiti.engine.impl.interceptor.CommandContext.close(CommandContext.java:137)
        at org.activiti.engine.impl.interceptor.CommandContextInterceptor.execute(CommandContextInterceptor.java:66)
        at org.activiti.spring.SpringTransactionInterceptor$1.doInTransaction(SpringTransactionInterceptor.java:47)
        at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:140)
        at org.activiti.spring.SpringTransactionInterceptor.execute(SpringTransactionInterceptor.java:45)
        at org.activiti.engine.impl.interceptor.LogInterceptor.execute(LogInterceptor.java:37)
        at org.activiti.engine.impl.cfg.CommandExecutorImpl.execute(CommandExecutorImpl.java:40)
        at org.activiti.engine.impl.cfg.CommandExecutorImpl.execute(CommandExecutorImpl.java:35)
        at org.activiti.engine.impl.RuntimeServiceImpl.startProcessInstanceByKey(RuntimeServiceImpl.java:77)
        at com.tcl.gvg.itsm.activiti.ActivitiServiceImpl.executeActivity(ActivitiServiceImpl.java:518)
        ... 199 more
Caused by: org.activiti.engine.ActivitiClassLoadingException: Class not found: int
        at org.activiti.engine.impl.util.ReflectUtil.loadClass(ReflectUtil.java:87)
[...]

PR #1909, opened 2018-06-05
added fix for class loader not handling primitive types

raja-anbazhagan:
In regards to issue #1905 

I added  a new private method `forNamePrimitive()` which checks if the class to be loaded is a primitive type. 

The code has been ported from tomcat-embedded project `org.apache.el.util.ReflectionUtil.java`. [Similar workaround is done in similar way in many other popular frameworks like spring.](https://stackoverflow.com/a/11284822/2557818) .

### DIFF
--- a/modules/activiti-engine/src/main/java/org/activiti/engine/impl/util/ReflectUtil.java
+++ b/modules/activiti-engine/src/main/java/org/activiti/engine/impl/util/ReflectUtil.java
@@ -38,7 +38,22 @@ public abstract class ReflectUtil {
 
   private static final Pattern GETTER_PATTERN = Pattern.compile("(get|is)[A-Z].*");
   private static final Pattern SETTER_PATTERN = Pattern.compile("set[A-Z].*");
+  private static final String[] PRIMITIVE_NAMES = new String[] { "boolean",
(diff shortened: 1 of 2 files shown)


raja-anbazhagan:
I closed the issue #1905 with appropriate details.

However, it bothers me why the  default implementation of  `ObjectInputStream.resolveClass(ObjectStreamClass desc)` considers loading primitive types. 

    protected Class<?> resolveClass(ObjectStreamClass desc)
        throws IOException, ClassNotFoundException
    {
        String name = desc.getName();
        try {
            return Class.forName(name, false, latestUserDefinedLoader());
        } catch (ClassNotFoundException ex) {
            Class<?> cl = primClasses.get(name);
            if (cl != null) {
                return cl;
            } else {
                throw ex;
            }
        }
    }

But the overridden resolveClass in `SerializableType.createObjectInputStream()`

```
 protected ObjectInputStream createObjectInputStream(InputStream is) throws IOException {
    return new ObjectInputStream(is) {
      protected Class<?> resolveClass(ObjectStreamClass desc) throws IOException, ClassNotFoundException {
        return ReflectUtil.loadClass(desc.getName());
      }
    };
  }
```

does not support loading primitive types. Why?

PR #1973, opened 2018-08-22
1972 add connector model

ffazzini:


activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,81 @@
...
+import org.activiti.model.connector.Connector;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ConnectorService {
+
+    @Value("${connector.root:connector}")
```
Salaboy: @ffazzini let's call this activiti.connectors.dir (I do accept other suggestions, but we need to have activiti. at the beginning.. and activiti.connectors might be a good name for something around connectors.

activiti-spring-connector/src/test/java/org/activiti/spring/connector/EmptyConnectorServiceIT.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void emptyConnectors() throws IOException {
+
+        List<Connector> connectors = connectorService.get();
+        assertNotNull(connectors);
+        assertEquals(0, connectors.size());
+    }
+
+    @org.springframework.context.annotation.Configuration
```
Salaboy: @ffazzini do we need the FQN here? 
Also this highlight the need for an AutoConfiguration, can we make sure that we add one?

activiti-spring-connector/src/test/resources/connector/connector.json
```diff
@@ -0,0 +1,19 @@
+{
+	"name": "Name-of-the-connector",
```
Salaboy: @ffazzini  formatting?

ffazzini: sure

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: sounds better yes

activiti-spring-connector/src/test/java/org/activiti/spring/connector/EmptyConnectorServiceIT.java
ffazzini: If it is possible i would just use autoconfiguration rather than componentscan, about annotation or not i don't see much improvement as it still needs manual import by the dev. I would like something more transparent like an autoconfiguration. I will investigate how to do that

activiti-spring-connector/src/test/resources/connector/connector.json
ffazzini: done

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: done

activiti-spring-connector/src/test/java/org/activiti/spring/connector/ConnectorServiceIT.java
```diff
@@ -0,0 +1,60 @@
...
+        assertNotNull(connectors);
+        assertEquals(1, connectors.size());
+        assertEquals("Name-of-the-connector", connectors.get(0).getName());
+        assertEquals("Description of the connector", connectors.get(0).getDescription());
+        assertEquals("path to icon", connectors.get(0).getIcon());
+        assertEquals(2, connectors.get(0).getActions().size());
+        assertEquals("input-variable-name-1", connectors.get(0).getActions().get("actionName1").getInput().get(0).getName());
+    }
+
+//    @org.springframework.context.annotation.Configuration
```
Salaboy: @ffazzini let's not commit comments, if they are not needed just remove them.

activiti-spring-connector/src/test/java/org/activiti/spring/connector/EmptyConnectorServiceIT.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    public void emptyConnectors() throws IOException {
+
+        List<Connector> connectors = connectorService.get();
+        assertNotNull(connectors);
+        assertEquals(0, connectors.size());
+    }
+
+//    @org.springframework.context.annotation.Configuration
```
Salaboy: @ffazzini same here..

activiti-spring-connector/src/test/java/org/activiti/spring/connector/MultipleConnectorServiceIT.java
```diff
@@ -0,0 +1,59 @@
...
+     **/
+    @Test
+    public void connectors() throws IOException {
+
+        List<Connector> connectors = connectorService.get();
+        assertNotNull(connectors);
+        assertEquals(3, connectors.size());
+    }
+
+//    @org.springframework.context.annotation.Configuration
```
Salaboy: @ffazzini same here.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/autoconfigure/ConnectorAutoConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+package org.activiti.spring.connector.autoconfigure;
+
+import org.activiti.spring.connector.ConnectorService;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ConnectorAutoConfiguration {
+
+    @Bean
```
Salaboy: @ffazzini instead of returning a bean, this guy can add a @ComponentScan here with it's own package. Does it make sense? we need to follow that convention every time that we have a @Component or a @Service.

activiti-spring-connector/src/test/java/org/activiti/spring/connector/ConnectorServiceIT.java
```diff
@@ -0,0 +1,60 @@
...
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = ConnectorAutoConfiguration.class, webEnvironment = SpringBootTest.WebEnvironment.NONE)
+@TestPropertySource(locations = "classpath:application-single-test.properties")
+public class ConnectorServiceIT {
+
+    @Autowired
+    private ConnectorService connectorService;
+
+    @Test
+    public void connector() throws IOException {
```
Salaboy: @ffazzini  I know that @erdemedeiros would ask for a more meaningful name.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,81 @@
...
+import org.activiti.model.connector.Connector;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ConnectorService {
+
+    @Value("${activiti.connectors.dir:connector}")
```
Salaboy: @ffazzini the default should be "connectors" (in plural) as we have processes as the default dir for processes.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/autoconfigure/ConnectorAutoConfiguration.java
ffazzini: Don't understand, can you code what you mean? Thanks.
Talking with @ryandawsonuk earlier i think you guys have in mind to refactor the existing design you have
which has a mix of autoconfiguration and componentscans?

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: ok

activiti-spring-connector/src/test/java/org/activiti/spring/connector/ConnectorServiceIT.java
ffazzini: i will change it yes, but the @testpropertysource annotation indicates the objective of testing a single connector file

activiti-spring-connector/src/test/java/org/activiti/spring/connector/ConnectorServiceIT.java
ffazzini: yes they won't be merged don't worry, it was just to highlight you the difference between the first version using componentscan and the second one using autoconfiguration

activiti-spring-connector/src/main/java/org/activiti/spring/connector/autoconfigure/ConnectorAutoConfiguration.java
Salaboy: @ffazzini 👍 sorry for the delay, that was exactly what I meant!

activiti-spring-connector/src/test/java/org/activiti/spring/connector/ConnectorServiceIT.java
```diff
@@ -0,0 +1,60 @@
...
+
+        List<Connector> connectors = connectorService.get();
+        assertNotNull(connectors);
```
erdemedeiros: Please use AssertJ assertions instead of JUnit ones:
`assertThat(connectors).isNotNull()`, actually this line can be omitted as `hasSize` is already checking for nulls
`assertThat(connectors).hasSize(1)`
`assertThat(connectors.get(0).getName()).isEqualsTo("Name-of-the-connector")`

ffazzini: yes i can do that no problem, the dependency is already included in spring test starter i think

activiti-spring-connector/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>activiti-spring-connector</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.activiti</groupId>
+      <artifactId>activiti-connector-model</artifactId>
+      <version>${version}</version>
```
erdemedeiros: Should go under dependency management before merge.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Optional;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.activiti.model.connector.ConnectorDefinition;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+import org.springframework.stereotype.Service;
+
+@Service
```
erdemedeiros: It's probably better to instantiate the service from an auto-configuration using the constructor. This avoid the need to have this package as a target of spring scan.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,81 @@
...
+public class ConnectorService {
+
+    @Value("${activiti.connectors.dir:connectors}")
+    private String connectorRoot;
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
erdemedeiros: Here it's probably better to autowire (via constructor) the ObjectMapper, so that we know that the same configuration is used across the engine.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,81 @@
...
+        Optional<File[]> connectorFiles = Optional.empty();
+
+        Resource connectorRootPath = new ClassPathResource(connectorRoot);
+        if (connectorRootPath.exists()) {
+            connectorFiles = Optional.ofNullable(connectorRootPath.getFile().listFiles(new FilenameFilter() {
+
+                                                     @Override
+                                                     public boolean accept(File dir,
+                                                                           String name) {
+                                                         if (name.toLowerCase().endsWith(".json")) {
```
erdemedeiros: This block can be simplified to `return name.toLowerCase().endsWith(".json")`

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,81 @@
...
+                                                             return false;
+                                                         }
+                                                     }
+                                                 })
+            );
+        }
+        return connectorFiles;
+    }
+
+    private ConnectorDefinition save(File file) throws IOException {
```
erdemedeiros: I'm a little bit confused by the method name. I was expecting to have the file saved somewhere, maybe `read` would be a better name. What do you think?

activiti-spring-connector/src/main/java/org/activiti/spring/connector/autoconfigure/ConnectorAutoConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package org.activiti.spring.connector.autoconfigure;
+
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ComponentScan(basePackages = "org.activiti.spring.connector")
```
erdemedeiros: Any special reason for adding a `@componentScan` here?. In our recent developments we are rather instantiating beans programatically.

activiti-spring-connector/pom.xml
ffazzini: i needed that to code locally, you want me to remove it now or later?

activiti-spring-connector/src/main/java/org/activiti/spring/connector/autoconfigure/ConnectorAutoConfiguration.java
ffazzini: we discussed this already, have a look here: 
https://github.com/Activiti/Activiti/pull/1973/files/6e31e7d82698cfb8714514092d9f2f7c28471593#r213432583

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: what autoconfiguration should i import to include the objectmapper bean?
Otherwise i will get an unsatisified dependency exception just by replacing the line above
with constructor

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: done

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: done

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
erdemedeiros: Hmm, interesting, I tough that this was provided out of the box by Spring, but maybe it's only provided by the REST layer. So I'd suggest to create a new bean annotated with `@ConditionalOnMissingBean` in the autoconfiguration.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/autoconfigure/ConnectorAutoConfiguration.java
erdemedeiros: Sorry, man, I missed this conversation.

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
ffazzini: pushed

daisuke-yoshimoto:
@Salaboy @ffazzini 

Why do we define ConnectorDefinition with JSON?
If define it in JSON, cannot we set the ConnectorDefinition with the Process modeler GUI?

How about setting ConnectorDefinition on xml as extensionElements of BPMN?

activiti-spring-connector/src/main/java/org/activiti/spring/connector/ConnectorService.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.stereotype.Service;
+
+import java.io.File;
+import java.io.FilenameFilter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+@Service
```
erdemedeiros: I think the `@Service` is no longer necessary now that you are instantiating it via constructor in the auto-configuration.

ffazzini: True

Issue #1994, opened 2018-09-10
Create pipelines for Activiti Frameworks


salaboy:
This work includes the refactoring of the following repositories (branch: salaboy-1994-CI-CD):
- [activiti-build](http://gitHub.com/activiti/activiti-build)
- [activiti-api](http://gitHub.com/activiti/activiti-api)
- [activiti](http://gitHub.com/activiti/activiti)
- [activiti-dependencies](http://gitHub.com/activiti/activiti-dependencies)
- [activiti-examples](http://gitHub.com/activiti/activiti-examples)

We are moving away from using Maven SNAPSHOTS in favor of fixed releases using Semantic Versioning. Frameworks will be under 7.x.x versioning. 

The outcome of this task should be to publish this artifacts to nexus from our JX environment. 
This work is done based on a PoC by @ryandawsonuk and @salaboy

PR #1996, opened 2018-09-11
Final Pom Alignment for CI/CD

salaboy:
This PR contains changes to implement #1994 first iteration

### DIFF
--- a/activiti-api-impl/activiti-api-process-model-impl/pom.xml
+++ b/activiti-api-impl/activiti-api-process-model-impl/pom.xml
@@ -19,7 +19,7 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.activiti.api</groupId>
+    <groupId>org.activiti</groupId>
     <artifactId>activiti-api-impl</artifactId>
     <version>7.0.0-SNAPSHOT</version>
   </parent>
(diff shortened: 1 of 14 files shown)


activiti-api-impl/activiti-api-process-model-impl/pom.xml
```diff
@@ -19,7 +19,7 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.activiti.api</groupId>
+    <groupId>org.activiti</groupId>
```
erdemedeiros: @Salaboy this make me think about the discussion we had earlier about naming group ids based on the repository where they are.
I don't have a strong feeling about it, but IMO it would make sense that the model interface and the model implementation have the same group id. If we follow the repository approach they will be different:

```xml
     <dependency>
       <groupId>org.activiti.api</groupId>
       <artifactId>activiti-api-process-model</artifactId>
     </dependency>
```
vs
```xml
     <dependency>
       <artifactId>activiti-api-process-model-impl</artifactId>
     </dependency>
```

Salaboy: Yes, they are different in here.. I do believe that we get more by having different groupIds that having them aligned. I will advice to start with this approach to see if bother us or if it works well. I cannot think of any downside of not having the interface and impl under a different groupId, for me it makes it clear that one is the implementation and the other the interface.

PR #1998, opened 2018-09-12
Adds indexes on ACT_RU_INTEGRATION foreign keys

jtama:
Adds indexes on ACT_RU_INTEGRATION foreign keys to prevent deadlock caused by lock table mode on delete operations


Salaboy:
@jtama-op can you explain the main reason behind this PR?

jtama-op:
You should check creation scripts for other SGBD, you will see they have already diverged...

See https://github.com/Activiti/Activiti/commit/22f6c8d0cf126320addb90f86495e89e7d656939.

As for the scenario, we are launching many processes in parallel with a max async executor set to 20. Deadlock where happening all the time, we checked oracle trace files, added the indexed, relaunched activiti, and the deadlock had disappeared.

As for the consistency, I do not want to commit code for database i can't test which is the case for the other SGBS.

jtama-op:
Hi,
We are not using cloud features, all the test takes place in one JVM.

Issue #2010, opened 2018-09-15
Unable to obtain process instance variables if one of them is serializable object


mkawczynski07:
I created process instance with this code:
```
    @Timed
    @GetMapping("/start")
    public ResponseEntity<ProcessInstance> start() throws URISyntaxException {
        Offer offer = new Offer();
        StartProcessPayload payload = ProcessPayloadBuilder
                .start()
                .withProcessDefinitionKey(PROCESS_NAME)
                .withProcessInstanceName("New offer")
                .withVariable("offer", offer)
                .withVariable("owner", SecurityUtils.getCurrentUserLogin().get())
                .build();
        ProcessInstance instance = processRuntime.start(payload);
        return ResponseEntity.created(new URI("/api/process/start/" + instance.getId()))
                .headers(HeaderUtil.createEntityCreationAlert(PROCESS_NAME, instance.getId()))
                .body(instance);
    }
```
Offer is simple POJO with name and description. I try to get process instance variables in this way:
```
        GetVariablesPayload payload = new GetVariablesPayload(processInstanceId);
        List<VariableInstance> variables = processRuntime.variables(payload);
```

Unfortunately, I get an error:
```
java.lang.NullPointerException: null
        at org.activiti.engine.impl.persistence.entity.ByteArrayRef.ensureInitialized(ByteArrayRef.java:87)
        at org.activiti.engine.impl.persistence.entity.ByteArrayRef.getBytes(ByteArrayRef.java:40)
        at org.activiti.engine.impl.persistence.entity.VariableInstanceEntityImpl.getBytes(VariableInstanceEntityImpl.java:101)
        at org.activiti.engine.impl.variable.ByteArrayType.getValue(ByteArrayType.java:31)
        at org.activiti.engine.impl.variable.SerializableType.getValue(SerializableType.java:59)
        at org.activiti.engine.impl.persistence.entity.VariableInstanceEntityImpl.getValue(VariableInstanceEntityImpl.java:124)
        at org.activiti.runtime.api.model.impl.APIVariableInstanceConverter.from(APIVariableInstanceConverter.java:29)
        at org.activiti.runtime.api.model.impl.APIVariableInstanceConverter.from(APIVariableInstanceConverter.java:21)
        at org.activiti.runtime.api.model.impl.ListConverter.from(ListConverter.java:27)
        at org.activiti.runtime.api.impl.ProcessRuntimeImpl.variables(ProcessRuntimeImpl.java:276)
[...]

PR #2046, opened 2018-10-03
Fix issue 2010

daisuke-yoshimoto:
Fixed the following issue.
https://github.com/Activiti/Activiti/issues/2010

### DIFF
--- a/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/GetExecutionVariableInstanceCmd.java
+++ b/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/GetExecutionVariableInstanceCmd.java
@@ -1,9 +1,9 @@
 /* Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


balsarori:
I don't think it's a good idea to always fetch the values. There could be 100s for variables with unknown  size and fetching could cause issues.

Probably making it optional whether to fetch values would be better (default false).

daisuke-yoshimoto:
@balsarori 
After all it is necessary to fetch to access value.
So if we try to fetch in the command like this fix, it is same performance as fetching in the calling code.

balsarori:
@daisuke-yoshimoto that's true. Old REST API didn't fetch the values and provided a different endpoint for fetching values separately. I don't think it's a good idea to fetch by default and have no other option. 

I would suggest to keep the old behaviour and introduce a new method in RuntimeService that allows fetching variables with there values. Caller can decide which one to use.

daisuke-yoshimoto:
@balsarori 

Thanks. Make sense.

At the moment, this pull request is the fix of bugs of Activiti Engine itself, and secondly to solve the problem of issue 2010 as well. So, I think that this fix itself can be merged.

But, your proposal is not a fix but a spec change, so how about implementing it at the activiti-api layer in the future?
https://github.com/Activiti/activiti-api/blob/develop/activiti-api-process-runtime/src/main/java/org/activiti/api/process/runtime/ProcessRuntime.java

balsarori:
**Note** that variable values could be binary (i.e. a stream of bytes). I don't think we want to always fetch such values.

It makes sense to add this in the API layer and provide a way to fetch the value when it's not of primitive type.

daisuke-yoshimoto:
@balsarori 

> Note that variable values could be binary (i.e. a stream of bytes).

Storing the binary of the large file is not a use case of variables.
First of all, the feature itself to store and retrieve variables via streams will not be provided.

If we need to store the binary of the large file, we should use AttachmentService.
TaskService.createAttachment/getAttachmentContent
https://www.activiti.org/javadocs/org/activiti/engine/taskservice#getAttachmentContent-java.lang.String-
https://www.activiti.org/javadocs/org/activiti/engine/taskservice#createAttachment-java.lang.String-java.lang.String-java.lang.String-java.lang.String-java.lang.String-java.io.InputStream-

> I don't think we want to always fetch such values.

As you say, it is best to fetch binary at the timing when caller trigger getValue method, but it is structurally difficult.
VariableInstance is just a POJO outside the command. So, it is difficult for the Caller to fetch when calling the getValue method outside the command.

In addition, lots of APIs that acquire variables in Activiti Engine have the same behavior as this fix that binary is fetched beforehand in the command. So, I think that as a policy of current revision, this is fine.

For examples.
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/GetExecutionsVariablesCmd.java#L51
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/GetTasksLocalVariablesCmd.java#L52
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/HistoricVariableInstanceQueryImpl.java#L229
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityImpl.java#L819
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/HistoricProcessInstanceEntityImpl.java#L218
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/HistoricTaskInstanceEntityImpl.java#L249
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityImpl.java#L557
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessInstanceHistoryLogQueryImpl.java#L117

balsarori:
@daisuke-yoshimoto 
my point is that caller should have an option whether or not to fetch values from external table (in case they aren't primitive type). 

It might be that the caller isn't interested in fetching the values and is only interested in knowing basic variables information (i.e. name, type, taskId, process instance id, etc! which is what differentiates ```RuntimeService.getVariables``` from ```RuntimeService.getVariableInstances```) of the variables.

This is a valid use case where only variables info are listed and values are fetched as needed.

daisuke-yoshimoto:
@balsarori 
I understand your point, but your proposal is function addition.
This pr is only fixing existing bugs like other parts.

* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/ProcessInstanceHistoryLogQueryImpl.java#L117
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/GetExecutionsVariablesCmd.java#L51
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/GetTasksLocalVariablesCmd.java#L52
* https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/HistoricVariableInstanceQueryImpl.java#L229

balsarori:
@daisuke-yoshimoto 
Here there is an option to not initialize variable values i.e.  
https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/HistoricVariableInstanceQueryImpl.java#L224

GetExecutionsVariablesCmd and GetTasksLocalVariablesCmd (called by ``` RuntimeService.getVariables``` and ```taskService.getVariables```) are used to fetch the variable values so it's normal to always fetch the values.

ProcessInstanceHistoryLogQueryImpl.java doesn't have an option to fetch values but including the vars is optional.

balsarori:
@daisuke-yoshimoto 
I'd rather have a different method instead of overloading the existing one i.e. ```getInitializedVariableInstances``` but for the command you can add a boolean to determine whether or not to initialize the values.

daisuke-yoshimoto:
@balsarori 
I will add flag excludeVariableInitialization like HistoricVariableInstanceQuery. So, in default mode, query fetch serializable variables. If only explicitly excludeVariableInitialization is true, query don't fetch serializable variables.

balsarori:
@daisuke-yoshimoto 
I had a quick discussion about this with @salaboy and we think we shouldn't change the default behaviour. It's better to introduce a new method i.e. ```getInitializedVariableInstances``` which initializes the values. You can still use the same ```GetExecutionVariableInstancesCmd``` with a boolean that determines whether or not to fetch values (by default doesn't fetch values).

daisuke-yoshimoto:
@balsarori @salaboy 

Ok. So, I will fix as the following as follows.

1. Add getInitializedVariableInstances to RuntimeService
-> add flag to GetExecutionVariableInstancesCmd

2. Add flag to ProcessRuntime.variables
-> Add flag to GetVariablesPayload
https://github.com/Activiti/activiti-api/blob/develop/activiti-api-process-model/src/main/java/org/activiti/api/process/model/payloads/GetVariablesPayload.java

daisuke-yoshimoto:
@balsarori @salaboy 

Well, after all, I tried reconsidering, but I cannot understand the behavior of not initializing object variables(fetching object variables) by default. By default , initializing object variables(fetching object variables) should executed in commands. Optionally Operations that are not  initialized(fetced) should be provided.

### Performance

* I think that performance concerns are not so much even in the operation of fetching by default.
* Since the part that I fixed in this pr is an API that individually specifies the taskId or executionId, it is assumed that there are not so many variables. If the number of variables is not large, there should be not much overhead at the time of fetching.

### Consistency with other parts

* Lots of APIs that acquire variables in Activiti Engine have the same behavior as this fix that binary is fetched beforehand in the command. So, I think that as a policy of current revision, this is fine.

* Also, for HistoricVariableInstanceQuery which you presented, the option is false by default. By default, HistoricVariableInstanceQuery fetch object variables in command.
  * https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/history/HistoricVariableInstanceQuery.java#L56

salaboy:
@daisuke-yoshimoto you are right, this is really important.. I will review this now.. 
I'm trying to understand what was the decision at the end.. if we are going to fetch serialised variables by default or not.

daisuke-yoshimoto:
@salaboy 
I summarized. 
https://github.com/Activiti/Activiti/issues/2010#issuecomment-430327242

daisuke-yoshimoto:
@salaboy 
Any updates?

daisuke-yoshimoto:
@salaboy Thanks.

PR #2056, opened 2018-10-10
Add unit tests generated by Diffblue Deeptest

diffblue-assistant:
The test cases contained in this PR were automatically generated by [Diffblue's](http://www.diffblue.com) [Deeptest](http://www.diffblue.com/papers/overview.pdf) software. 

Diffblue is an Oxford-based start-up that aims to automate traditional coding tasks to increase developer productivity. One of our missions is to provide useful tests to open source projects. In return, we would appreciate your feedback about our tests to help us further improve our tool. 
    
These tests are part of a batch of 3015 in total, providing an increase of 12.54% in line coverage. We would be happy to provide the remaining tests via a subsequent PR.

Results for various open source projects can be viewed at:

   [http://diff.blue/demo](http://diff.blue/demo)

We welcome any feedback you have on these tests.


salaboy:
@erdemedeiros comments on this one? I would say that we shouldn't accept automated tests on random classes, even if they increase the coverage.

ryandawsonuk:
@salaboy I agree

diffblue-assistant:
How would you like us to contribute unit tests?

salaboy:
@diffblue-assistant it will be great to have more focused tests on things that matter for example, adding tests on modules such as https://github.com/Activiti/Activiti/tree/develop/activiti-bpmn-converter/ and https://github.com/Activiti/Activiti/tree/develop/activiti-bpmn-model will be more useful

diffblue-assistant:
Thanks for your suggestion. We'll have a look into these.

erdemedeiros:
@diffblue-assistant I'm a big fun of automation and code generation in general, but I'm not very confident with auto-generated tests. This sounds wrong when you are used to apply Test Driven Development where you are suppose to write your tests before coding your implementations: so it could not be autogenerated. Maybe I missed something, but how do you see interaction of you tool and TDD approach?
I can imagine that it can help to decrease the technical dept on legacy code by increasing the code coverage on that part, but still if you haven't written tests from the beginning your code is probably not testable, so test generation cannot do much for you.

Another detail: we are currently using AssertJ for our test assertions so we avoid adding new tests using default JUnit assertions. Is that possible to configure which assertion library to use while generating tests?

diffblue-assistant:
While TDD with auto-generated tests sounds counter-intuitive at first, we have seen such a process implemented in real-world projects. The main advantage lies in the ability for auto-generated testing to save huge amounts of time writing the actual template code for a TDD test. Running the signature of a method, without any logic implemented, through the tool will output essentially a pre-set up test. Simply changing the parameters (often simple value changes) as required will leave you with a series of tests which can be used as benchmarks for TDD development for the method in question.

We produced a video on this very subject which can be viewed [here](https://youtu.be/Tz_BGTKpOIk) if you are interested in seeing this in depth and in practice.

erdemedeiros:
@diffblue-assistant thank you for you explanation. Indeed, the interaction with Deeptest seems to be more interactive than I initially tough. I guess the best way to make my own opinion is by getting my hands dirty when I have some time.

Any chance to replace JUnit assertions by AssertJ assertions?

salaboy:
closing this pr due the lack of feedback

Issue #1991, opened 2018-09-06
Query - Task is not updated


lucianoprea:
When a task is updated in runtime-bundle service with success the changes are not also propagated to query service.

magemello:
@salaboy is this bug fixed?

mauriziovitale:
The bug is fixed with Beta5. I was able to update the `desc` and `priority` properties of a task but wasn't able to update the `name` value. Should it be possible? @salaboy

PR #2065, opened 2018-10-17
ACTIVITI-1991 Task is not updated in Query

lucianoprea:
-added listener delegate for TaskUpdate event
-created a test for this listener

refs [1991](https://github.com/Activiti/Activiti/issues/1991)

### DIFF
--- a/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/ToAPITaskUpdatedEventConverter.java
+++ b/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/ToAPITaskUpdatedEventConverter.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2018 Alfresco, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.activiti.runtime.api.event.impl;
+
+import java.util.Optional;
+
+import org.activiti.api.task.runtime.events.TaskUpdatedEvent;
+import org.activiti.engine.delegate.event.ActivitiEntityEvent;
+import org.activiti.engine.task.Task;
+import org.activiti.runtime.api.model.impl.APITaskConverter;
+
+public class ToAPITaskUpdatedEventConverter implements EventConverter<TaskUpdatedEvent, ActivitiEntityEvent> {
+
+    private final APITaskConverter taskConverter;
+
+    public ToAPITaskUpdatedEventConverter(APITaskConverter taskConverter) {
+        this.taskConverter = taskConverter;
+    }
+
+    @Override
+    public Optional<TaskUpdatedEvent> from(ActivitiEntityEvent internalEvent) {
+        TaskUpdatedEventImpl event = null;
+        if (isTaskEvent(internalEvent)) {
+            event = new TaskUpdatedEventImpl(taskConverter.from((Task) internalEvent.getEntity()));
+        }
+        return Optional.ofNullable(event);
+    }
+
+    private boolean isTaskEvent(ActivitiEntityEvent internal) {
+        return internal.getEntity() instanceof Task;
+    }
+}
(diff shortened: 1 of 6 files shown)


activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/ToAPITaskUpdatedEventConverter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private final APITaskConverter taskConverter;
+
+    public ToAPITaskUpdatedEventConverter(APITaskConverter taskConverter) {
+        this.taskConverter = taskConverter;
+    }
+
+    @Override
+    public Optional<TaskUpdatedEvent> from(ActivitiEntityEvent internalEvent) {
+        return Optional.of(new TaskUpdatedEventImpl(taskConverter.from((org.activiti.engine.task.Task) internalEvent.getEntity())));
```
erdemedeiros: Here you need to check before if it's a Task update or an update of another entity. As you are using `ActivitiEventType.ENTITY_UPDATED` to register the listener it's listening for all the entity updates, not only for tasks. See https://github.com/Activiti/Activiti/blob/develop/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/ToTaskSuspendedConverter.java#L37.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeUpdateTaskTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.springframework.security.test.context.support.WithUserDetails;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.*;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE)
+@ContextConfiguration
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
erdemedeiros: We definitely need to find a better way of doing this. I don't like this FixMethodOrder, but for now that's the only way we got it working with `@WithUserDetails` annotation.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeUpdateTaskTest.java
```diff
@@ -0,0 +1,153 @@
...
+                                                         .withPriority(50)
+                                                         .withAssignee(authenticatedUserId)
+                                                         .build());
+
+        assertThat(RuntimeTestConfiguration.createdTasks).contains(standaloneTask.getId());
+
+        Page<Task> tasks = taskRuntime.tasks(Pageable.of(0,
+                                                         50));
+
+        assertThat(tasks.getTotalItems()).isEqualTo(1);
```
erdemedeiros: IMO this assertion is not necessary.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeUpdateTaskTest.java
```diff
@@ -0,0 +1,153 @@
...
+        // claim
+        taskRuntime.claim(TaskPayloadBuilder.claim().withTaskId(standaloneTask.getId()).build());
+
+        // update
+        final Task updatedTask = taskRuntime.update(updateTaskPayload);
+        tasks = taskRuntime.tasks(Pageable.of(0,
+                                              50));
+
+        assertThat(RuntimeTestConfiguration.updatedTasks).contains(updatedTask.getId());
+        assertThat(tasks.getTotalItems()).isEqualTo(2);
```
erdemedeiros: Same here, I don't think you need to check the total elements.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeUpdateTaskTest.java
lucianoprea: I don't like this either but this way are configured all tests for tasks due to the user scenarios I believe.

erdemedeiros: yeah, unfortunately we'll need to live with that until we figure out how to switch of user when using `@WithUserDetails`.

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/ToAPITaskUpdatedEventConverter.java
lucianoprea: I missed that, makes total sense to have this check.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeUpdateTaskTest.java
lucianoprea: I removed it.

lucianoprea:
@erdemedeiros I did the changes requests but github is not showing the latests changes. I've rebased my branch on latest develop

Issue #2047, opened 2018-10-04
Read Modeller Zip file from Runtime Bundle and core Activiti Deployer


salaboy:
Enable the Runtime Bundle core services and the internal engine deployer to read the zip file coming from the export feature of modeler app.

salaboy:
done in develop

PR #2067, opened 2018-10-17
[WIP] Import processes from application zip file

erdemedeiros:
First draft, do not merge yet.


erdemedeiros:
Refers to https://github.com/Activiti/Activiti/issues/2047

salaboy:
@erdemedeiros this is really good.. basically what I was expecting for an initial iteration. 
The only thing that I will notice here is that based on our latest refactorings we should move activiti-applications to activiti-core-common and rename it to activiti-spring-applications (due its dependency on autoconfigurations and spring annotations). Also, is there a class that represent the Application itself? 

In the same way as connectors, when we read an application from /src/main/resources/applications/ we should have a Bean, called ApplicationsService which collects the names of the application zip files (the only thing that we have so far). Does that sounds like doable this week? if not I am happy to move forward with this loading bits first and add the ApplicationsService later on.

erdemedeiros:
@salaboy yeah, it makes sense to move it to core common. I'll do that.

Concerning the ApplicationsService I think it just a matter of renaming things: currently I have `ApplicationContent` that can be renamed to `Application` and add the zip folder name to it. And I have `ApplicationLoader` that can be renamed to ApplicationsService.

salaboy:
@erdemedeiros great! please do that renaming

PR #2090, opened 2018-10-29
Add test in CallActivity

miguelruizdev:
New test covering the case when the process definitions have been modeled with the new activiti-modeling-app

### DIFF
--- a/activiti-engine/src/test/java/org/activiti/examples/bpmn/callactivity/CallActivityTest.java
+++ b/activiti-engine/src/test/java/org/activiti/examples/bpmn/callactivity/CallActivityTest.java
@@ -87,4 +87,17 @@ public class CallActivityTest extends PluggableActivitiTestCase {
     subProcessInstance = runtimeService.createProcessInstanceQuery().superProcessInstanceId(pi.getId()).singleResult();
     assertEquals("123", subProcessInstance.getBusinessKey());
   }
+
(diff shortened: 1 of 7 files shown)


activiti-engine/src/test/java/org/activiti/examples/bpmn/callactivity/CallActivityTest.java
```diff
@@ -87,4 +87,17 @@ public void testCallActivityWithBusinessKey() {
     subProcessInstance = runtimeService.createProcessInstanceQuery().superProcessInstanceId(pi.getId()).singleResult();
     assertEquals("123", subProcessInstance.getBusinessKey());
   }
+
+  @Deployment(resources = {"org/activiti/examples/bpmn/callactivity/parentProcess.bpmn20.xml",
+                            "org/activiti/examples/bpmn/callactivity/subProcess.bpmn20.xml"})
+  public void testCheckSubProcessTaskWhenCallActivity(){
```
salaboy: @miguelruizdev this is good.. but we need to have a test with the new APIs, if we don't do that we will never cover this use case at that level. NEW API = tests in here: https://github.com/Activiti/Activiti/tree/develop/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot
Please add such test in this PR

Issue #2079, opened 2018-10-24
ActivitiObjectNotFoundException: Form with formKey 'repair.form' does not exist


Shaun-Yao:
I using spring boot v2.0.6, activiti v6.0.0
`<startEvent id="testRepairSt" name="StartEvent" activiti:formKey="repair.form" />`
Object renderedStartForm = formService.getRenderedStartForm(processDefinitionId);
<img width="210" alt="formkey" src="https://user-images.githubusercontent.com/18024481/47690700-024ae500-dc2a-11e8-908b-cbf1e7b37871.png">

salaboy:
@Shaun-Yao can you create a reproducer for this problem? 
Also sharing more information about how are you deploying your process will be good.

Shaun-Yao:
@salaboy  https://github.com/Shaun-Yao/activiti-test
the processes folder would be automatically deployed, and The process will start when the project starts.

`@SpringBootApplication(exclude = SecurityAutoConfiguration.class)
public class ActivitiTestApplication implements CommandLineRunner {

    @Autowired
    private RuntimeService runtimeService;

    @Autowired
    private TaskService taskService;

    @Autowired
    private RepositoryService repositoryService;

    public static void main(String[] args) {
        SpringApplication.run(ActivitiTestApplication.class, args);
    }

    @Override
    public void run(String... args) throws Exception {
        final String processId = "testRepair";
        System.out.println("Number of process definitions : "
                + repositoryService.createProcessDefinitionQuery().count());
        System.out.println("Number of tasks : " + taskService.createTaskQuery().count());
        runtimeService.startProcessInstanceByKey(processId);

        System.out.println("Number of tasks after process start: " + taskService.createTaskQuery().count());
    }
}`

`@GetMapping("/repairForm/{processDefinitionId}")
    public String repairForm(@PathVariable("processDefinitionId") String processDefinitionId, Model model) {

        Object renderedStartForm = formService.getRenderedStartForm(processDefinitionId);
        model.addAttribute("renderedStartForm", renderedStartForm);
        model.addAttribute("processDefinitionId", processDefinitionId);

        return "repairForm";
    }
`
The issue will appear when visit http://localhost:8080/repairForm/testRepair:1:4 .

Shaun-Yao:
@salaboy Any suggestions?

salaboy:
[...]

PR #2095, opened 2018-10-30
Fix issue #2079(Change resource name)

daisuke-yoshimoto:
https://github.com/Activiti/Activiti/issues/2079

### DIFF
--- a/modules/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
+++ b/modules/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
@@ -1,9 +1,9 @@
 /* Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


modules/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
```diff
@@ -13,56 +13,54 @@
 
 package org.activiti.spring.autodeployment;
 
+imamework.core.io.Resource;
```
ryandawsonuk:
```suggestion
import org.springframework.core.io.Resource;
```

modules/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
```diff
@@ -13,56 +13,54 @@
...
+
+
+
+port java.io.IOException;
```
ryandawsonuk:
```suggestion
import java.io.IOException;
```

modules/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
```diff
@@ -13,56 +13,54 @@
...
+
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.core.io.ByteArrayResource;
 import org.springframework.core.io.ContextResource;
-import org.springframework.core.io.Resource;
-
-import java.io.IOException;
-
-/**
+import org.spring/**
```
ryandawsonuk:
```suggestion
```

salaboy:
I think this is for @balsarori to merge.

salaboy:
this is targeting 6.x so I will defer to @balsarori

daisuke-yoshimoto:
@salaboy 
Not. This is same behavior in Activiti 7.x too.
https://github.com/Activiti/Activiti/blob/develop/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java#L60

Resource name is absolute path.

salaboy:
@daisuke-yoshimoto now we have a different approach in 7.x and 6.x -> https://github.com/Activiti/Activiti/blob/develop/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java#L69

Should I merge this one as well?

daisuke-yoshimoto:
@salaboy 

In the current implementation(6.x and 7.x), the absolute path of the file is registered in the resource name. 

It should be hard to handle as an absolute path. For example.
`/Users/yoshimoto/git/activiti-cloud-runtime-bundle-service-apply-option/activiti-cloud-starter-runtime-bundle/src/test/resources/processes/AsyncErrorProcess.bpmn`

PR #2098, opened 2018-10-31
Fetch group from task payload if available

cristianflorescu:


activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -116,7 +116,8 @@ public Task task(String taskId) {
         }
         String authenticatedUserId = securityManager.getAuthenticatedUserId();
         if (authenticatedUserId != null && !authenticatedUserId.isEmpty()) {
-            List<String> userGroups = userGroupManager.getUserGroups(authenticatedUserId);
+            List<String> userGroups = (getTasksPayload.getGroups() == null) ?
```
salaboy: @cristianflorescu @erdemedeiros this might introduce a security risk, because the caller can add the groups in the payload and override the real groups coming from the userGroupManager. 
@erdemedeiros please be careful if you are reviewing these changes.

Issue #2092, opened 2018-10-29
When a task is completed in RB, it should show as completed in query


ryandawsonuk:
Claim and complete a task. Check the task status in `/tasks/{taskId}` in the query module. The status should be `COMPLETED`. We are currently observing a situation where it remains `ASSIGNED` in query, despite having moved to `COMPLETED` in the runtime bundle.

We do have an acceptance test for this - https://github.com/Activiti/activiti-cloud-acceptance-tests/blob/develop/runtime-acceptance-tests/src/main/java/org/activiti/cloud/qa/story/ProcessInstanceTasks.java#L174

erdemedeiros:
The problem is coming from the event sent by the runtime bundle which is inferring the status as `ASSIGNED` instead of `COMPLETED`.

PR #2105, opened 2018-10-31
Fix task status for task completed event

erdemedeiros:
Instead of trying to infer the status, just set it to completed


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/2092

Issue #1907, opened 2018-06-05
HistoricTaskInstanceQuery Issue when trying to get Task & Process Variables


gonzalocamarero:
Hi there, i found an error when I execute:

`HistoricTaskInstanceQuery taskQuery = historyService
				.createHistoricTaskInstanceQuery()
				.finished()
				.taskVariableValueEquals(Variable.ESTADO_TAREA, filter.get(Variable.ESTADO_TAREA))
				.includeTaskLocalVariables()
				.includeProcessVariables()
				.orderByHistoricTaskInstanceStartTime()
				.desc();`

It's generate the SQL like this:

`select * from 
	( 
		select a.*, ROWNUM rnum from 
			( select distinct RES.*, VAR.ID_ as VAR_ID_, VAR.NAME_ as VAR_NAME_, VAR.VAR_TYPE_ as VAR_TYPE_, VAR.REV_ as VAR_REV_, VAR.PROC_INST_ID_ as VAR_PROC_INST_ID_, VAR.EXECUTION_ID_ as VAR_EXECUTION_ID_, VAR.TASK_ID_ as VAR_TASK_ID_, VAR.BYTEARRAY_ID_ as VAR_BYTEARRAY_ID_, VAR.DOUBLE_ as VAR_DOUBLE_, VAR.TEXT_ as VAR_TEXT_, VAR.TEXT2_ as VAR_TEXT2_, VAR.LAST_UPDATED_TIME_ as VAR_LAST_UPDATED_TIME_, VAR.LONG_ as VAR_LONG_ 
				from ACT_HI_TASKINST RES 
				left outer join ACT_HI_VARINST VAR ON RES.ID_ = VAR.TASK_ID_ or RES.PROC_INST_ID_ = VAR.EXECUTION_ID_ 
				inner join ACT_HI_VARINST A0 on RES.ID_ = A0.TASK_ID_ 
				WHERE RES.PROC_INST_ID_ = 13428937 and RES.END_TIME_ is not null and A0.NAME_= 'ESTADO_TAREA' and A0.VAR_TYPE_ = 'string' and A0.TEXT_ = '4' order by RES.START_TIME_ desc, VAR.LAST_UPDATED_TIME_ asc 
			) a 
		where ROWNUM < 20001
	) 
where rnum >= 1;`

[...]

PR #2122, opened 2018-11-05
Added condition to check that a Task owns a variable before we set it…

gonzalocamarero:
… , for issue #1907


salaboy:
@gonzalocamarero look that the PR cannot be merged.. you probably need to rebase the code in order to add your commit on top of the last commit into the branch. If the PR is not green we usually cannot merge it. Also I notice that the PR is not including any test, and that is a problem, because we cannot validate that you are not breaking any other functionality with your addition.

gonzalocamarero:
@salaboy Ok, I will try to do the test, and upload a correct PR, sorry!

salaboy:
@gonzalocamarero should we close this one? or are you still working in a test for it? Notice that you will need to rebase it now

gonzalocamarero:
Please, close it, I can't do the right test by now. When I can, I try to do
it by the correct way (with test and correct tag/version). Sorry!

Issue #2049, opened 2018-10-05
Process variables extensions


balsarori:
Allow to start a process with initial variables in addition to enforcing required constraints.

PR #2195, opened 2018-11-22
Process extensions

balsarori:
#2049


activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
```diff
@@ -0,0 +1,101 @@
...
+import org.activiti.spring.process.model.VariableDefinition;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.text.ParseException;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class ProcessVariablesInitiator extends ProcessInstanceHelper {
```
salaboy: @balsarori @ryandawsonuk this class doesn't make any sense. If we want to modify the behaviour of ProcessInstanceHelper we should do so to support process variables definitions. This class hierarchy will cause us a lot of trouble so this is definitely not the way to go. We need to change this.

salaboy: What I suggest is to push this logic to ProcessInstanceHelper, which will push us to make ProcessInstanceHelper more flexible and generic.

activiti-spring-boot-starter/src/test/resources/processes/initial-vars-extensions.json
```diff
@@ -0,0 +1,34 @@
+{
+  "id":"initialVarsProcess",
```
salaboy: @balsarori @ryandawsonuk where is "id" and "name" stored in the model created by this PR? We cannot read half of the information.

salaboy: What I would suggest is to have a model for the entire file.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ExtensionVariableTypes.java
```diff
@@ -0,0 +1,38 @@
+package org.activiti.spring.process;
+
+import java.text.Format;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public enum ExtensionVariableTypes {
+
+    BOOLEAN("boolean",Boolean.class),
```
salaboy: @balsarori @ryandawsonuk having an ENUM with hardcoded types is quite limited and it will cause problems. What I would suggest is to think about other options to make sure that we can register new types.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ExtensionVariableTypes.java
```diff
@@ -0,0 +1,38 @@
...
+    STRING("string",String.class),
+    INTEGER("integer",Integer.class),
+    DATE("date",Date.class,new SimpleDateFormat("yyyy-MM-dd"));
+
+    String name;
```
salaboy: @balsarori @ryandawsonuk use the right accessor.

activiti-spring-boot-starter/src/test/resources/processes/initial-vars-extensions.json
balsarori: ``ProcessExtensionModel`` represents this file. Seems that the name is missing from that class but we don't use it. It could be added. I can check what the modeler produces and make sure that all properties a represented.

salaboy: @balsarori then all the information needs to be present and validated.. so is ID covered already?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ExtensionVariableTypes.java
ryandawsonuk: @salaboy you're thinking we could use the [variableTypes as the engine does](https://github.com/Activiti/Activiti/blob/8b90d1095c221880a43f16cdcf5f8e145dff2267/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java#L1954) so that when additional types are added by projects extending the engine then they would also be usable ?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
salaboy: @ryandawsonuk I still dislike the inheritance here. Unless we are planing to haa ProcessInstanceHelper for Cloud, having more fine grained versions of it doesn't make too much sense. The name "ProcessInstanceHelperForExtensions" implies that it will handle only Extensions, which means that if we want to add any other customization for cloud/spring boot the name will become a problem. 
@balsarori even if the engine doesn't need to know, the new APIs will need to know, so we just need to make sure that we make it available at the right level and with the right name.

activiti-spring-boot-starter/src/test/resources/processes/initial-vars-extensions.json
ryandawsonuk: ID at the top level is [in ProcessExtensionModel](https://github.com/Activiti/Activiti/blob/ca05e8b8b4825e3f70dc7ecc88e1a5a741bebb9e/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/ProcessExtensionModel.java#L18) then below that [in VariableDefintion](https://github.com/Activiti/Activiti/pull/2195/files#diff-a86fb2b9c77a756fe46931bb51c01a31R18)

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
balsarori: @salaboy  APIs don't cover all possible ways of starting a new process instance. I don't like the inheritance either so I would suggest that we have allow the ability to register process instance initiators.

salaboy: @balsarori "APIs don't cover all possible ways of starting a new process instance." we need to make sure that the supported ways would work without any extra configuration. Right now we are not interested in covering all the possible ways, we will add more ways when the basics are more stable. 
Adding more than ProcessInstanceInitiator will come with more challenges such as order and priority. If by register you mean to select which one you would like then yes.. that is more being able to configure which one to use.

activiti-spring-boot-starter/src/test/resources/processes/initial-vars-extensions.json
```diff
@@ -0,0 +1,34 @@
...
+  "name":"myProcess.bpmn",
+  "extensions":{
+    "properties":{
```
ryandawsonuk: Could this be called `variables` instead of `properties`?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
balsarori: > "we need to make sure that the supported ways would work without any extra configuration" 

that's not possible. I don't understand why shouldn't we cover all possible process starts if there is an easy hook that we can add to the engine to do that.

> Adding more than ProcessInstanceInitiator will come with more challenges such as order and priority. 

Why order? I don't think that should be a concern until a need for that arises. 

I guess we are over complicating things while they are straight forward and easy to implement. Your concerns might or might not be valid however we can always refactor if things prove to not be right later.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.activiti.spring.process.model;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Extension {
+
+    private Map<String, VariableDefinition> properties = new HashMap<>();
```
salaboy: @ryandawsonuk let's rename this to variables here.. and use Jackson annotations to map to the "properties" field in the json.. this is what Constantin is doing in modeling as well.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public Map<String, VariableDefinition> getProperties() {
```
salaboy: @ryandawsonuk so here we have getVariables() and setVariables()

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/VariableDefinition.java
```diff
@@ -0,0 +1,61 @@
...
+package org.activiti.spring.process.model;
+
+public class VariableDefinition {
+
+    private String id;
+    private String name;
+    private String description;
+    private String type;
+    private boolean required;
+    private Object value;
```
salaboy: @ryandawsonuk @balsarori this should be defaultValue right? let's make sure that we don't confuse people.

salaboy: Also.. does this push us to deserialise the default value?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/VariableParsingService.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.activiti.engine.ActivitiException;
+import org.activiti.spring.process.model.VariableDefinition;
+import org.activiti.spring.process.variable.types.ExtensionVariableType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class VariableParsingService {
```
salaboy: @ryandawsonuk @balsarori odd name..

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/VariableValidationService.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.activiti.engine.ActivitiException;
+import org.activiti.spring.process.model.VariableDefinition;
+import org.activiti.spring.process.variable.types.ExtensionVariableType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class VariableValidationService {
```
salaboy: @ryandawsonuk @balsarori also.. odd name. or maybe it is me

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/ExtensionVariableType.java
```diff
@@ -0,0 +1,24 @@
+package org.activiti.spring.process.variable.types;
+
+import java.util.List;
+
+import org.activiti.engine.ActivitiException;
+
+public abstract class ExtensionVariableType {
```
salaboy: @ryandawsonuk why so complicated? VariableType will do as a name, right?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/JavaObjectExtensionVariableType.java
```diff
@@ -0,0 +1,37 @@
+package org.activiti.spring.process.variable.types;
+
+import java.util.List;
+
+import org.activiti.engine.ActivitiException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JavaObjectExtensionVariableType extends ExtensionVariableType {
```
salaboy: @ryandawsonuk same here.. name too long and it is not adding any value: JavaVariableType will do, is there any reason why not to use that simple name?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/autoconfigure/ProcessExtensionsAutoConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+                                                                             String processExtensionsSuffix,
+                                                                   ObjectMapper objectMapper,
+                                                                   ResourcePatternResolver resourceLoader) throws IOException {
+        ProcessExtensionService processExtensionService = new ProcessExtensionService(processExtensionsRoot, processExtensionsSuffix, objectMapper, resourceLoader);
+        return processExtensionService.get();
+
+    }
+
+    @Bean
+    public Map<String, ExtensionVariableType> variableTypeMap(){
```
salaboy: @ryandawsonuk here you got it right.. variableTypeMap +1

Issue #2213, opened 2018-11-27
Connectors variable mappings


balsarori:
Enable mapping connectors input/outputs to variables.

salaboy:
done in develop.. acc tests running

PR #2227, opened 2018-12-01
Variable mappings for connectors

balsarori:
#2213 
Just basic variable mapping here for connectors. No validation (i.e. checking if mapped variable types match).


activiti-spring-process-extensions/pom.xml
```diff
@@ -24,6 +24,10 @@
             <groupId>org.activiti</groupId>
             <artifactId>activiti-spring</artifactId>
         </dependency>
+        <dependency>
```
balsarori: I don't like adding this dependency here. Would be great if [VariableDefinition](https://github.com/Activiti/activiti-core-common/blob/develop/activiti-connector-model/src/main/java/org/activiti/core/common/model/connector/VariableDefinition.java) was moved to a common module since it's not related to connectors only.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/VariableDefinition.java
```diff
@@ -13,49 +13,14 @@
 
 package org.activiti.spring.process.model;
 
-public class VariableDefinition {
+public class VariableDefinition extends org.activiti.core.common.model.connector.VariableDefinition {
```
balsarori: see above comment about moving VariableDefinition to a common module. (will also need to add a value property)

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/connector/VariablesMatchHelper.java
```diff
@@ -1,15 +1,87 @@
...
+            if (mappings.containsKey(k)) {
+                String mappedVariableId = mappings.get(v.getId());
+                VariableDefinition mappedVariableDefinition = variableDefinitionMap.get(mappedVariableId);
+                if (mappedVariableDefinition != null) {
+                    Optional<Object> outBoundVariableValueOptional = mappedValueProvider.getValue(results, v, mappedVariableDefinition);
+                    if (outBoundVariableValueOptional.isPresent()) {
+                        variables.put(mappedVariableDefinition.getName(), outBoundVariableValueOptional.get());
+                    }
+                }
+                // TODO what if the mapped variable definition wasn't found?
```
balsarori: What to do when mapping variable isn't found in connector definition? Should through exception or simple ignore? There should be a validation done before this i.e. during modelling or at startup

salaboy:
@erdemedeiros @ffazzini we need to rebase, review and understand these changes

activiti-spring-boot-starter/src/test/resources/connectors/variable-mapping.json
```diff
@@ -0,0 +1,59 @@
...
+  "id": "variableMappingConnectorId",
+  "name": "Name of the connector",
+  "description": "Description of the connector",
+  "icon": "icon-name",
+  "actions": {
+    "variableMappingActionId": {
+      "id": "variableMappingActionId",
+      "name": "variableMappingActionName",
+      "description": "...",
+      "input": [
```
ffazzini: plural inputs and outputs: from https://github.com/Activiti/Activiti/blob/develop/activiti-spring-boot-starter/src/test/resources/connectors/process-image.json#L11
and rfc doc

activiti-spring-boot-starter/src/test/resources/processes/variable-mapping-extensions.json
```diff
@@ -0,0 +1,47 @@
...
+      },
+      "input-unmapped-variable-name-id": {
+        "id": "input-unmapped-variable-name-id",
+        "name": "input-unmapped-variable-name",
+        "type": "string",
+        "required": false,
+        "value": "inTest"
+      }
+    },
+    "variablesMappings": {
```
magemello: The variablesMappings section of this json it's not anymore valid. @popovicsandras can you add here the new format we agreed together with @salaboy ?

activiti-spring-boot-starter/src/test/resources/processes/connector-var-mapping.bpmn20.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
magemello: the name of this file is misleading...this json is a connector definition it's not a variable mapping file.

magemello: I would suggest to change it for clarity

activiti-spring-boot-starter/src/test/resources/processes/variable-mapping-extensions.json
magemello: This should be the new format of the variablesMappings section:
```
"mappings": {
    "Task_1hkl8bn": {
      "input": {
        "726478cc-91ac-40bb-8f3c-f0782ab81af0": {
          "type": "variable",
          "value": "5e4947c2-d263-4b86-ab98-5627ff5f461a"
        },
        "33ec3b54-90ff-4a97-91be-178976a4dd26": {
          "type": "value",
          // not process variable, comes from the connector configuration, BE magic
          "value": "${host}"
        }
      }
    }
  }
```

erdemedeiros: @ffazzini @magemello @popovicsandras `input` here (singular) is conflict with `inputs` (plural) suggested by @ffazzini for [connector definitions](https://github.com/Activiti/Activiti/pull/2227/files#r251249714). It should be plural as well I guess.

activiti-spring-process-extensions/pom.xml
erdemedeiros: Totally agree @balsarori. I'll find a better place for it.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/VariableDefinition.java
erdemedeiros: @balsarori then I guess we can have a single `VariableDefinition`, no reason to get it in two different places, right?

balsarori: yes we need to added ``value``

activiti-spring-boot-starter/src/test/resources/processes/variable-mapping-extensions.json
popovicsandras: It was plural always. I don't know where this information came from, the connector editor created a definition with the plurals (inputs, outputs) from the very beginning.

erdemedeiros:
Depends on https://github.com/Activiti/activiti-core-common/pull/54

PR #2229, opened 2018-12-03
BroadcastSignal not work with Activiti Core API and Activiti Cloud Rest API

daisuke-yoshimoto:


daisuke-yoshimoto:
@salaboy @ryandawsonuk @erdemedeiros 
I completed fixing. Please review this pr.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
```diff
@@ -297,6 +301,7 @@ public void signal(SignalPayload signalPayload) {
         //@TODO: define security policies for signalling
         runtimeService.signalEventReceived(signalPayload.getName(),
                                            signalPayload.getVariables());
+        eventPublisher.publishEvent(signalPayload);
```
igdianov: @daisuke-yoshimoto The signal() business method is the transactional boundary, which we basically want fully succeed or fully fail not partially succeed. So, it makes sense to define `@Transactional` annotation at the ProcessRuntime service layer and remove the `@Transactional` annotation from sendSignal() method in ProcessInstanceControllerImpl class.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
```diff
@@ -191,6 +196,7 @@ public void signal(SignalPayload signalPayload) {
         //@TODO: define security policies for signalling
         runtimeService.signalEventReceived(signalPayload.getName(),
                                            signalPayload.getVariables());
+        eventPublisher.publishEvent(signalPayload);
```
igdianov: @daisuke-yoshimoto he signal() business method in ProcessRuntimeAdminImpl is the transactional service layer boundary, which we basically want fully succeed or fully fail not partially succeed when it is invoked from any client facade, i.e. Rest controller. So, it makes sense to define @Transactional annotation at the ProcessRuntime service layer and remove the @Transactional from ProcessInstanceAdminController Rest facade.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
daisuke-yoshimoto: @igdianov 
Thanks for your review. I fixed. Please review again.
https://github.com/Activiti/Activiti/pull/2229/commits/9437d7d7474bf71793ff437996025097894b3104
https://github.com/Activiti/activiti-cloud-runtime-bundle-service/pull/171/commits/57da80954874de6926e8eed25031af8c42d1b1f1

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
daisuke-yoshimoto: @igdianov 
Thanks for your review. I fixed. Please review again.
https://github.com/Activiti/Activiti/pull/2229/commits/9437d7d7474bf71793ff437996025097894b3104
https://github.com/Activiti/activiti-cloud-runtime-bundle-service/pull/171/commits/57da80954874de6926e8eed25031af8c42d1b1f1

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
```diff
@@ -187,10 +193,12 @@ public ProcessInstance delete(DeleteProcessPayload deleteProcessPayload) {
     }
 
     @Override
+    @Transactional
     public void signal(SignalPayload signalPayload) {
         //@TODO: define security policies for signalling
         runtimeService.signalEventReceived(signalPayload.getName(),
```
igdianov: @daisuke-yoshimoto  This code sends the same signal two times.  The `signalEventReceived` method should not be invoked at all. Instead, the `eventPublisher` should simply call `publishEvent(signalPayload)` to broadcast the signal via Cloud Stream signalEvent destination

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
```diff
@@ -293,10 +298,12 @@ public void setVariables(SetProcessVariablesPayload setProcessVariablesPayload)
     }
 
     @Override
+    @Transactional
     public void signal(SignalPayload signalPayload) {
         //@TODO: define security policies for signalling
         runtimeService.signalEventReceived(signalPayload.getName(),
```
igdianov: @daisuke-yoshimoto  This code sends the same signal two times.  The `signalEventReceived` method should not be invoked at all. Instead, the `eventPublisher` should simply call `publishEvent(signalPayload)` to broadcast the signal via Cloud Stream signalEvent destination

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
daisuke-yoshimoto: @igdianov 
Thanks for your comment.

I also think that it should be only once, but at the time of the first PR, it is the specification that salaboy decided.
https://github.com/Activiti/activiti-cloud-runtime-bundle-service/pull/38

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
daisuke-yoshimoto: @igdianov
Thanks for your comment.

I also think that it should be only once, but at the time of the first PR, it is the specification that salaboy decided.
Activiti/activiti-cloud-runtime-bundle-service#38

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
igdianov: @daisuke-yoshimoto I do not understand the use case why do we need to send the signal via runtimeService Java Api and broadcast it via Cloud Stream. There will be side effects like a process instance with two signal catch events in your test. May be we need to add a scope attribute to signal broadcast payload to distinguish between cloud and runtime bundle scopes? 

@salaboy Can you clarify the requirements for signal broadcast in this use case?

daisuke-yoshimoto: @igdianov 
I think we needn't send the signal twice.
Simply, if it's a local Runtime Bundle, we can do it via runtimeService Java Api.
If it is another Runtime Bundle, it is necessary to propagate it via Spring Cloud Stream.

The problem is that it will run again on the local Runtime Bundle via the Spring Cloud Stream.

igdianov: @daisuke-yoshimoto We need to be careful with signal re-transmission side effects like you describe if it sent twice. We also have to consider signal definitions scopes (i.e. process, global, cloud) to delegate to appropriate behavior when broadcasting signals via Apis.

daisuke-yoshimoto: @igdianov 
Thanks for your comment. It seems better to reexamine broadcasting signal itself rather than this issue. I will summarize the current behavior.

By the way, I thought about introducing the third scope as well, but with the opinion of salaboy in the global scope it became the specification to propagate to other runtime bundles.

daisuke-yoshimoto: @igdianov 
Well, sorry, it certainly needs to consider process instance scope for Intermediate Signal ThrowEvent. Currently, signal of process instance scope also propagates to other runtime bundle.

https://github.com/Activiti/activiti-cloud-runtime-bundle-service/blob/develop/activiti-cloud-services-runtime-bundle/activiti-cloud-services-subscriptions/src/main/java/org/activiti/services/subscriptions/behavior/BroadcastSignalEventActivityBehavior.java#L31

igdianov: @daisuke-yoshimoto @salaboy This behaviour is not desirable. The Cloud signal event producer should not send process instance scoped signals outside of runtime bundle.  Only globally  scoped signals can be broadcasted via cloud stream. 

Would it make sense to create and use unique destination for each signal name with runtime bundle and scope definition attributes? Signal event  producers can resolve the destination for each signal to send the event message via cloud stream, while signal consumers can dynamically resolve and subscribe to signal sources to only receive signal events  for process definitions signals  inside runtime bundles. 

Does it make any sense?

daisuke-yoshimoto:
@salaboy @igdianov Hi. Sorry I made you wait. Fixing signal 's PR is completed. Please review again.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
salaboy: @igdianov this does make sense.. in general what we were trying to achieve (and what we discussed with @daisuke-yoshimoto ) was:
1) Signal method on the API should work with messages. meaning that you can send SignalPayload to the message driven APIs and it should work
2) We discussed about scopes but we don't have that information right now, so we will need to define global (which before was "Engine" scope to mean multi Runtime Bundle. Local will now mean "inside a Runtime Bundle". finally Process Instance scope ... only to a process instance.  But because we don't have this implemented, we tried to come up with some default behaviour for signals until we have these scopes defined somewhere. 
For now what I propose is that we provide a flag in the SignalPayload to mark the signal as global (or broadcast) to invoke the ```eventPublisher.publishEvent()``` to other runtime bundles. By default it should be local only.

3) The Throw Event BPMN element should also provide the scope, we can do this as an extension, by default it should be local, but if the flag is true it should broadcast outside the runtime bundle. 

@daisuke-yoshimoto @igdianov i think that is a simple change that we can merge now and then improve with a more concrete definition of the scopes.

salaboy: @igdianov regarding "unique destination for each signal name with runtime bundle" do we really need that? meaning that we can decide to which destination each RB is listening to for signals. Conceptually I was thinking about a single destination for all signals, but it is true that we might need to support a more flexible approach. The question is.. do we need that now?

igdianov: @salaboy We can deal with scopes later, I think.  Please, see https://github.com/Activiti/activiti-cloud-runtime-bundle-service/pull/171#issuecomment-449688482

PR #2251, opened 2018-12-10
Fix: dispatch task and process updated events from CommandContext

### LINKED ISSUE #2252: Task and Process Instance updated entity events are not dispatched from Command Context
which creates a problem with transaction synchronization in Runtime Bundle which sends aggregated events batch as a message only when transaction completes as well as applies message headers from runtime execution context to all outgoing messages.

igdianov:
This PR fixes #2252 event dispatching for task and process name update  commands to ensure that events are always generated and dispatched from existing CommandContext inside Activiti transaction, so that they can be aggregated and sent as event messages when Activiti transaction successfully completes.

### DIFF
--- a/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
+++ b/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
@@ -38,8 +38,6 @@ import org.activiti.api.process.model.payloads.SuspendProcessPayload;
 import org.activiti.api.process.model.payloads.UpdateProcessPayload;
 import org.activiti.api.process.runtime.ProcessRuntime;
 import org.activiti.api.process.runtime.conf.ProcessRuntimeConfiguration;
-import org.activiti.api.process.runtime.events.ProcessUpdatedEvent;
-import org.activiti.api.process.runtime.events.listener.ProcessRuntimeEventListener;
 import org.activiti.api.runtime.model.impl.ProcessDefinitionMetaImpl;
 import org.activiti.api.runtime.model.impl.ProcessInstanceImpl;
 import org.activiti.api.runtime.model.impl.ProcessInstanceMetaImpl;
@@ -53,7 +51,6 @@ import org.activiti.engine.ActivitiObjectNotFoundException;
 import org.activiti.engine.RepositoryService;
 import org.activiti.engine.RuntimeService;
 import org.activiti.engine.repository.ProcessDefinitionQuery;
-import org.activiti.runtime.api.event.impl.ProcessUpdatedEventImpl;
 import org.activiti.runtime.api.model.impl.APIProcessDefinitionConverter;
 import org.activiti.runtime.api.model.impl.APIProcessInstanceConverter;
 import org.activiti.runtime.api.model.impl.APIVariableInstanceConverter;
@@ -77,16 +74,14 @@ public class ProcessRuntimeImpl implements ProcessRuntime {
 
     private final ProcessSecurityPoliciesManager securityPoliciesManager;
     
-    private final List<ProcessRuntimeEventListener<ProcessUpdatedEvent>> processUpdatedListeners;
-
     public ProcessRuntimeImpl(RepositoryService repositoryService,
                               APIProcessDefinitionConverter processDefinitionConverter,
                               RuntimeService runtimeService,
                               ProcessSecurityPoliciesManager securityPoliciesManager,
                               APIProcessInstanceConverter processInstanceConverter,
                               APIVariableInstanceConverter variableInstanceConverter,
-                              ProcessRuntimeConfiguration configuration,
-                              List<ProcessRuntimeEventListener<ProcessUpdatedEvent>> processUpdatedListeners) {
+                              ProcessRuntimeConfiguration configuration
+                              ) {
         this.repositoryService = repositoryService;
         this.processDefinitionConverter = processDefinitionConverter;
         this.runtimeService = runtimeService;
@@ -94,7 +89,6 @@ public class ProcessRuntimeImpl implements ProcessRuntime {
         this.processInstanceConverter = processInstanceConverter;
         this.variableInstanceConverter = variableInstanceConverter;
         this.configuration = configuration;
-        this.processUpdatedListeners=processUpdatedListeners;
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/conf/TaskRuntimeAutoConfiguration.java
```diff
@@ -129,6 +128,16 @@ public InitializingBean registerTaskCreatedEventListener(RuntimeService runtimeS
...
                                                      ActivitiEventType.TASK_CREATED);
     }
+    
+    @Bean
+    public InitializingBean registerTaskUpdatedEventListener(RuntimeService runtimeService,
+                                                             @Autowired(required = false) List<TaskRuntimeEventListener<TaskUpdatedEvent>> listeners,
+                                                             ToAPITaskUpdatedEventConverter taskCreatedEventConverter) {
+        return () -> runtimeService.addEventListener(new TaskUpdatedListenerDelegate(getInitializedTaskRuntimeEventListeners(listeners),
+                                                                                     taskCreatedEventConverter),
+                                                     ActivitiEventType.ENTITY_UPDATED);
```
erdemedeiros: @igdianov @salaboy if I remember well we moved away from this solution (listening to internal engine events), because for a task, when it has an assignee in the definition, the event 'ENTITY_UPDATED' happens before the event `TASK_CREATED` event.
We have: `ENTITY_CREATED`, `ENTITY_UPDATED`, `TASK_CREATED`). This will make the current implementation of query-service fail. If we want keep listening to internal events we need either:
- use to `ENTITY_CREATED` (instead of TASK_CREATED) to trigger `TaskCreatedEvent`. In this case we need to check that `ENTITY_CREATED` event provides exactly the same info as `TASK_CREATED`. We may need to re-trigger an event on `TASK_CREATED` to update info if necessary.

**Or**

- handle taskUpdatedEvent as a taskCreatedEvent in the case where no task is found in query-service.

**Or**

- any other solution?

The first option (`ENTITY_CREATED`) seems be the better approach IMO.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
```diff
@@ -337,14 +331,7 @@ public ProcessInstance update(UpdateProcessPayload updateProcessPayload) {
                                                                              .processInstanceId(updateProcessPayload.getProcessInstanceId())
                                                                              .singleResult());
         
-        fireProcessUpdatedEvent(updatedProcessInstance);
         return updatedProcessInstance;
  
     }
-    
-    private void fireProcessUpdatedEvent(ProcessInstance updatedProcessInstance) {
```
erdemedeiros: @igdianov I don't see a `registerProcessUpdatedEventListener` replacing this.

erdemedeiros: Ignore this comment, the registration is already taking place in develop https://github.com/Activiti/Activiti/blob/develop/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/conf/activiti/runtime/api/ProcessRuntimeAutoConfiguration.java#L185

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/conf/TaskRuntimeAutoConfiguration.java
igdianov: I fixed it in TaskEntityManagerImpl.java, that was ignoring fireEvents flag

activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManagerImpl.java
```diff
@@ -122,7 +122,7 @@ private void changeTaskAssignee(TaskEntity taskEntity, String assignee, boolean
       if (taskEntity.getId() != null) {
         getHistoryManager().recordTaskAssigneeChange(taskEntity.getId(), taskEntity.getAssignee());
         addAssigneeIdentityLinks(taskEntity);
-        update(taskEntity);
+        update(taskEntity, fireEvents);
```
igdianov: The update method was invoked without fireEvents flag, which was causing the task update entity event fired before task created in the engine.

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/conf/TaskRuntimeAutoConfiguration.java
erdemedeiros: Awesome @igdianov I've missed this small change in `TaskEntityManagerImpl` that fixed the ordering issue!

activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SetProcessInstanceNameCmd.java
```diff
@@ -56,6 +58,10 @@ public Void execute(CommandContext commandContext) {
 
     // Actually set the name
     execution.setName(name);
+    
+    if (commandContext.getEventDispatcher().isEnabled()) {
```
igdianov: Here, the dispatchEvent call will fire entity updated event for the process instance execution entity inside execution context

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -261,16 +254,9 @@ public Task update(UpdateTaskPayload updateTaskPayload) {
...
-        Task convertedTask = task(updateTaskPayload.getTaskId());
-        fireTaskUpdatedEvent(convertedTask);
-        return convertedTask;
-    }
-
-    private void fireTaskUpdatedEvent(Task convertedTask) {
-        for (TaskRuntimeEventListener<TaskUpdatedEvent> listener : taskUpdatedListeners) {
-            listener.onEvent(new TaskUpdatedEventImpl(convertedTask));
-        }
+        org.activiti.engine.task.Task updatedTask = taskService.saveTask(internalTask);
```
igdianov: Here, I simplified the code and modified taskService.saveTask to return updated task entity without need to requery task to return the result.

Issue #2113, opened 2018-11-02
Create new method to expose subprocess IDs to a given process instance


salaboy:
We should expose in the Java API and Runtime Bundle and Query Service. 
The endpoint should look like:
- /v1/process-instances/{processInstanceId}/subprocesses

PR #2260, opened 2018-12-12
Implementation of a method subprocesses

CTI777:
Depends on: https://github.com/Activiti/activiti-api/pull/62

### DIFF
--- a/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
+++ b/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
@@ -24,6 +24,7 @@ import org.activiti.api.process.model.builders.ProcessPayloadBuilder;
 import org.activiti.api.process.model.payloads.DeleteProcessPayload;
 import org.activiti.api.process.model.payloads.GetProcessDefinitionsPayload;
 import org.activiti.api.process.model.payloads.GetProcessInstancesPayload;
+import org.activiti.api.process.model.payloads.GetSubprocessesPayload;
 import org.activiti.api.process.model.payloads.RemoveProcessVariablesPayload;
 import org.activiti.api.process.model.payloads.ResumeProcessPayload;
 import org.activiti.api.process.model.payloads.SetProcessVariablesPayload;
@@ -228,5 +229,22 @@ public class ProcessAdminRuntimeImpl implements ProcessAdminRuntime {
     public void removeVariables(RemoveProcessVariablesPayload removeProcessVariablesPayload) {
         runtimeService.removeVariables(removeProcessVariablesPayload.getProcessInstanceId(),
                                            removeProcessVariablesPayload.getVariableNames());
+    }
+    
+    @Override
+    public Page<ProcessInstance> subprocesses(Pageable pageable,
+                                              GetSubprocessesPayload getSubprocessesPayload) {
+        
+        //Process Instance will check security policies on read
+        ProcessInstance processInstance = processInstance(getSubprocessesPayload.getProcessInstanceId());
+        
+        org.activiti.engine.runtime.ProcessInstanceQuery internalQuery = runtimeService.createProcessInstanceQuery();
+        
+        internalQuery.superProcessInstanceId(processInstance.getId());
+        
+        return new PageImpl<>(processInstanceConverter.from(internalQuery.listPage(pageable.getStartIndex(),
+                                                                                   pageable.getMaxItems())),
+                              Math.toIntExact(internalQuery.count()));
+        
     }
 }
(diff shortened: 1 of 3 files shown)


activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/model/impl/APIProcessInstanceConverter.java
```diff
@@ -34,6 +42,33 @@ public ProcessInstance from(org.activiti.engine.runtime.ProcessInstance internal
         processInstance.setProcessDefinitionKey(internalProcessInstance.getProcessDefinitionKey());
         processInstance.setBusinessKey(internalProcessInstance.getBusinessKey());
         processInstance.setStatus(calculateStatus(internalProcessInstance));
+        processInstance.setProcessDefinitionVersion(internalProcessInstance.getProcessDefinitionVersion());
+        
+        //To do: it is not the best way to search parentProcessId by this method, it will require extra queries!
+
```
igdianov: It will be a general problem for lazy loaded associations when using converters outside of Activiti engine transaction. This will also create N+1 additional queries to database if we need to resolve child entities for each record. We should be able to solve this by implementing or using a query to join associated entities in one query from the database.

erdemedeiros:
Refs https://github.com/Activiti/Activiti/issues/2113

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessAdminRuntimeImpl.java
```diff
@@ -229,4 +230,21 @@ public void removeVariables(RemoveProcessVariablesPayload removeProcessVariables
         runtimeService.removeVariables(removeProcessVariablesPayload.getProcessInstanceId(),
                                            removeProcessVariablesPayload.getVariableNames());
     }
+    
+    @Override
+    public Page<ProcessInstance> subprocesses(GetSubprocessesPayload getSubprocessesPayload,
+                                              Pageable pageable) {
+        
+        //Process Instance will check security policies on read
+        ProcessInstance processInstance = processInstance(getSubprocessesPayload.getProcessInstanceId());
```
erdemedeiros: Based on my comments on https://github.com/Activiti/activiti-api/pull/62 this extra method is not necessary you can update `processInstances` method instead.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
```diff
@@ -334,4 +335,21 @@ public ProcessInstance update(UpdateProcessPayload updateProcessPayload) {
         return updatedProcessInstance;
  
     }
+    
+    @Override
+    public Page<ProcessInstance> subprocesses(GetSubprocessesPayload getSubprocessesPayload,
```
erdemedeiros: Same here, you can update `processInstances` method instead.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/model/impl/APIProcessInstanceConverter.java
erdemedeiros: Another question: do we really need the `parentProcessInstanceId` or `rootProcessInstanceId` and `callerId`(not sure how to name this one) is enough? If we really need it, maybe we should consider to add an extra property to the internal process instance and store it in the database: similar to what is done for `rootProcessInstanceId`. @igdianov what do you think about that? @salaboy you may be interested in this conversation as well.

CTI777: rootProcessInstanceId - is present but not clear if it is supported (filled) at all.
parentId (parentProcessInstanceId) we will need for sure to get list if subprocesses...
What is callerId and do we need it is a question...

igdianov: @erdemedeiros It makes sense to provide parentProcessInstanceId for query, audit services. This use case demonstrates several problems. Problem 1 is that internal api converters need to run inside Activiti transaction in order to be able to resolve lazily associated entities, i.e. `superExecution`, to get its parent process instance id. Problem 2 is that resolving lazy associations this way for each record will create N+1 additional queries to the engine database from many places (aggregating events, apis, queries) that we cannot control because of Problem 1. 

The parent process instance seems to be a common property shared and used by entity and event classes, so this particular use case could be solved by making changes to the execution entity manager query in Activiti engine to always eagerly load super execution entity via join query if it exists. This should have less performance impact overall.

activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/model/impl/APIProcessInstanceConverter.java
```diff
@@ -47,27 +42,11 @@ public ProcessInstance from(org.activiti.engine.runtime.ProcessInstance internal
...
-                            Optional.ofNullable(executionEntity.getSuperExecution())
-                                    .ifPresent(superExecution -> processInstance.setParentId(superExecution.getProcessInstanceId()));
-
-                             return null;
-                        }});
-
-            } else {
-                  Optional.ofNullable(executionEntity.getSuperExecution())
-                          .ifPresent(superExecution -> processInstance.setParentId(superExecution.getProcessInstanceId()));
-            }
```
igdianov: @erdemedeiros @CTI777 I have removed the hack in APIProcessInstanceConverter.java above in favor of using Mybatis mapping to eagerly fetch superExecution entity via association property. It seems like we always need to resolve parent process instance attributes in event converters and queries, so I think it makes sense to always fetch it.

See Mabits changes below: https://github.com/Activiti/Activiti/pull/2260/commits/f3543483fbb600e1083f42b138d2c000c226ef2e#diff-b9f50b6ee46df36fb61073916d78de78 for details.

PR #2277, opened 2018-12-14
feat: VARIABLE_DELETED events will contain old value 

CTI777:


salaboy:
@CTI777 there seems to be an issue with other tests.. look at it in travis

Issue #2372, opened 2019-01-17
Add admin update task endpoint


erdemedeiros:
Add endpoint `PUT admin/v1/tasks/${taskId}`.

PR #2397, opened 2019-01-22
Update task is implemented for TaskAdminRuntimeImpl

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2372
Depends on https://github.com/Activiti/activiti-api/pull/75

### DIFF
--- a/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
+++ b/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
@@ -234,8 +234,4 @@ public class TaskAdminRuntimeImpl implements TaskAdminRuntime {
         return taskService.getIdentityLinksForTask(taskId);
     }
 
-    @Override
-    public Task update(UpdateTaskPayload updateTaskPayload) {
-        throw new IllegalArgumentException("To be implemented");
-    }
 }
--- a/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
+++ b/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
@@ -18,6 +18,7 @@ package org.activiti.runtime.api.impl;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import org.activiti.api.runtime.shared.NotFoundException;
 import org.activiti.api.runtime.shared.query.Page;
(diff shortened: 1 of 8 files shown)


activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
```diff
@@ -88,6 +90,70 @@ public Task task(String taskId) {
         return new PageImpl<>(tasks,
                               Math.toIntExact(taskQuery.count()));
     }
+    
+    @Override
+    public Task update(UpdateTaskPayload updateTaskPayload) {
+        org.activiti.engine.task.Task internalTask = taskService.createTaskQuery().taskId(updateTaskPayload.getTaskId()).singleResult();
```
erdemedeiros: Lets avoid duplication between `TaskRuntimeImpl` and here. The only difference seems to be that in `TaskRuntimeImpl` we need to check the assignee. You could extract this code inside a new class (`TaskUpdater`) and only check the assignee if the parameter `checkAssignee` is `true`.

CTI777: I have made refactoring. Mock tests for task update are currently **commented**, they should be changed after changes approved...

CTI777: @erdemedeiros 
Can you please check these tests (mock tests, commented) if we need them at all?

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
```diff
@@ -88,6 +89,12 @@ public Task task(String taskId) {
         return new PageImpl<>(tasks,
                               Math.toIntExact(taskQuery.count()));
     }
+    
+    @Override
+    public Task update(UpdateTaskPayload updateTaskPayload) {
+        TaskUpdater taskUpdater= new TaskUpdater(taskService,taskConverter,null,null);
```
erdemedeiros: @CTI777 let's instantiate `TaskUpdater` as a bean in `TaskRuntimeAutoConfiguration` and inject it via constructor here and in `TaskRuntimeImpl`.

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -210,80 +210,10 @@ public Task release(ReleaseTaskPayload releaseTaskPayload) {
...
-                internalTask.setFormKey(newValue);
-            }
-        }
-        
-        if (updates > 0) {
-            taskService.saveTask(internalTask);
-        }
-        
-        return taskConverter.from(getInternalTask(updateTaskPayload.getTaskId()));
+        TaskUpdater taskUpdater= new TaskUpdater(taskService,taskConverter,securityManager,userGroupManager);
```
erdemedeiros: Let's inject `TaskUpdater` via constructor.

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskUpdater.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        
+        if (updates > 0) {
+            taskService.saveTask(internalTask);
+        }
+        
+        return taskConverter.from(taskService.createTaskQuery().taskId(updateTaskPayload.getTaskId()).singleResult());
+    }
+    
+    public org.activiti.engine.task.Task getInternalTaskWithChecks(String taskId) {
```
erdemedeiros: Ok, we need to find a better name than `TaskUpdater` for this class, as we need remove the method `getInternalTaskWithChecks` from `TaskRuntimeImpl` and reuse this one (to avoid duplication). Maybe `TaskRuntimeHelper` while we wait for a better name (or anything else you think it makes sense).

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskUpdater.java
```diff
@@ -0,0 +1,122 @@
...
+                updates++;
+                internalTask.setFormKey(newValue);
+            }
+        }
+        
+        if (updates > 0) {
+            taskService.saveTask(internalTask);
+        }
+        
+        return taskConverter.from(taskService.createTaskQuery().taskId(updateTaskPayload.getTaskId()).singleResult());
```
erdemedeiros: Here I think you can replace 
`return taskConverter.from(taskService.createTaskQuery().taskId(updateTaskPayload.getTaskId()).singleResult());`
by
`return taskConverter.from(internalTask);`

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskUpdater.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        
+        if (updateTaskPayload.getDueDate() != null) {
+            if (!Objects.equals(internalTask.getDueDate(),updateTaskPayload.getDueDate())) {
+                updates++;
+                internalTask.setDueDate(updateTaskPayload.getDueDate());
+            }
+        }
+        
+        //@TODO: check if this value can be updated
```
erdemedeiros: You can remove this `//@TODO`

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskUpdater.java
CTI777: Internal task will not be good, because we save a record to DB, and then we want to read it to get latest state from the DB...

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskUpdater.java
CTI777: @erdemedeiros
Can you please check if need here mock tests? Currently I have commented them, because they are failing...
    //@Test
    public void updateShouldThrowExceptionWhenAssigneeIsNotSet() {

 //@Test
    public void updateShouldBeAbleToUpdateDescriptionOnly() {

activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskAdminRuntimeImpl.java
erdemedeiros: @CTI777 I'll help you with the test. Please don't comment them out as this might be merged by mistake, it's better to have them failing so we know we need to do something about them.

Issue #2272, opened 2018-12-14
Missing event api's for signals 


igdianov:
It seems there are missing events in activiti-api that should be produced when signal is emitted  received by the subscribed consumer.

salaboy:
You mean, the message based approach for this: https://github.com/Activiti/activiti-api/blob/develop/activiti-api-process-runtime/src/main/java/org/activiti/api/process/runtime/ProcessRuntime.java#L115 ? or the events that are generated by that operation?

igdianov:
@salaboy I mean the event are not generated for signal operations...

CTI777:
@igdianov @salaboy 
Can somebody advise me what is not working exactly, maybe an existing test to start with?
I have seen that signals are handled, in general, differently compared to other process, tasks, variables events. In audit service I see no traces of signal at all...

salaboy:
@CTI777 the problem is that in the Audit Service we don't have a handler for Signal Events, so when the Runtime Bundle creates a signal event, we just ignore this in Audit. We just need to create a new SignalEvent handler in Audit to store the event as is in the database.

CTI777:
@salaboy So the problem is only audit! And there are no problems in RB, etc....

salaboy:
only on Audit

CTI777:
@salaboy 
I have analyzed the issue again and am sure that this issue requires many changes for the Signal event: from activiti-api, cloud-activiti-api, activiti, converters... We do not have this event, do not have any delegates or converters for this event...
ActivitiEventType has ACTIVITY_SIGNALED, ACTIVITY_MESSAGE_RECEIVED, ACTIVITY_ERROR_RECEIVED which have relation to signal...
Maybe we can discuss this once in details...

salaboy:
@CTI777 let's do a quick zoom today.. ping me when you are around

salaboy:
You are right.. we don't even have the listener here: https://github.com/Activiti/Activiti/tree/develop/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/event/internal

That should be our first step.

CTI777:
@salaboy 
I think we need to define SignalEvent in activiti-api.
Should it be defined in model shared part, like below: 
https://github.com/Activiti/activiti-api/tree/develop/activiti-api-model-shared/src/main/java/org/activiti/api/model/shared/event
[...]

PR #2511, opened 2019-02-14
Implementation of ACTIVITY_SIGNALED event

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2272
Depends on https://github.com/Activiti/activiti-api/pull/84


activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/model/impl/ToSignalConverter.java
```diff
@@ -0,0 +1,39 @@
...
+    public BPMNSignal from(ActivitiSignalEvent internalEvent) {
+        BPMNSignalImpl signal = new BPMNSignalImpl(internalEvent.getActivityId(),
+                								   internalEvent.getActivityName(),
+                								   internalEvent.getActivityType());
+        
+        
+
+        signal.setProcessDefinitionId(internalEvent.getProcessDefinitionId());
+        signal.setProcessInstanceId(internalEvent.getProcessInstanceId());
+        signal.setSignalPayload((SignalPayload)internalEvent.getSignalData());
```
erdemedeiros: This is not going to work as `internalEvent.getSignalData()` is never a `SignalPayload`. It's actually going to be variables. See https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SignalEventReceivedCmd.java#L48 and https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SignalEventReceivedCmd.java#L97-L99

CTI777: Have made changes and added a test. It seems that internal ActivitiSignalEvent does not have processInstanceId, but has a processDefinitionId... I don't know if this is a big problem....

erdemedeiros: Thanks @CTI777. Ok, so `processInstanceId` is not set because you picked up a process with start signal event. For this specific test I think an intermediate catch signal event or a boundary signal event fits better. You can copy `ProcessWithBoundarySignal.bpmn20.xml` from runtime bundle starter, then `processInstanceId` will be set. Also, I'd recommend starting two instances of the process and verify in the test that two `SIGNAL_RECEIVED` are catch.

CTI777: @erdemedeiros 
I have added some more tests here and made some minor refactoring...

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeEvents.java
```diff
@@ -112,4 +117,105 @@ public void shouldGetJustThreeVariableCreatedEvent(){
...
+    
+    @Test
+    public void shouldGetSignalReceivedEventsForProcessWithSignalStart(){
+
+    	//In this test processInstanceId is null
+        //given
+        securityUtil.logInAs("salaboy");
+
+        //when
+        ProcessInstance categorizeProcess = processRuntime.start(ProcessPayloadBuilder.start()
```
erdemedeiros: In this case you should not start an instance of the process, as the instance will be created when the signal is triggered. Your test should still work if you remove `processRuntime.start`.

erdemedeiros:
@CTI777 it looks good to me. I'm not merging yet because I think @salaboy want to take a look on it as well.

PR #2604, opened 2019-03-05
Call activity tests

miguelruizdev:
Reprise of this old PR:
https://github.com/Activiti/Activiti/pull/2090


activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeCallActivityIT.java
```diff
@@ -0,0 +1,77 @@
...
+                        .build());
+
+        assertThat(processInstance).isNotNull();
+
+        //verify the existence of the sub process itself
+        ProcessInstance subProcessInstance = processRuntime.processInstances(
+                Pageable.of(0, 50),
+                ProcessPayloadBuilder
+                        .processInstances()
+                        .withProcessDefinitionKey(SUB_PROCESS_CALL_ACTIVITY)
```
erdemedeiros: I think you can use `.withParentProcessInstanceId(processInstance.getId())` instead of `.withProcessDefinitionKey(SUB_PROCESS_CALL_ACTIVITY)` to avoid failures in the case another test also starts an instance of this process definition.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeCallActivityIT.java
```diff
@@ -0,0 +1,77 @@
...
+                        .build())
+                .getContent()
```
erdemedeiros: It's better to assert that content is not empty before doing `get(0)`.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeCallActivityIT.java
```diff
@@ -0,0 +1,77 @@
...
+        assertThat(subProcessInstance.getProcessDefinitionKey()).isEqualTo(SUB_PROCESS_CALL_ACTIVITY);
+
+        Task task = taskRuntime.tasks(
+                Pageable.of(0, 50),
+                TaskPayloadBuilder
+                        .tasks()
+                        .withProcessInstanceId(subProcessInstance.getId())
+                        .build())
+                .getContent()
+                .get(0);
```
erdemedeiros: Same here, it's better to assert hat content is not empty before doing `get(0)`.

Issue #2664, opened 2019-03-27
Signal chain does not work


ggrnd0:
Must what works?
1) start SIGNALS_CHAIN process
2) trigger trigger message
3) process done

Actually process waits for ev2

![image](https://user-images.githubusercontent.com/5565134/55082287-7773f680-50b2-11e9-9bed-5670409d863e.png)

[ActivitiSignalsChainFeatureTest.testSignalsChain.bpmn.zip](https://github.com/Activiti/Activiti/files/3013520/ActivitiSignalsChainFeatureTest.testSignalsChain.bpmn.zip)

Reason, agenda&commandInvokers walks throught process bypassing wide.
And when ev2 fired, ev2 catchEvent not exists yet.

There is some strange with flow.

In one process that pattern works fine.
In second does not works.
In that test also does not works with exclusiveGateway added.

ggrnd0:
sorry, that is our mistake..

ggrnd0:
Oh, no. not only our XD

I updated image/scenario/description

ggrnd0:
overriding in Agenda does not work

```                    @Override
                    public Runnable getNextOperation() {
                        return operations.pollLast();
                    }
```

ggrnd0:
With overriding in Agenda ```public Runnable getNextOperation() {
                        return operations.pollLast();
                    }``` and in IntermediateThrowSignalEventActivityBehavior.execute ```
    // that line was last
    Context.getAgenda().planTakeOutgoingSequenceFlowsOperation((ExecutionEntity) execution, true);

    for (SignalEventSubscriptionEntity signalEventSubscriptionEntity : subscriptionEntities) {
      Context.getProcessEngineConfiguration().getEventDispatcher().dispatchEvent(
          ActivitiEventBuilder.createSignalEvent(ActivitiEventType.ACTIVITY_SIGNALED, signalEventSubscriptionEntity.getActivityId(), eventSubscriptionName,
              null, signalEventSubscriptionEntity.getExecutionId(), signalEventSubscriptionEntity.getProcessInstanceId(),
              signalEventSubscriptionEntity.getProcessDefinitionId()));

      eventSubscriptionEntityManager.eventReceived(signalEventSubscriptionEntity, null, signalEventDefinition.isAsync());
    }```

process ended correctly.

that 2 changes forces bypassing deep and place event rising at end.

But HOW that will affect other processes??? >_<

that little process repaired but original process still stucks...

ggrnd0:
[...]

Issue #2695, opened 2019-04-14
Event based GW with single catchEvent


ggrnd0:
That method do not check is eventActivityIds empty or not:
https://github.com/Activiti/Activiti/blob/c8666252c669d89f2c6d7063e4a89aadf6b73175/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/IntermediateCatchEventActivityBehavior.java#L86

And query to db fails
https://github.com/Activiti/Activiti/blob/35be47101f5450df20d9c7c3d74524c31788dbc6/activiti-engine/src/main/resources/org/activiti/db/mapping/entity/Execution.xml#L444-L453

How that must works? is it bug?

PR #2696, opened 2019-04-14
fix for #2664

ggrnd0:
https://github.com/Activiti/Activiti/issues/2664 fix with commandContext.agendaStack


ggrnd0:
Also include fix for https://github.com/Activiti/Activiti/issues/2695

Issue #2694, opened 2019-04-12
Static values in the service task mapping


magemello:
As a developer I want to be able to define in my process a task scope value, without have to define it in the connector definition.

```
    "mappings": {
      "Task_0h46wzd": {
   
//this is in the connector definition     
        "inputs": {
          "8ee05009-0e1d-4a7a-8300-455fba7543c9": {
            "type": "variable",
            "value": "e74251d3-e2fb-47d0-9637-753c9a6cc7b4"
          },
          "df3533aa-87be-49e6-b7d0-5875117ab1aa": {
            "type": "value",
            "value": "test"
          },

//this is not in the connector definition
          "my-custom-value-name": {
            "type": "static_value",
            "value": "test"
          }
        },

        "outputs": {
          "36f4dc97-c52f-452b-8ff7-3e79d605f13d": {
            "type": "variable",
            "value": "e74251d3-e2fb-47d0-9637-753c9a6cc7b4"
          },
          "b61d97a7-152c-4da9-8f5d-fadae37a03a1": {
            "type": "variable",
            "value": "07bf6d25-dbb2-47ad-8f1d-a10d753aa425"
          }
        }

      }
    },
```

magemello:
As part of this we need also to change the extension validator to accept static-value as type

magemello:
https://github.com/Activiti/activiti-cloud-org-service/pull/216

erdemedeiros:
Done.

PR #2699, opened 2019-04-15
#2694 static values inputs for connectors

balsarori:
#2694

### DIFF
--- a/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/RuntimeTestConfiguration.java
+++ b/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/RuntimeTestConfiguration.java
@@ -202,6 +202,7 @@ public class RuntimeTestConfiguration {
             String variableOne = "input-variable-name-1";
             String variableTwo = "input-variable-name-2";
             String variableThree = "input-variable-name-3";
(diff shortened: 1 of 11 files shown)


erdemedeiros:
It looks good to me. @salaboy any concern about it?

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/RuntimeTestConfiguration.java
```diff
@@ -221,7 +222,9 @@ public Connector variableMappingActionName() {
                             tuple(variableThree,
                                   5),
                             tuple(unmappedMatchingVariable,
-                                  "inTest"));
+                                  "inTest"),
```
erdemedeiros: @balsarori note that this connector has a connector definition file. In addition to this test, I think it would be judicious to add another one where the connector has no connector definition (as this is the case you want to cover).

erdemedeiros: @balsarori I've added the test.

balsarori: Thanks

Issue #2736, opened 2019-05-09
[Core] Process definition version not set for integration request


erdemedeiros:
Process definition version is not set (in the header and inside IntegrationContext object) when the connector is not the first step of the process.

How to reproduce:
1. Create a process: UserTask -> ServiceTask -> UserTask
2. Start the process and execute the user task
3. Inside the connector check the header `processDefinitionVersion` (it's missing) and integrationRequest.getIntegrationContext().getProcessDefinitionVersion (It's null).

igdianov:
![image](https://user-images.githubusercontent.com/20428629/57495400-c8b30080-7282-11e9-8e01-c22cf82ed30c.png)

igdianov:
We also need to add acceptance scenarios for this use case.

salaboy:
@igdianov is there any acc tests addition to test the PRs in the examples?

igdianov:
@salaboy I have added https://github.com/Activiti/activiti-cloud-acceptance-scenarios/pull/367

salaboy:
done and all merged with acc tests

PR #2737, opened 2019-05-10
fix: set process definition version and key for integration request

igdianov:
Part of https://github.com/Activiti/Activiti/issues/2736

### DIFF
--- a/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/connector/IntegrationContextBuilder.java
+++ b/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/connector/IntegrationContextBuilder.java
@@ -68,7 +68,6 @@ public class IntegrationContextBuilder {
 
             if(processDefinition != null) {
                 integrationContext.setProcessDefinitionKey(processDefinition.getKey());
-                integrationContext.setProcessDefinitionKey(processDefinition.getKey());
                 integrationContext.setProcessDefinitionVersion(processDefinition.getVersion());
             }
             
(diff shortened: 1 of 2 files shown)


activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/connector/IntegrationContextBuilder.java
```diff
@@ -53,13 +55,22 @@ private IntegrationContextImpl buildFromExecution(DelegateExecution execution,
         integrationContext.setClientId(execution.getCurrentActivityId());
 
         if(ExecutionEntity.class.isInstance(execution)) {
-            ExecutionEntity processInstance = ExecutionEntity.class.cast(execution)
-                                                                   .getProcessInstance();
+            ExecutionContext executionContext = new ExecutionContext(ExecutionEntity.class.cast(execution));
```
erdemedeiros: @igdianov I was wondering if we shouldn't inject `RepositoryService` here and retrieve the process definition from it. `RepositoryService` implementation is already providing cache via [DeploymentManager](https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/deploy/DeploymentManager.java#L68) and it will be easier to mock than `new ExecutionContext()`. What do you think?

igdianov: @erdemedeiros The ExecutionContext uses ProcessEngineConfiguration to lookup Deployment via ProcessDefinition caching optimization as well as using DeploymentManager entity caching. It think it is better to use existing infrastructure. The mocking is not hard at all.

Issue #2766, opened 2019-05-24
Add support for taskDefinitionKey in the Task model


CTI777:
In our Task model we do not have taskDefinitionKey which represents task Id from BPMN model:
https://github.com/Activiti/activiti-api/blob/develop/activiti-api-task-model/src/main/java/org/activiti/api/task/model/Task.java
Internal TaskInfo from Activiti does have this field:
https://github.com/Activiti/Activiti/blob/develop/activiti-engine/src/main/java/org/activiti/engine/task/TaskInfo.java#L75

PR #2797, opened 2019-06-13
AAE-771 Support taskDefinitionKey in the Task model 

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2766
Depends on https://github.com/Activiti/activiti-api/pull/126


eromano:
should we add a unit test?

PR #2834, opened 2019-07-23
Use named bean for behaviorFactoryConfigurer

almerico:
- Allow override this bean in runtime bundle


almerico:
ACTIVITI-3415

erdemedeiros:
issues.alfresco.com/jira/browse/ACTIVITI-3310

PR #2837, opened 2019-07-24
extension remove from call activity

almerico:
remove support for  extensionElements like
  <callActivity id="callactivity1" name="Call activity" calledElement="process2">
        <extensionElements>
          <activiti:in source="sendFor" target="sendFor" />
        </extensionElements>
    </callActivity>


Klapsa2503:
Hi, could you please tell me what was the driver of this change?

Aunixtoyou:
why?

PR #2843, opened 2019-07-30
Update ExecutionEntityImpl.java

ckheman:
Condition checked against the wrong variable.


erdemedeiros:
Fixed by https://github.com/Activiti/Activiti/pull/3303.

PR #2848, opened 2019-08-06
Activiti[3288] - Fix mapping for null variables

erdemedeiros:
In some cases a mapping can reference variables that are not set. In this case the entry of the mapping should be ignored instead of throwing NPE


igdianov:
@erdemedeiros  So, the strategy is to omit the non-required variables, so that mapping behavior is automatically implied for non-required variables, which does not propagate missing variables downstream. 

I do not see a problem with this, but I think in the next milestone, it would be useful to add `required` mapping attribute for the behavior to validate input/output mappings explicitly to avoid runtime NPEs.

erdemedeiros:
@igdianov I'm not sure I completely understood what you mean. Even variables that are required could be set to null at some point during the process execution.

Issue #2886, opened 2019-09-04
We need to move the DateTimeProvider to a common place in activiti


CTI777:
Related to https://github.com/Activiti/Activiti/issues/2869

PR #2888, opened 2019-09-04
Move DateFormatterProvider to activiti core

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2886
DateFormatterProvider is moved to activiti core
Default process variables having date type are supported

### DIFF
--- a/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/DateFormatterProvider.java
+++ b/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/DateFormatterProvider.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.activiti.spring.process.variable;
+
+import java.text.MessageFormat;
+import java.time.DateTimeException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.Date;
+
+import org.springframework.beans.factory.annotation.Value;
+
+public class DateFormatterProvider  {
+   
+    @Value("${spring.activiti.date-format-pattern.date-format-pattern:yyyy-MM-dd[['T'][ ]HH:mm:ss[.SSS'Z']]}")
+    private String dateFormatPattern;
+    
+    private ZoneId zoneId = ZoneOffset.UTC;
+ 
+    public DateFormatterProvider() {
+    }
+    
+    public ZoneId getZoneId() {
+        return zoneId;
+    }
+    
+    public String getDateFormatPattern() {
+        return dateFormatPattern;
+    }
+    
+    public void setDateFormatPattern(String dateFormatPattern) {
+        this.dateFormatPattern = dateFormatPattern;
+    }
+ 
+    public Date convert2Date(String value) throws DateTimeException { 
+        DateTimeFormatter dateTimeFormatter = new DateTimeFormatterBuilder()
+                                                      .appendPattern(getDateFormatPattern())
+                                                      .toFormatter()
+                                                      .withZone(getZoneId());
+        
+        try {
+            LocalDateTime localDateTime = dateTimeFormatter.parse(value,
+                                                                  LocalDateTime::from);
+            return Date.from(localDateTime.atZone(getZoneId()).toInstant());                
+        } catch (DateTimeException e) {
+            LocalDate localDate = dateTimeFormatter.parse(String.valueOf(value),
+                                                          LocalDate::from);
+            return Date.from(localDate.atStartOfDay().atZone(getZoneId()).toInstant());
+        }
+    }    
+    
+    public Date convert2Date(Object value) throws DateTimeException {
+        if (value instanceof String) {
+            return convert2Date(String.valueOf(value));
+        }
+
+        if (value instanceof Date) {
+            return (Date)value;
+        }
+        
+        if (value instanceof Long) {
+            return new Date((long)value);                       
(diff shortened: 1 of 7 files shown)


activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/variable/DateFormatterProviderTest.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+public class DateFormatterProviderTest {
+
+    private DateFormatterProvider provider = new DateFormatterProvider("yyyy-MM-dd[['T'][ ]HH:mm:ss[.SSS'Z']]");
+
+    @Test
+    public void should_returnDate_when_stringRepresentsADate() {
+        //given
```
ffazzini: can we remove the comments?

activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/variable/DateFormatterProviderTest.java
```diff
@@ -0,0 +1,102 @@
...
+        String dateStr = "1970-01-01";
+
+        //when
+        Date date = provider.toDate(dateStr);
+
+        //then
+        assertThat(date).hasTime(0);
```
ffazzini: should not we assert two things? 
1) the date has a date part: 1970-01-01
2) assertThat(date).hasTime(0);

should we add 1) ?

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/DateFormatterProvider.java
```diff
@@ -0,0 +1,78 @@
...
+        DateTimeFormatter dateTimeFormatter = new DateTimeFormatterBuilder()
+                                                      .appendPattern(getDateFormatPattern())
+                                                      .toFormatter()
+                                                      .withZone(getZoneId());
+        
+        try {
+            LocalDateTime localDateTime = dateTimeFormatter.parse(value,
+                                                                  LocalDateTime::from);
+            return Date.from(localDateTime.atZone(getZoneId()).toInstant());                
+        } catch (DateTimeException e) {
```
ffazzini: will catch input values like "1970-01-01" ? 
I was just wondering if it is necessary.. i was expecting localdatetime to return "1970-01-01T00:00:00.000Z" for an input like "1970-01-01"

activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/variable/DateFormatterProviderTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    public void should_returnDate_when_longIsProvided() {
+        //given
+        long time = 1000;
+
+        //when
+        Date date = provider.toDate(time);
+
+        //then
+        assertThat(date).hasTime(time);
```
ffazzini: should we make it clearer what we test in the "long" scenario? 
That we produce a valid date? That we produce a specific date value?

activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/variable/DateFormatterProviderTest.java
CTI777: Comments given/when/then removed

activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/variable/DateFormatterProviderTest.java
CTI777: Have added check for date parts

ffazzini: nice!

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/DateFormatterProvider.java
CTI777: It is possible that input string will come in the date format (without time, for example from extension file): "1970-01-01"... And we have to convert this kind of strings to Date...

ffazzini: "1970-01-01" gets converted to "1970-01-01T00:00:00.000Z" so i think they should be the same thing, because if i am not wrong they both should have => hasTime(0)

We can leave it if you want, or we can add a test to verify that having only LocalDateTime does not break things. Up to you.

erdemedeiros: Good point, @ffazzini. The test is covering this case, so we can remove the catch block and see if the test is still passing. If that's the case we can remove the catch block.

CTI777: >"1970-01-01" gets converted to "1970-01-01T00:00:00.000Z" so i think they should be the same thing, because if i am not wrong they both should have => hasTime(0)

Yes, true, they will both have hasTime(0)...

But, parsing of this string "1970-01-01" using a code below throw exception:
```
LocalDateTime localDateTime = dateTimeFormatter.parse(value,
                                                                  LocalDateTime::from);
            return Date.from(localDateTime.atZone(getZoneId()).toInstant()); 
```
That's why we are using this part:
```
} catch (DateTimeException e) {
            LocalDate localDate = dateTimeFormatter.parse(String.valueOf(value),
                                                          LocalDate::from);
            return Date.from(localDate.atStartOfDay().atZone(getZoneId()).toInstant());
        }
```

CTI777: Strings like "1970-01-01" this may come as variable value from any place, and we should handle this, I think...

activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/variable/DateFormatterProviderTest.java
erdemedeiros: The only problem is that methods other than `hasTime()` are timezone specific. So the test may fail depending on the location from where you're running your test.

activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/DateFormatterProvider.java
erdemedeiros: yes, we should keep it in this case.

ffazzini: ok

PR #2894, opened 2019-09-07
Detailed exception when wrong variable type

ramunask:
Throw detailed exception about wrong variable type in VariableParsingService. NPE was thrown when variable type was incorrect.


ramunask:
ProcessRuntimeBPMNTimerIT test failed. But my change isn't related to this test. Trying to reopen pull request.

Issue #2637, opened 2019-03-19
[BPMN] Add Support for BPMN Message elements


salaboy:
we need to define which message scenarios we will support for the first iteration this includes

- Define conformance tests in spring-boot-starter
- Define acceptance tests to validate correct behaviour
- Enable elements in BPMN modeler
    - Define which properties are needed

igdianov:
@salaboy We also need to specify and implement audit event apis for BPMN messages

PR #2895, opened 2019-09-09
Support for BPMN message events

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2637


CTI777:
Test still should be done and updated

PR #2898, opened 2019-09-09
Support for BPMN message events

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2637

### DIFF
--- a/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/MessageTestConfiguration.java
+++ b/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/MessageTestConfiguration.java
@@ -1,0 +1,34 @@
+package org.activiti.spring.boot;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.activiti.api.process.model.events.BPMNMessageEvent;
+import org.activiti.api.process.model.events.BPMNMessageReceivedEvent;
+import org.activiti.api.process.model.events.BPMNMessageSentEvent;
+import org.activiti.api.process.model.events.BPMNMessageWaitingEvent;
+import org.activiti.api.process.runtime.events.listener.BPMNElementEventListener;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class MessageTestConfiguration {
+
+    public static Set<BPMNMessageEvent> messageEvents = new HashSet<>();
+
+    @Bean
+    public BPMNElementEventListener<BPMNMessageSentEvent> messageSentEventListener() {
+        return bpmnMessageSentEvent -> messageEvents.add(bpmnMessageSentEvent);
+    }
+    
+    @Bean
+    public BPMNElementEventListener<BPMNMessageReceivedEvent> messageReceivedEventListener() {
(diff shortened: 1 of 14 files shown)


CTI777:
Unit tests are not yet implemented

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/MessageTestConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+import org.activiti.api.process.model.events.BPMNMessageSentEvent;
+import org.activiti.api.process.model.events.BPMNMessageWaitingEvent;
+import org.activiti.api.process.runtime.events.listener.BPMNElementEventListener;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class MessageTestConfiguration {
+
+    public static Set<BPMNMessageEvent> messageEvents = new HashSet<>();
```
igdianov: This should be list to preserve order and distinct events.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeBPMNMessageIT.java
```diff
@@ -122,20 +93,23 @@ public void shouldThrowIntermediateMessageEvent() throws Exception{
...
+                            event -> ((BPMNMessage)(event.getEntity())).getMessagePayload().getBusinessKey(),
+                            event -> ((BPMNMessage)(event.getEntity())).getMessagePayload().getVariables()  
+                )
+                .contains(Tuple.tuple(BPMNMessageEvent.MessageEvents.MESSAGE_SENT,
+                                      process.getProcessDefinitionId(),
+                                      process.getId(),
                                       process.getProcessDefinitionId(),
                                       process.getId(),
-                                      "throwEvent",
                                       "Test Message",
```
igdianov: We seem to have lost activity type and other attributes from original engine event, i. e. process businessKey. It would be useful info for downstream consumers I think.

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/MessageTestConfiguration.java
CTI777: Set replaced with a list

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeBPMNMessageIT.java
CTI777: For events we have a payload where we store this info:
https://github.com/Activiti/Activiti/pull/2898/files#diff-01257cc01747b6a40cd9228263a55296R37
I think in cloud events (later) we store such kind of info for events...

But I agree, that we may need activity type for this event... Should I add this to API?

igdianov: I think this requires global attribute `activityType` which is out of the scope of this PR

CTI777:
Unit tests are added

Issue #2743, opened 2019-05-14
[BPMN] Add support to Error events


salaboy:
Add support to BMPN Error events

PR #2908, opened 2019-09-12
ACTIVITI-2743 Implementation of Error Events

CTI777:
Related to https://github.com/Activiti/Activiti/issues/2743

### DIFF
--- a/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeBPMNErrorReceivedIT.java
+++ b/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeBPMNErrorReceivedIT.java
@@ -25,6 +25,10 @@ import org.activiti.api.process.model.events.BPMNErrorReceivedEvent;
 import org.activiti.api.process.runtime.ProcessRuntime;
 import org.activiti.api.runtime.shared.query.Page;
 import org.activiti.api.runtime.shared.query.Pageable;
+import org.activiti.api.task.model.Task;
+import org.activiti.api.task.model.builders.GetTasksPayloadBuilder;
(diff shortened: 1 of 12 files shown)


CTI777:
Following tests are covered:
ErrorEndEvent, Error BoundaryEvent for subprocess
image

ErrorEndEvent, ErrorStartEvent for Event subProcess
image

Error BoundaryEvent for CallActivty
image

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeBPMNErrorReceivedIT.java
```diff
@@ -0,0 +1,231 @@
...
+                ProcessPayloadBuilder
+                        .start()
+                        .withProcessDefinitionKey(ERROR_BOUNDARY_EVENT_SUBPROCESS)
+                        .build());
+        
+        //then
+        assertNotNull(processInstance);
+        
+        //Error-handling should end the process
+        Page<ProcessInstance> processInstancePage = processRuntime.processInstances(Pageable.of(0,
```
erdemedeiros: Here, and in the following tests, I think it would be interesting to have a user task following the error catch event, so you can test that the process is still running and it's pointing to the user task.

CTI777: I have modified test processes... Looks good.
But for on case I still have to figure out, what is going on:
image

CTI777: The test is fixed now: all work like expected...

PR #2920, opened 2019-09-20
[ACTIVITI-3763] store extension file to db

robfrank:
add support to store extension files (json) on the database as process definition as well


erdemedeiros:
Covered by https://github.com/Activiti/Activiti/pull/2940

PR #2935, opened 2019-10-08
[AAE-240] fix BpmnXMLConverter for errors inside xml

roxanaspatariu:


activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.java
```diff
@@ -54,14 +54,26 @@ public void testThrowErrorWithEmptyErrorCode() {
...
-  @Deployment
+  public void testThrowErrorWithoutErrorCode() {
+    try {
+      repositoryService.createDeployment().addClasspathResource("org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.testThrowErrorWithoutErrorCode.bpmn20.xml").deploy();
+      fail("ActivitiException expected");
+    } catch (ActivitiException re) {
+    }
+  }
+
   public void testCatchErrorOnEmbeddedSubprocessWithEmptyErrorCode() {
```
erdemedeiros: Shouldn't this test keep initial `@Deployment` annotation here, so it keeps using the same process definition as before?

activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.java
```diff
@@ -54,14 +54,26 @@ public void testThrowErrorWithEmptyErrorCode() {
     }
   }
 
-  @Deployment
+  public void testThrowErrorWithoutErrorCode() {
+    try {
+      repositoryService.createDeployment().addClasspathResource("org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.testThrowErrorWithoutErrorCode.bpmn20.xml").deploy();
+      fail("ActivitiException expected");
```
erdemedeiros: I know that this class is using this pattern to test exceptions, but do you mind to migrate it to `catchThrowable` from `AssertJ`. It's more readable in my option. You can find an example here https://github.com/Activiti/activiti-cloud-query-service/blob/develop/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/events/handlers/TaskEntityCandidateGroupAddedEventHandlerTest.java#L80-L86

activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.java
```diff
@@ -54,14 +54,26 @@ public void testThrowErrorWithEmptyErrorCode() {
...
+    } catch (ActivitiException re) {
+    }
+  }
+
   public void testCatchErrorOnEmbeddedSubprocessWithEmptyErrorCode() {
+    try {
     testCatchErrorOnEmbeddedSubprocess();
+    } catch (ActivitiException re) {
```
erdemedeiros: Is this test suppose the throw exception or not? Note that you are not calling `fail` inside the `try` block, so it's going to succeed in both ways.

activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.java
```diff
@@ -54,14 +54,26 @@ public void testThrowErrorWithEmptyErrorCode() {
...
+    }
   }
 
-  @Deployment
```
erdemedeiros: Same question here: shouldn't you keep `@Deployment` annotation?

activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventTest.java
```diff
@@ -54,14 +54,26 @@ public void testThrowErrorWithEmptyErrorCode() {
...
   public void testCatchErrorOnEmbeddedSubprocessWithoutErrorCode() {
+    try {
     testCatchErrorOnEmbeddedSubprocess();
+    } catch (ActivitiException re) {
```
erdemedeiros: If it's suppose to throw an exception you need to call `fail` method or migrate to `catchThrowable`.

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3006

PR #2942, opened 2019-10-10
Refactoring for DateFormatterProvider

CTI777:
Depends on 
https://github.com/Activiti/activiti-core-common/pull/122
Refactoring for DateFormatterProvider


activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/ProcessExtensionTestApplication.java
```diff
@@ -0,0 +1,38 @@
...
+package org.activiti.spring.process;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+
+@SpringBootApplication
+public class ProcessExtensionTestApplication {
```
ffazzini: why do we need a spring boot application test in this common module?

erdemedeiros: The test is not in the common module, it's in the extensions file module. The tests were already in place before, but they were referencing specific configurations to load. Now that the data format provider was moved in another auto-configuration, we would need to add it to the list of configurations to load or let spring boot load the ones available. I preferred the second approach as it's easier to maintain.

PR #2953, opened 2019-10-25
[AAE-688] Make Process extend from BaseElement to include Message Events

davidcanonieto:
At the moment, when searching for Flow Elements ti does not include Message Events. We have a new feature that brings Message Payload Mapping and that makes use of the current mapping. We need to include messages for that. 

It needs to extend from BaseElements to be able to contain message events.

More info: https://issues.alfresco.com/jira/browse/AAE-688

### DIFF
--- a/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
+++ b/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
@@ -279,7 +279,7 @@ public class Process extends BaseElement implements FlowElementsContainer, HasEx
     this.eventListeners = eventListeners;
   }
 
-  public <BaseElementType extends BaseElement> List<BaseElementType> findBaseElementsOfType(Class<BaseElementType> type) {
+  public <BaseElementType extends BaseElement> List<BaseElementType> findFlowElementsOfType(Class<BaseElementType> type) {
     return findBaseElementsOfType(type, true);
   }
 
--- a/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
+++ b/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
@@ -280,11 +280,11 @@ public class Process extends BaseElement implements FlowElementsContainer, HasEx
   }
 
   public <BaseElementType extends BaseElement> List<BaseElementType> findFlowElementsOfType(Class<BaseElementType> type) {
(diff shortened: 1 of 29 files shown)


activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
```diff
@@ -279,45 +279,45 @@ public void setEventListeners(List<EventListener> eventListeners) {
...
       if (flowElement instanceof SubProcess) {
         if (goIntoSubprocesses) {
-          foundFlowElements.addAll(findFlowElementsInSubProcessOfType((SubProcess) flowElement, type));
+            foundBaseElements.addAll(findBaseElementsInSubProcessOfType((SubProcess) flowElement, type));
         }
       }
     }
-    return foundFlowElements;
+    return foundBaseElements;
   }
```
igdianov: @davidcanonieto What is reason to make a breaking change in public method of Process class Api and cascade all the changes into engine implementation? I am worried it may also cause problems in downstream and external projects that are using Bpmn model classes. 

It may be better to add another method in the Process class to handle the export use case if really needed, but it may not be need at all. Just changing method signature should work without breaking changes, i.e.

```java
public <FlowElementType extends FlowElement> List<FlowElementType> findFlowElementsOfType(Class<FlowElementType> type)
```
```java
public <BaseElementType extends BaseElement> List<BaseElementType> findFlowElementsOfType(Class<BaseElementType> type)

```

davidcanonieto: @igdianov I reverted the method's name. It should not break anything. This change was done in order to keep meaningful method names but i understand that it can be a breaking change. Please re-review it. Thank you

erdemedeiros:
@davidcanonieto, as discussed, this is not going to work, as messages are stored outside of the process. You can retrieve them via [BpmnModel.getMessages()](https://github.com/Activiti/Activiti/blob/develop/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/BpmnModel.java#L379). In addition, [the list where the search is performed](https://github.com/Activiti/Activiti/blob/develop/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java#L31) is of type `List<FlowElement>`, so it cannot contain `BaseElement` that is not a `FlowElement`. I'm closing this PR as the changes are no longer needed.

PR #2955, opened 2019-10-28
ACTIVITI-3987 - added bpmn2 prefix for process

VitoAlbano:


VitoAlbano:
I've added an extra unit test to check that the conversion and the xml with the prefix are equals.

PR #2963, opened 2019-11-05
AAE-23 Support placeholders in variables

pmartinezga:
This PR fix the rebasing issues on the former one: [2946](https://github.com/Activiti/Activiti/pull/2946)

Variables can be initialized when modeling with an expression that will be resolved in the input mapping of a connector or a form.

Added checks that validate that expressions are not used as a variable value in:
 - Output mappings
 - Process start payload
 - Set variables payload


pmartinezga:
Thank you for your improvements @erdemedeiros :+1:! Any further suggestions from you or @mmartinadanx will be appreciated.

erdemedeiros:
Ref https://github.com/Activiti/Activiti/pull/2963

PR #2970, opened 2019-11-11
AAE-783 Implementation of a cancel message event

CTI777:
Implementation of a cancel message event
Depends on: https://github.com/Activiti/activiti-api/pull/150


CTI777:
@igdianov 
BusinessKey is not available in EventSubscriptionEntityImpl...
Do we need it here?
It will be available later in Cloud Events

CTI777:
Will close this, because implementation was moved to:
https://github.com/Activiti/Activiti/pull/2971

PR #3001, opened 2019-11-26
[AAE-785] Add application version field to processes and tasks in Activiti Core

miguelruizdev:
Addressing this issue: https://issues.alfresco.com/jira/browse/AAE-785

### DIFF
--- a/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
+++ b/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
@@ -133,7 +133,7 @@ public class DbSqlSession implements Session {
 
         // Version 7
         ACTIVITI_VERSIONS.add(new ActivitiVersion("7.0.0.0"));
-        ACTIVITI_VERSIONS.add(new ActivitiVersion("7.1.0-M5"));
+        ACTIVITI_VERSIONS.add(new ActivitiVersion("7.1.0-M6"));
 
         /* Current */
         ACTIVITI_VERSIONS.add(new ActivitiVersion(ProcessEngine.VERSION));
--- a/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
+++ b/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
@@ -1,9 +1,9 @@
 /* Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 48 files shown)


activiti-engine/src/main/resources/org/activiti/db/create/activiti.postgres.create.engine.sql
```diff
@@ -205,6 +209,7 @@ create table ACT_RU_TASK (
     TENANT_ID_ varchar(255) default '',
     FORM_KEY_ varchar(255),
     CLAIM_TIME_ timestamp,
+    APP_VERSION_ integer,
```
erdemedeiros: IMO, as far as we keep other database scripts here, we need to keep them up to date. Could you, please, update the script for other DBs?

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeAppVersionTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE)
+public class TaskRuntimeAppVersionTest {
```
erdemedeiros: I know that most of the classes under this package are not following the convention, but it's better to use `IT` suffix instead of `Test` as it's an integration test.

activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
```diff
@@ -141,6 +133,7 @@
 
         // Version 7
         ACTIVITI_VERSIONS.add(new ActivitiVersion("7.0.0.0"));
+        ACTIVITI_VERSIONS.add(new ActivitiVersion("7.1.0-M6"));
```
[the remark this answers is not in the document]
zoltanpalfi: Yes it should, fixing it

activiti-engine/src/main/resources/org/activiti/db/create/activiti.h2.create.engine.sql
```diff
@@ -1,15 +1,15 @@
 create table ACT_GE_PROPERTY (
     NAME_ varchar(64),
-    VALUE_ varchar(300),
+    VALUE_ varchar(2000),
```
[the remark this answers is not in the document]
zoltanpalfi: To fix failing test that indicates that after a number of versions are added, the value exceeds 300

activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeAppVersionTest.java
zoltanpalfi: Changed it.

activiti-engine/src/main/resources/org/activiti/db/create/activiti.postgres.create.engine.sql
zoltanpalfi: Alright, I can add it to others as well, yes

activiti-engine/src/main/resources/org/activiti/db/create/activiti.h2.create.engine.sql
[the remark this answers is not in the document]
erdemedeiros: Can you point me to that test? Possibly we should consider updating the test. In this column we are storing values such as `7.1.0-M6`, right? In this case, 300 should be ok for a long time yet.

PR #3024, opened 2019-12-05
AAE-475 Apply variable mapping on StartEvent

CTI777:
Apply variable mapping on StartEvent


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/ProcessInstanceHelper.java
```diff
@@ -129,11 +129,13 @@ else if (bpmnModel.containsMessageId(messageRef)) {
       throw new ActivitiException("No message start event found for process definition " + processDefinition.getId() + " and message name " + messageName);
     }
 
-    // Map message payload variables before creating process instance
-    Map<String, Object> processVariables = commandContext.getProcessEngineConfiguration()
-                                                         .getEventSubscriptionPayloadMappingProvider()
-                                                         .apply(messageVariables,
-                                                                eventSubscription);
+//    // Map message payload variables before creating process instance
+//    Map<String, Object> processVariables = commandContext.getProcessEngineConfiguration()
```
erdemedeiros: If this code is not necessary anymore it should be completely removed. I guess that this is removing support for variable mappings for messages when declared inside the XML itself, but maybe it's not a big deal, as all the variable mappings are done using the extension files, right now. @igdianov do you any concern about that?

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessExtensionsTest.java
```diff
@@ -150,4 +151,36 @@ public void processInstanceFailsIfVariableTypeIncorrect() {
...
+        .isNotNull()
+        .hasSize(2)
+        .extracting(VariableInstance::getName,
+                    VariableInstance::getValue)
+        .containsOnly(
+                      tuple("name", "name_value"),
+                      tuple("email", "email_value")
+        );
+
+        processRuntime.delete(ProcessPayloadBuilder.delete(process));
```
erdemedeiros: You can skip the process clean up here as the processes are already being deleted in `@After` method.

CTI777: Done

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/ProcessInstanceHelper.java
CTI777: Done

PR #3035, opened 2019-12-16
[AAE-871] Add support for expressions in output mapping

pmartinezga:
Expression resolution in output mapping is only allowed when the mapping type is value


pmartinezga:
> Hi @pmartinezga
.... 
What do you think? Does it make sense?


Happy new year @erdemedeiros! 
As you may know, I haven't a deep knowledge of the project yet, so what I tried was implementing the desired functionality with the least changes so I could not impact other behaviors. I decided not to modify the classes for the `ExpressionManager`, `Expression`,... and tried an approach that reuses the methods in the `ExpressionResolver` as they were. But, as you pointed, introducing a class (`OutputMappingVariableScope`) that is aimed just to solve this specific problem with lots of unuseful methods, maybe is not the best choice.

Your solution looks smarter and reusable in other contexts, so I adopted it. 

Thank you!

erdemedeiros:
Happy New Year, @pmartinezga !
Great! It makes sense: I would also go with a new implementation of `VariableScope` if this interface was not that complex.

Concerning the new implementation, I noticed that the new method on `ExpressionManager` introduced some duplication on `ExpressionResolver`. I've created a new interface to remove this duplication. Please, have a look, if it seems ok for you I'll merge the PR.

Thank you!

pmartinezga:
Beautiful refactoring 👏! 
It looks perfect for me, you can merge whenever you want. 

Thank you, it's always a pleasure to work and learn from you!

erdemedeiros:

PR #3037, opened 2019-12-17
Fetching only process definitions from latest version

miguelruizdev:
Addressing this issue: https://issues.alfresco.com/jira/browse/AAE-76


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/DeploymentQueryImpl.java
```diff
@@ -197,6 +198,25 @@ public long executeCount(CommandContext commandContext) {
...
+    if (deployment != null){
+      return deployment;
+    } else {
+      List<Deployment> results = executeList(commandContext, null);
+      if (results.size() == 1) {
+        return results.get(0);
+      } else if (results.size() > 1) {
+        throw new ActivitiException("Query return " + results.size() + " results instead of max 1");
+      }
+      return null;
```
miguelruizdev: not proud of this, just mimicking the behavior of the overridden method
without it, all sorts of history tests break
fixing 10-year-old code is not my idea of fun

robfrank: No words for that.

Issue #3041, opened 2019-12-20
Collect multi-instance results


erdemedeiros:
There exists two attributes which can be used in combination with the <bpmn:multiInstanceLoopCharacteristics/> element:

`outputDataItem`: declares a variable that exists in the scope of the multi-instance activity, whereby the value of it should be collected.
`loopDataOutputRef`: declares a collection which will be available in the process instance as a variable. This collection collects the values of the result variables.

An example could look like this:
```
<multiInstanceLoopCharacteristics isSequential="false">
        <loopCardinality>2</loopCardinality>
        <loopDataOutputRef>meals</loopDataOutputRef>
        <outputDataItem name="meal"/>
      </multiInstanceLoopCharacteristics>
```
Semantic:
Whenever an instance of the specific activity completes, then the value of the variable `meal` declared by the attribute `outputDataItem` will be added to the declared collection `meals` by the attribute `loopDataOutputRef`.

PR #3042, opened 2019-12-20
AAE-506 add support to multi-instance result collection

erdemedeiros:
Mapped as `loopDataOutputRef` and `outputDataItem` as defined in the BPMN specification
Ref https://github.com/Activiti/Activiti/issues/3041

### DIFF
--- a/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/SequentialMultiInstanceBehavior.java
+++ b/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/SequentialMultiInstanceBehavior.java
@@ -81,6 +81,8 @@ public class SequentialMultiInstanceBehavior extends MultiInstanceActivityBehavi
     setLoopVariable(multiInstanceRootExecution, NUMBER_OF_COMPLETED_INSTANCES, nrOfCompletedInstances);
     setLoopVariable(childExecution, getCollectionElementIndexVariable(), loopCounter);
     logLoopDetails(childExecution, "instance completed", loopCounter, nrOfCompletedInstances, nrOfActiveInstances, nrOfInstances);
+
+    updateResultCollection(childExecution, multiInstanceRootExecution);
     
     Context.getCommandContext().getHistoryManager().recordActivityEnd((ExecutionEntity) childExecution, null);
     callActivityEndListeners(childExecution);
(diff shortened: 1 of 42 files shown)


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/variable/JsonType.java
```diff
@@ -50,40 +49,25 @@ public boolean isCachable() {
...
-    return jsonValue;
+    return loadedValue;
   }
 
-
   public void setValue(Object value, ValueFields valueFields) {
     try {
       valueFields.setTextValue(objectMapper.writeValueAsString(value));
+      if (value != null) {
+          valueFields.setTextValue2(value.getClass().getName());
```
erdemedeiros: Here, I'm reusing a field that already exists but that's not used by JSON type (`textValue2`) to store the value of the original Java Class. Is it ok or is it better to create a new column in the database base (`javaType`) for that?

erdemedeiros: I'll create a specific field: it will be cleaner.

eromano: I agree

balsarori: FYI, for JPA variables types textValue and textValue2 were used on for the value and the other for the class name. 
https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/variable/JPAEntityVariableType.java#L67

erdemedeiros: Thanks, @balsarori. Yes, I reached that code as well when I was trying to understand what was the purpose of `textValue2`. It's just hard to know who is who, with such generic names. That's why I was thinking about creating an `originalClassName` attribute.

erdemedeiros: For simplicity, I'll use `textValue2` without adding a new field.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/SequentialMultiInstanceBehavior.java
```diff
@@ -81,13 +81,16 @@ public void leave(DelegateExecution childExecution) {
     setLoopVariable(multiInstanceRootExecution, NUMBER_OF_COMPLETED_INSTANCES, nrOfCompletedInstances);
     setLoopVariable(childExecution, getCollectionElementIndexVariable(), loopCounter);
     logLoopDetails(childExecution, "instance completed", loopCounter, nrOfCompletedInstances, nrOfActiveInstances, nrOfInstances);
+
+    updateResultCollection(childExecution, multiInstanceRootExecution);
     
     Context.getCommandContext().getHistoryManager().recordActivityEnd((ExecutionEntity) childExecution, null);
     callActivityEndListeners(childExecution);
     
     //executeCompensationBoundaryEvents(execution.getCurrentFlowElement(), execution);
```
eromano: is still needed this commented code?

erdemedeiros: Commented code is never needed :). I removed it.

PR #3055, opened 2020-01-07
fix(Updatebot): Revert removed propagation to cloud repositories

igdianov:
This PR fixes broken Updatebot propagation dependencies to `activiti-cloud-runtime-bundle-service`, `ctiviti-cloud-query-service`, and `activiti-cloud-modeling-service` repositories.

### DIFF
--- a/.updatebot.yml
+++ b/.updatebot.yml
@@ -2,9 +2,18 @@ github:
   organisations:
   - name: activiti
     repositories:
+    - name: activiti-cloud-runtime-bundle-service
+      branch: develop
+      useSinglePullRequest: true
+    - name: activiti-cloud-query-service
+      branch: develop
+      useSinglePullRequest: true
     - name: activiti-cloud-service-common
       branch: develop
       useSinglePullRequest: true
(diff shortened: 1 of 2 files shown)


igdianov:
> this should be reverted, there is no need to add it back, everything in this projects is resolved from the reactor

Can you clarify what should be reverted? The repository management have been removed from all modules in this commit: 6b198218f8809cdbd500280d0c1253bcfeaac65d. I have added it to main  pom.xml to be able to resolve it from reactor.

mteodori:
> > this should be reverted, there is no need to add it back, everything in this projects is resolved from the reactor
> 
> Can you clarify what should be reverted? The repository management have been removed from all modules in this commit: [6b19821](https://github.com/Activiti/Activiti/commit/6b198218f8809cdbd500280d0c1253bcfeaac65d). I have added it to main pom.xml to be able to resolve it from reactor because it was missing.

the repository block is required only by projects referencing the dependencies bom from this one, this project is self contained so it doesn't need that configuration

Issue #3025, opened 2019-12-07
Oldest process definition is being picked up when starting process by key


ramunask:
When starting new process with `StartProcessPayload` and passing `processDefinitionKey` new process is being started with oldest process definition version instead of newest.
Issue is in `org.activiti.runtime.api.impl.ProcessRuntimeImpl` file. Instead of '.orderByProcessDefinitionVersion().asc()' should be '.orderByProcessDefinitionVersion().desc()':
```java
    @Override
    public ProcessDefinition processDefinition(String processDefinitionId) {
        org.activiti.engine.repository.ProcessDefinition processDefinition;
        // try searching by Key if there is no matching by Id
        List<org.activiti.engine.repository.ProcessDefinition> list = repositoryService
                .createProcessDefinitionQuery()
                .processDefinitionKey(processDefinitionId)
                .orderByProcessDefinitionVersion()
                .asc()
                .list();
        if (!list.isEmpty()) {
            processDefinition = list.get(0);
        } else {
            processDefinition = repositoryService.getProcessDefinition(processDefinitionId);
        }
        if (!securityPoliciesManager.canRead(processDefinition.getKey())) {
            throw new ActivitiObjectNotFoundException("Unable to find process definition for the given id:'" + processDefinitionId + "'");
        }
        return processDefinitionConverter.from(processDefinition);
    }
```
Issue is in `7.1.0.M5` version.

PR #3056, opened 2020-01-10
fix: returning ProcessDefinition latest version when finding by key

zoltanpalfi:


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3025

PR #3086, opened 2020-01-21
AAE-944 - Provide candidates info with in the getTaskDetailsById API

jesty:


activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeCandidatesTest.java
```diff
@@ -93,13 +101,73 @@ public void createStandaloneTaskAndDeleteAndAddUserCandidates() {
...
         userCandidates = taskRuntime.userCandidates(createTask.getId());
         assertThat(userCandidates).isNotNull();
         assertThat(userCandidates.size()).isEqualTo(1);
-           
      }
 
+    @Test
+    public void should_addAndRemoveCandidateGroup() {
+
+        RuntimeTestConfiguration.taskCandidateUserRemovedEvents.clear();
```
miguelruizdev: Don't we have an @After test cleanup?
All good otherwise 👍

jesty: The cleanup already exists at the beginning of the test::
```
 @After
    public void taskCleanUp(){
        taskCleanUpUtil.cleanUpWithAdmin();
    }
```

PR #3093, opened 2020-01-23
AAE-1538 Create a task set logged user as a candidate user

jesty:
Create a task set logged user as a candidate user


activiti-api/activiti-api-task-model/src/main/java/org/activiti/api/task/model/builders/CreateTaskPayloadBuilder.java
```diff
@@ -75,7 +75,7 @@ public CreateTaskPayloadBuilder withCandidateUsers(List<String> candidateUsers)
     }
 
     public CreateTaskPayloadBuilder withCandidateUsers(String candidateUsers) {
-        this.candidateGroups.add(candidateUsers);
+        this.candidateUsers.add(candidateUsers);
```
erdemedeiros: Nice catch!

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeCandidatesTest.java
```diff
@@ -166,8 +160,22 @@ public void should_addAndRemoveCandidateGroup() {
...
+        taskRuntime.addCandidateGroups(TaskPayloadBuilder
+                                               .addCandidateGroups()
+                                               .withTaskId(createTask.getId())
+                                               .withCandidateGroup("test")
+                                               .build());
+
+        task = taskRuntime.task(createTask.getId());
+        groupCandidatesOnTask = task.getCandidateGroups();
+        assertThat(groupCandidatesOnTask).isNotNull();
+        assertThat(groupCandidatesOnTask.size()).isEqualTo(1);
```
erdemedeiros: Tip: you can use the shortcut `assertThat(groupCandidatesOnTask).hasSize(1)`. This will also check that `groupCandidatesOnTask` is not null, so you can remove line `172`.

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -262,8 +262,6 @@ public Task create(CreateTaskPayload createTaskPayload) {
         task.setFormKey(createTaskPayload.getFormKey());
         task.setOwner(securityManager.getAuthenticatedUserId());
         taskService.saveTask(task);
-        taskService.addCandidateUser(task.getId(),
```
erdemedeiros: With this change, the newly created task will be available only for the admin with no candidate user/group is provided in the payload. Is that the expected behavior?

jesty: As far as I know yes. The front-end check that there is at least an assignee or a candidate user o group.
We can improve this code doing an additional check: if there is an assignee or a candidate user/group, nothing will happen, otherwise, we can set the current user as candidate user. 
@mauriziovitale what do you think?

mauriziovitale: Every time we create a standalone task we provide the assignee. So the task will be available for him and because there are no candidate that user should not be allowed to release a task. @erdemedeiros

erdemedeiros: Currently, there is no check at the API level, but it should be ok for now.

PR #3094, opened 2020-01-23
fix: throw unprocessable entry exception when process doesn't belongs to latest deployment

zoltanpalfi:
Part of AAE-1503

### DIFF
--- a/activiti-api/activiti-api-runtime-shared/src/main/java/org/activiti/api/runtime/shared/UnprocessableEntryException.java
+++ b/activiti-api/activiti-api-runtime-shared/src/main/java/org/activiti/api/runtime/shared/UnprocessableEntryException.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2020 Alfresco, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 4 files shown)


activiti-api/activiti-api-runtime-shared/src/main/java/org/activiti/api/runtime/shared/UnprocessableEntryException.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.activiti.api.runtime.shared;
+
+public class UnprocessableEntryException extends IllegalStateException {
```
zoltanpalfi: as per @miguelruizdev `Entry`-> `Entity`,  IllegalArgumentException would be better

PR #3111, opened 2020-01-27
AAE-1572 Enforced application version lib

miguelruizdev:
Addressing this issue: https://issues.alfresco.com/jira/browse/AAE-1572


activiti-core/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
```diff
@@ -128,13 +128,21 @@ private boolean isProcessDefinitionResource(String resource) {
...
+        if (applicationUpgradeContextService != null && applicationUpgradeContextService.hasProjectManifest()) {
             try {
-                deploymentBuilder.setProjectManifest(projectModelService.loadProjectManifest());
+                deploymentBuilder.setProjectManifest(applicationUpgradeContextService.loadProjectManifest());
             } catch (IOException e) {
                 LOGGER.warn("Manifest of application not found. Project release version will not be set for deployment.");
             }
         }
+
+        if(applicationUpgradeContextService != null && applicationUpgradeContextService.hasEnforcedAppVersion()){
```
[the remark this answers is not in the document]
zoltanpalfi: Will `applicationUpgradeContextService` ever be null? I don't think so, and that should be some sort of invalid state of the application, and checked earlier (constructor)

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/repository/DeploymentBuilderImpl.java
```diff
@@ -97,6 +98,19 @@ public boolean hasProjectManifestSet(){
...
+      this.enforcedAppVersion = enforcedAppVersion;
+      return this;
+  }
+
+  public Integer getEnforcedAppVersion(){
+      return this.enforcedAppVersion;
+  }
+
+  public boolean hasEnforcedAppVersion(){
+      return this.enforcedAppVersion != null;
```
zoltanpalfi: Why is in here a different type of check? Isn't this supposed to behave the same way? We should be using different names for different concepts and same names only to the same concepts

activiti-core-common/activiti-spring-project/src/main/java/org/activiti/core/common/spring/project/conf/ApplicationUpgradeContextAutoConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnMissingClass(value = "org.springframework.http.converter.json.Jackson2ObjectMapperBuilder")
+    public ObjectMapper objectMapper() {
+        return new ObjectMapper();
+    }
+
+    @Bean
+    public ApplicationUpgradeContextService applicationUpgradeContextService(@Value("${project.manifest.file.path:classpath:/default-app.json}") String absolutePath,
+                                                                             @Value("${application.version:0}") Integer enforcedAppVersion,
```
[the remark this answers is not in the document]
miguelruizdev: It has to be configurable since this variable is coming from Deployment service in the same fashion is being set for form and DMN services.

activiti-core/activiti-spring/src/main/java/org/activiti/spring/autodeployment/AbstractAutoDeploymentStrategy.java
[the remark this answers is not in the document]
miguelruizdev: Yes, applicationUpgradeContextService can actually be null, since its instantiation depends on the existence of a manifest file in the project, something that does not happing in the engine modules.
But can perform a null check before this point, sure 👍

activiti-core-common/activiti-spring-project/src/test/java/org/activiti/core/common/spring/project/ApplicationUpgradeContextServiceNullValuesIT.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.*;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = ApplicationUpgradeContextAutoConfiguration.class, webEnvironment = SpringBootTest.WebEnvironment.NONE)
```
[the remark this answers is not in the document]
miguelruizdev: This test is covering the ability of our service to read a property set externally (by the deployment service in the case of AAE), in the same fashion as in here: 
https://github.com/Activiti/Activiti/blob/develop/activiti-core-common/activiti-spring-project/src/test/java/org/activiti/core/common/spring/project/ProjectModelServiceNoManifestIT.java
and here:
https://github.com/Activiti/Activiti/blob/develop/activiti-core-common/activiti-spring-project/src/test/java/org/activiti/core/common/spring/project/ProjectModelServiceIT.java

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/repository/DeploymentBuilderImpl.java
miguelruizdev: this `enforcedAppVersion` check is done after the  enforcedAppVersion loading from the `applicationUpgradeContextService`
they hold the same value, but they are different fields belonging to different logic layers.
Hence, their default status is different (0 for the `applicationUpgradeContextService`, and null for this one in the DeploymentBuilder

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/DeployCmd.java
```diff
@@ -126,7 +131,12 @@ protected boolean deploymentsDiffer(DeploymentEntity deployment,
 
         if (deploymentBuilder.hasProjectManifestSet()) {
 
-            return !deployment.getProjectReleaseVersion().equals(saved.getProjectReleaseVersion());
+            if (!deployment.getProjectReleaseVersion().equals(saved.getProjectReleaseVersion())) {
```
ffazzini: what about if the project release versions mismatch (returns true)
but the appVersions match instead?

Based on the logic that the enforces appVersion has priority.. would we have a problem in this scenario?

activiti-core-common/activiti-spring-project/src/test/java/org/activiti/core/common/spring/project/ApplicationUpgradeContextServiceIT.java
```diff
@@ -14,19 +14,19 @@
...
-public class ProjectModelServiceIT {
+public class ApplicationUpgradeContextServiceIT {
 
     @Autowired
-    private ProjectModelService projectModelService;
+    private ApplicationUpgradeContextService applicationUpgradeContextService;
 
     @Test
-    public void should_RetrieveManifest_When_Called() throws IOException {
+    public void should_retrieveManifest_when_called() throws IOException {
```
ffazzini: "when_called" can be removed if not giving a specific meaning

activiti-core-common/activiti-spring-project/src/main/java/org/activiti/core/common/spring/project/ApplicationUpgradeContextService.java
```diff
@@ -45,11 +49,19 @@ public ProjectManifest loadProjectManifest() throws IOException {
...
+                        .getInputStream());
     }
 
-    public boolean hasProjectManifest(){
+    public boolean hasProjectManifest() {
         return retrieveResource().isPresent();
     }
+
+    public boolean hasEnforcedAppVersion() {
+        return this.enforcedAppVersion != 0;
```
ffazzini: would > 0 be better in case of negative app version passed?

activiti-core-common/activiti-spring-project/src/test/java/org/activiti/core/common/spring/project/ApplicationUpgradeContextServiceNullValuesIT.java
```diff
@@ -0,0 +1,38 @@
...
+
+        assertThat(applicationUpgradeContextService.hasProjectManifest()).isFalse();
+        Throwable thrown = catchThrowable(() -> applicationUpgradeContextService.loadProjectManifest());
+        assertThat(thrown)
+            .isInstanceOf(FileNotFoundException.class)
+            .hasMessageContaining("manifest not found");
+    }
+
+    @Test
+    public void should_not_haveDefaultEnforcedAppVersion() {
```
ffazzini: should_notHaveDefault....

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/ApplicationUpgradeIT.java
```diff
@@ -96,6 +96,35 @@ public void should_getLatestProcessDefinitionByKey_when_multipleVersions() {
 
     }
 
+    @Test
+    public void should_UpdateDeploymentVersion_when_SameProjectManifestVersionAndEnforcedAppVersionIsSet(){
```
ffazzini: "should_update..." 
u lowercase
"when_sameProject..."
s lowercase

activiti-core-common/activiti-spring-project/src/main/java/org/activiti/core/common/spring/project/ApplicationUpgradeContextService.java
miguelruizdev: This enforced version is set by the deployment service, and it is a counter AFAIK
But we can add that check just in case the env variable is set via other mechanisms we don't control/expect

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/DeployCmd.java
miguelruizdev: The enforced app version has priority only when project release versions mismatch (in the rest of cases is the original upgrade logic that is set into motion, that has been released and fully tested).
This is to make a distinction between the creation of a new pod when we upgrade (and there's a new app version enforced) and the creation of a new pod when there's an outage (no upgrade, no +1 in the counter, so the service relies on the env variable, that if it does not change, effectively means there was no upgrade at all )

ffazzini: the enforce app version has priority on everything.
i don't understand why this logic is applied "only when project release versions mismatch"

if you rely on enforced app version first, then you don't have problem with this either:
"This is to make a distinction between the creation of a new pod when we upgrade (and there's a new app version enforced) and the creation of a new pod when there's an outage (no upgrade, no +1 in the counter, so the service relies on the env variable, that if it does not change, effectively means there was no upgrade at all )"

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/DeployCmd.java
```diff
@@ -124,9 +127,9 @@ private void setProjectReleaseVersion(DeploymentEntity deployment) {
     protected boolean deploymentsDiffer(DeploymentEntity deployment,
                                         DeploymentEntity saved) {
 
-        if (deploymentBuilder.hasProjectManifestSet()) {
+        if (deploymentBuilder.hasEnforcedAppVersion()) {
+            return !deploymentBuilder.getEnforcedAppVersion().equals(saved.getVersion());
 
-            return !deployment.getProjectReleaseVersion().equals(saved.getProjectReleaseVersion());
```
ffazzini: without app version enforced set, what will be the default behaviour? Will it try to use the manifest file to compare the project release versions? Removing this line looks not

miguelruizdev: if the enforced application version is not set, no upgrade logic will be triggered
In the event of no enforced version set, do we want it to fall back into the old upgrade logic that was relying on the manifest?

miguelruizdev: do we want this:
```

        if (deploymentBuilder.hasEnforcedAppVersion()) {
            return !deploymentBuilder.getEnforcedAppVersion().equals(saved.getVersion());
        }else if (deploymentBuilder.hasProjectManifestSet()){
            return !deployment.getProjectReleaseVersion().equals(saved.getProjectReleaseVersion());
        } else {
           //old code
        }
```

ffazzini: yes

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/DeployCmd.java
```diff
@@ -47,48 +47,86 @@ public Deployment execute(CommandContext commandContext) {
...
+                }
+            }
+
+            DeploymentEntity existingDeployment = null;
+            if (!existingDeployments.isEmpty()) {
+                existingDeployment = (DeploymentEntity) existingDeployments.get(0);
+            }
+
+            if (existingDeployment != null) {
+  ///////////////////
```
ffazzini: comments?

PR #3116, opened 2020-01-29
test: create with correct state and version executions

zoltanpalfi:
Tests for create process instance, sub process, child process.
Closes AAE-1603, AAE-1600

### DIFF
--- a/activiti-core/activiti-engine/src/test/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImplTest.java
+++ b/activiti-core/activiti-engine/src/test/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImplTest.java
@@ -61,8 +61,7 @@ public class ExecutionEntityManagerImplTest {
     }
 
     @Test
-    public void should_createProcessInstanceExecution_when_NoInitialVariable() {
-        //given
+    public void should_createProcessInstanceExecution_when_initialVariableNameIsNull() {
         ProcessDefinitionEntity processDefinition = new ProcessDefinitionEntityImpl();
         processDefinition.setId("processDefinitionId");
         processDefinition.setKey("processDefinitionKey");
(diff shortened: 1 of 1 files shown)


activiti-core/activiti-engine/src/test/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImplTest.java
```diff
@@ -0,0 +1,183 @@
...
+    @Before
+    public void setUp() throws Exception {
+        initMocks(this);
+        given(processEngineConfiguration.getClock()).willReturn(clock);
+        given(processEngineConfiguration.getEventDispatcher()).willReturn(eventDispatcher);
+        Context.setProcessEngineConfiguration(processEngineConfiguration);
+    }
+
+    @Test
+    public void should_createProcessInstanceExecution_when_NoInitialVariable() {
```
ffazzini: i don't understand the method name -> "NoInitialVariable" What variable does it refer to?

activiti-core/activiti-engine/src/test/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImplTest.java
```diff
@@ -0,0 +1,183 @@
...
+        //given
```
ffazzini: given when then comments can be removed

activiti-core/activiti-engine/src/test/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImplTest.java
zoltanpalfi: `when_initialVariableNull` is better?

zoltanpalfi: basically I test the case when the on the processInstance creation we are not setting any variable

ffazzini: when_initialVariableNameIsNull ? 
I would like to add "Name" as it is in the method signature
https://github.com/Activiti/Activiti/blob/fc2793f5f896c9d9b0acdb52cf44904b67e882b1/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImpl.java#L210

ffazzini: is there a special behaviour triggered when initialVariableName is null?

PR #3117, opened 2020-01-30
AAE-1610 - fixing mapping variable with the new multi process structure

VitoAlbano:


activiti-core-common/activiti-spring-resource-finder/src/main/java/org/activiti/spring/resources/ResourceReader.java
```diff
@@ -24,6 +24,6 @@
 
     Predicate<String> getResourceNameSelector();
 
-    T read(InputStream inputStream) throws IOException;
+    T read(InputStream inputStream, String processDefinitionKey) throws IOException;
```
erdemedeiros: I'd prefer to not modify this method signature.

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/VariablesMappingProvider.java
```diff
@@ -69,13 +69,13 @@ public VariablesMappingProvider(ProcessExtensionService processExtensionService,
 
         Map<String, Object> constants = calculateConstants(execution, extensions);
 
-        if (extensions.getExtensions().hasEmptyInputsMapping(execution.getCurrentActivityId())) {
+        if (extensions.getExtensionsByProcessDefinitionId(execution.getProcessDefinitionId()).hasEmptyInputsMapping(execution.getCurrentActivityId())) {
```
erdemedeiros: I think it's better to return directly an `Extension` instead of `ProcessExtensionModel` as we are already providing the process `definitionId`. Let's modify the signature of `ProcessExtensionService.getExtensionsForId`

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/ProcessExtensionModel.java
```diff
@@ -13,21 +13,32 @@
...
     }
-    public Extension getExtensions() {
-        return extensions;
+
+    public Extension getExtensions(String processDefinitionKey) {
+        return extensions.get(processDefinitionKey);
+    }
+
+    public Extension getExtensionsByProcessDefinitionId(String processDefinitionId) {
+        String processDefinitionKey = processDefinitionId.split(":")[0];
```
erdemedeiros: Do not make this assumption, I'm not sure that this is true all the time.

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessVariablesPayloadValidator.java
```diff
@@ -63,7 +63,7 @@ public ProcessVariablesPayloadValidator(DateFormatterProvider dateFormatterProvi
                                                       null;
 
         return Optional.ofNullable(processExtensionModel)
-                .map(ProcessExtensionModel::getExtensions)
+                .map(p -> p.getExtensionsByProcessDefinitionId(processDefinitionId))
```
ffazzini: I thought we did not have the processDefinitionId in the extensions json
I thought we had only the process definition key in the extensions json

VitoAlbano: processDefinitionId is basically the combination of processDefinitionKey:appVersion:deploymentId and it's immutable as far as i know in the whole life of that version of the process.

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/ProcessExtensionModel.java
VitoAlbano: When this is not true?

erdemedeiros: The good news is that after the refactoring this code is not needed anymore!

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3138

Issue #3119, opened 2020-01-31
Wrong Invalid reference in diagram warning


eromano:
All the time we validate a process that contains a Collaboration that shows the Message Flows. We receive a warning message from the Validator:
```
[main|org.activiti.engine.impl.bpmn.parser.BpmnParse] WARN  Invalid reference in diagram interchange definition: could not find messageflow2
```
In Collaboration Diagrams, a Message Flow can be extended to show the message that is passed from one Participant to another. 
Message Flow is defined it is contained either within a Collaboration. We need to let the validator search for a bpmnReference of the bpmnElement also inside the MessageFlow elements

PR #3120, opened 2020-01-31
[AAE-1614] fix warning for message flow reference

eromano:
https://github.com/Activiti/Activiti/issues/3119

All the time we validate a process that contains a Collaboration that shows the Message Flows. We receive a warning message from the Validator:

[main|org.activiti.engine.impl.bpmn.parser.BpmnParse] WARN  Invalid reference in diagram interchange definition: could not find messageflow2
In Collaboration Diagrams, a Message Flow can be extended to show the message that is passed from one Participant to another.
Message Flow is defined it is contained either within a Collaboration. We need to let the validator search for a bpmnReference of the bpmnElement also inside the MessageFlow elements


erdemedeiros:
@Mergifyio refresh

PR #3122, opened 2020-02-03
AAE-1665 Add missing ElResolvers to output mapping expressions

erdemedeiros:
In addition to MapResolver, other base resolver should be included as well, i.e. ListELResolver, JsonNodeELResolver


erdemedeiros:

PR #3124, opened 2020-02-04
AAE-1665 Inject ExpressionManager during expressions evaluation

erdemedeiros:
Allow evaluating expressions even when internal Context is not set. This is useful in the case of cloud connectors.


erdemedeiros:

PR #3125, opened 2020-02-04
Set the reason for cancellation in TerminateEnd event

miguelruizdev:
Addressing this issue: https://issues.alfresco.com/jira/browse/AAE-189


activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/TaskCancelledImpl.java
```diff
@@ -23,16 +23,23 @@
 
 public class TaskCancelledImpl extends RuntimeEventImpl<Task, TaskRuntimeEvent.TaskEvents> implements TaskCancelledEvent {
 
+    private String reason;
+
     public TaskCancelledImpl() {
```
robfrank: is this constructor still useful?
if not, remove and make "reason" final

activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/event/impl/ActivitiEntityEventImpl.java
```diff
@@ -24,18 +24,30 @@
...
-  @Override
-  public Object getEntity() {
-    return entity;
-  }
+    @Override
+    public String getReason() {
+        return reason;
+    }
+
+    public void setReason(String reason) {
```
robfrank: is this setter really necessary? can thew reason be passed as constructor param?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/event/impl/ActivitiEventBuilder.java
```diff
@@ -388,6 +389,7 @@ protected static void populateEventWithCurrentContext(ActivitiEventImpl event) {
         event.setProcessInstanceId(((Task) persistedObject).getProcessInstanceId());
         event.setExecutionId(((Task) persistedObject).getExecutionId());
         event.setProcessDefinitionId(((Task) persistedObject).getProcessDefinitionId());
+        ((ActivitiEntityEventImpl)event).setReason(TerminateEndEventActivityBehavior.createDeleteReason(null));
```
robfrank: this is hurting me

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/TaskCancelledImpl.java
miguelruizdev: True, we can get rid of that no-argument constructor

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/TaskCancelledImpl.java
```diff
@@ -23,16 +23,23 @@
...
     }
 
-    public TaskCancelledImpl(Task entity) {
+    public TaskCancelledImpl(Task entity, String reason) {
```
robfrank: Can we remove the empty constructor and let "reason" to be final?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/event/impl/ActivitiEventBuilder.java
miguelruizdev: it hurts for a reason
this should have never been committed, just a leftover from the debugging session I went through 😄

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/event/impl/TaskCancelledImpl.java
miguelruizdev: true, that constructor is not used anymore

miguelruizdev: good catch about the reason field being final (it should never change once the particular instance of the object is created)

activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/event/impl/ActivitiEntityEventImpl.java
miguelruizdev: Also, commited by mistake and directly related to the above comment

PR #3146, opened 2020-02-14
fixing timer job lock issue if fetch more than 1 record

deanwong:
The original AcquireTimerJobsCmd as below
`
final AcquiredTimerJobEntities acquiredJobs = commandExecutor.execute(new AcquireTimerJobsCmd(asyncExecutor));
`
will fetch timer jobs and lock one by one;
Then begin a new transaction to `moveTimerJobToExecutableJob`, if error occurred the loop will break. But the AcquiredTimerJobs have been locked like this
|  ID_  | REV_ | TYPE_ | LOCK_EXP_TIME_ | LOCK_OWNER_
|  ----  | ----  | ----  | ----  | ----  |
| 811f2eb2-4f00-11ea-bef1-0a580a81015e  | 1 | timer | 2020-02-14 14:54:41 | 34e166b8-4f01-11ea-b993-0a580a820166 |

If the `maxAsyncJobsDuePerAcquisition` is more than 1, it will cause lots of timer jobs sleep in `act_ru_timer_job` and can not be woke up any more.  
There is no methods to query expired jobs in `TimerJobEntityManager`.


erdemedeiros:
Hi @deanwong, thank you for your contribution. However, the version `6.x` is no longer maintained. Your PR should targe `develop` branch.

PR #3161, opened 2020-02-25
#3160 change order of generated xml elements

CTI777:
Related to 
https://github.com/Activiti/Activiti/issues/3160

### DIFF
--- a/activiti-core/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/BaseBpmnXMLConverter.java
+++ b/activiti-core/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/BaseBpmnXMLConverter.java
@@ -205,13 +205,13 @@ public abstract class BaseBpmnXMLConverter implements BpmnXMLConstants {
       xtw.writeEndElement();
     }
 
+    writeAdditionalChildElements(baseElement, model, xtw);
+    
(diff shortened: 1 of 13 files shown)


activiti-core/activiti-bpmn-converter/src/main/java/org/activiti/bpmn/converter/BaseBpmnXMLConverter.java
```diff
@@ -205,14 +205,14 @@ public void convertToXML(XMLStreamWriter xtw, BaseElement baseElement, BpmnModel
       xtw.writeEndElement();
     }
 
+    writeAdditionalChildElements(baseElement, model, xtw);
```
erdemedeiros: This change is going to break multi-instanciated script tasks because it will move `<script>` element before `<multiInstanceLoopCharacteristics>`. I'd suggest stopping using the generic method `writeAdditionalChildElements` to write `incoming` and `outgoing` flow elements and create a new method `writeIncomingOutgoingFlowElements` for that. Then we have:
- writeIncomingOutgoingFlowElements
- writeMultiInstances
- writeAdditionalChildElements

Does it make sense?

CTI777: Changes are made

PR #3162, opened 2020-02-27
fix updatebot configuration to propagate to new activiti cloud monorepo

robfrank:
fix updatebot configuration to propagate to new activiti cloud monorepo


erdemedeiros:
@Mergify refresh

Issue #3165, opened 2020-02-28
Basic connector example in Activiti core is broken


CTI777:
When running the example:
 https://github.com/Activiti/Activiti/blob/develop/activiti-examples/activiti-api-basic-connector-example/src/main/java/org/activiti/examples/DemoApplication.java
an error occured: deployment version is 1, and processDefinition version - null...

Most probably it happens because this file is missing 
https://github.com/Activiti/example-runtime-bundle/blob/master/src/main/resources/default-project.json

it ’s referenced here https://github.com/Activiti/example-runtime-bundle/blob/master/src/main/resources/application.properties#L48

CTI777:
When we get all processDefinitions 
(we check `appVersion==null || appVersion().equals(currentVersion)`)

 ```
private List<org.activiti.engine.repository.ProcessDefinition> filterCurrentVersionDefinitions (List<org.activiti.engine.repository.ProcessDefinition> allDefinitions){
        int currentVersion = selectLatestDeployment().getVersion();
        return allDefinitions
                .stream()
                .filter(processDefinition -> processDefinition.getAppVersion() == null ||
                                             processDefinition.getAppVersion().equals(currentVersion))
                //we fetch possible unversioned definitions from different types of deployments
                .collect(Collectors.toList());
    }
```

Then, when we start a process, we check appVersion (and it fails here):
 ```
private void checkProcessDefinitionBelongsToLatestDeployment(org.activiti.engine.repository.ProcessDefinition processDefinition){
        if (!selectLatestDeployment().getVersion().equals(processDefinition.getAppVersion())) {
            throw new UnprocessableEntityException("Process definition with the given id:'" + processDefinition.getId() + "' belongs to a different application version.");
        }
  }
```

We can do following to fix the problem:
private void `checkProcessDefinitionBelongsToLatestDeployment(org.activiti.engine.repository.ProcessDefinition processDefinition){`
        **if (processDefinition.getAppVersion()  != null) {**
```
 if (!selectLatestDeployment().getVersion().equals(processDefinition.getAppVersion())) {
      throw new UnprocessableEntityException("Process definition with the given id:'" + processDefinition.getId() + "' belongs to a different application version.");
         }
```
        }
` }`

PR #3166, opened 2020-02-28
#3165 fix problem with null appVersion in processDefinition

CTI777:
Related to https://github.com/Activiti/Activiti/issues/3165


erdemedeiros:
This should fix https://github.com/Activiti/Activiti/issues/3194 as well.

zoltanpalfi:
This should also fix: #3194

PR #3174, opened 2020-03-04
Form ID generation policy

heartlhj:
Form ID generation policy,by checking the database.Reference resources activiti-engine.
Of course,can be extends FormEngineConfigurator,override configure method.


heartlhj:
travis passed

heartlhj:
Codac analyse

Issue #3186, opened 2020-03-15
Enable integration tests by default


mteodori:
Currently, integration tests are only launched when the property `-DskipITs=false` is set. This is counterintuitive and causes a lot of confusion. It's better to have them enabled by default and have the possibility to skip them with `-DskipITs=true`.

PR #3190, opened 2020-03-19
Enable integration tests by default (#3186)

mteodori:


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3186

PR #3199, opened 2020-03-25
[DW-1664] - added a create process api

VitoAlbano:


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/CreateProcessInstanceCmd.java
```diff
@@ -0,0 +1,100 @@
...
+        DeploymentManager deploymentCache = commandContext.getProcessEngineConfiguration().getDeploymentManager();
+        ProcessDefinition processDefinition = this.getProcessDefinitionByProcessDefinitionId(processDefinitionId, deploymentCache);
+        if(processDefinition == null) {
+            processDefinition = (processDefinitionKey != null && hasNoTenant(tenantId)) ?
+                this.getProcessDefinitionByProcessDefinitionKey(processDefinitionKey, tenantId, deploymentCache):
+                this.getProcessDefinitionByProcessDefinitionKeyAndTenantId(processDefinitionKey, tenantId, deploymentCache);
+            if (processDefinition == null) {
+                throw new ActivitiObjectNotFoundException("No process definition found for key '" + processDefinitionKey + "' for tenant identifier " + tenantId, ProcessDefinition.class);
+            }
+        }
```
erdemedeiros: I'd prefer to have `CreateProcessInstanceCmd` and `StartProcessInstanceCmd` sharing the same code to retrieve the process definition.

VitoAlbano: I have created a separated task for this, and will do this when picking up that one.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/StartProcessInstanceByMessageCmd.java
```diff
@@ -1,87 +1,87 @@
...
+
+    DeploymentManager deploymentCache = commandContext.getProcessEngineConfiguration().getDeploymentManager();
+
+    ProcessDefinition processDefinition = deploymentCache.findDeployedProcessDefinitionById(processDefinitionId);
+    if (processDefinition == null) {
+      throw new ActivitiObjectNotFoundException("No process definition found for id '" + processDefinitionId + "'", ProcessDefinition.class);
+    }
+
+    ProcessInstanceHelper processInstanceHelper = commandContext.getProcessEngineConfiguration().getProcessInstanceHelper();
+    ProcessInstance processInstance = processInstanceHelper.createAndStartProcessInstanceByMessage(processDefinition, businessKey, messageName, processVariables, transientVariables);
```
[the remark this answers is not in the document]
VitoAlbano: it was unused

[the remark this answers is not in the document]
erdemedeiros: That's a very good reason to remove it then ;)

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/test/java/org/activiti/runtime/api/impl/ProcessRuntimeImplTest.java
```diff
@@ -175,4 +187,56 @@ public void should_throwActivitiObjectNotFoundException_when_canReadFalse() {
             .hasMessage("Unable to find process definition for the given id:'processDefinitionId'");
     }
 
+    @Test
+    public void should_createAprocessInstance_whenCreateIsCalled() {
```
erdemedeiros: IMO, this test is two complex because it became a mix of unit test and integration test (I know, other tests in the class are doing the same, but they shouldn't :)). I think that at this level `RepositoryService` should be a mock, then there should be no reference to `CommandExecutor`. You should simply check that `ValidatorService` `RuntimeService` and  `builder` are called.

Then, in addition to this unit test, you should have an Integration Test in the starter covering the integration aspects. Could be added [here](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeIT.java).

PR #3209, opened 2020-03-30
Set application version when deployment exists already

erdemedeiros:
Set application version when deployment exists already


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3211

PR #3226, opened 2020-04-08
fix missed database version: 6.0.0.4

xukeek:
Caused by: org.activiti.engine.ActivitiException: Could not update Activiti database schema: unknown version from database: '6.0.0.4'


balsarori:
Thanks for the PR. `6.0.1.0` should also be added.

balsarori:
And seems an upgrade step from 6 to 7 is also missing

Issue #3152, opened 2020-02-17
Update to JUnit 5


mteodori:
See: https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4

PR #3239, opened 2020-04-20
upgrade to junit 5 and assertj as in spring boot 2.2

mteodori:


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3152

PR #3277, opened 2020-05-07
Support variables on start created process

VitoAlbano:


activiti-api/activiti-api-process-model/src/main/java/org/activiti/api/process/model/builders/CreateProcessPayloadBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+public class CreateProcessPayloadBuilder {
+
+    private String processDefinitionId;
+    private String processDefinitionKey;
+    private String name;
+
+    public CreateProcessPayloadBuilder() {
+    }
+
+    public CreateProcessPayloadBuilder(String processDefinitionId, String processDefinitionKey, String name) {
```
erdemedeiros, line 14: This constructor seems never used. Let's remove it in this case.

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3278.

Issue #3215, opened 2020-03-31
Add support for variable type BigDecimal


erdemedeiros:
Make sure the engine is able to manipulate variables of type BigDecimals.

- [x] Create and retrieve variables of type BigDecimal via Java API
- [x] Create and retrieve variables of type BigDecimal via REST API
- [x] Transfer variables of type BigDecimal to/from cloud connectors

PR #3287, opened 2020-05-14
fix(AAE-493): add support to convert Java scalar, list and set types. 

igdianov:


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3215

Issue #3302, opened 2020-05-21
There is a problem with the ensureRootProcessInstanceInitialized method


CalvinY058:
There is a problem with this method.
org.activiti.engine.impl.persistence.entity.ExecutionEntityImpl#ensureRootProcessInstanceInitialized

PR #3303, opened 2020-05-22
#3302 Fixd

CalvinY058:
fix the bug


mteodori:
we should reopen as this change should go to develop, on master we keep the last tag only

Issue #3309, opened 2020-06-01
Query service should support filtering tasks on variable name and value


erdemedeiros:
As a user, I want to perform a task using a specific variable name and value. Later on, as a user, I want to query my list of the completed task and I want to filter them based on these variable names and values.

Use case:
We have different categories, such as “Approved” and “Rejected” on our Task Manager application, which is supposed to show only the relevant, completed items, we need to filter on the task variable level for this.

Example:
`GET v1/tasks?variables.name=outcome&variables.value=Approved`
It's also possible to specify the type to be used by `variables.value` by specifying `variables.type` (default is String):
`GET v1/tasks?variables.name=approved&variables.value=true&variables.type=boolean`

PR #3310, opened 2020-06-01
Allow passing default type as parameter to ProcessVariablesMapTypeRegistry

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3309


igdianov (approved):
lgtm

PR #3318, opened 2020-06-08
fix(AAE-2825): New serializer for IntegrationRequests produces wrong results #3316

### LINKED ISSUE #3316: New serializer for IntegrationRequests produces wrong results
In some cases, the new [ProcessVariablesMapSerializer](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapSerializer.java) produces inconsitent values that [ProcessVariablesMapDeserializer](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapDeserializer.java) is not able to deserialize. For instance, before adding support to [LocalDateTime](https://github.com/Activiti/Activiti/pull/3315), it was producing the following json: 
```json
{
  "id": "3e718a62-ce9d-4f6b-b7e4-ad0ff114d41f",
  "inBoundVariables": {},
  "outBoundVariables": {
    "variable": {
      "type": "map",
      "value": "05/06/2020, 15:02"
    }
  }
}
```

Note that `type` is defined as `map`, but the `value` is not a map.

igdianov:
Fixes #3316

This PR applies object type for serializing/deserializing integration request process variables that do not have registered type. 

The deserializer will fall back to regular key/value map in order to maintain backward compatibility with previous versions of Activiti Cloud integration context serialization using standard object mapper.

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2010-2020 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.api.runtime.model.impl;
+
+import java.util.Map;
+
+import org.springframework.core.convert.converter.Converter;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@ProcessVariableTypeConverter
+public class ObjectValueToStringConverter implements Converter<ObjectValue, String> {
+    private final ObjectMapper objectMapper;
+
+    public ObjectValueToStringConverter(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public String convert(ObjectValue source) {
+
+        try {
+            Map<String, Object> value = objectMapper.convertValue(source, Map.class);
+
+            if (Map.class.isInstance(value.get("object"))) {
+                Map<String, Object> object = objectMapper.convertValue(source.getObject(), Map.class);
+
+                if (object.containsKey("@class")) {
+                    Map.class.cast(value.get("object")).put("@class", object.get("@class"));
+                }
+            }
+
+            return objectMapper.writeValueAsString(value);
+        } catch (Exception cause) {
+            throw new RuntimeException(cause);
+        }
+    }
+}
--- a/activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
@@ -23,23 +23,26 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 @ProcessVariableTypeConverter
 public class ObjectValueToStringConverter implements Converter<ObjectValue, String> {
+    private static final String CLASS = "@class";
+    private static final String OBJECT = "object";
     private final ObjectMapper objectMapper;
 
     public ObjectValueToStringConverter(ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public String convert(ObjectValue source) {
 
         try {
             Map<String, Object> value = objectMapper.convertValue(source, Map.class);
 
-            if (Map.class.isInstance(value.get("object"))) {
+            if (Map.class.isInstance(value.get(OBJECT))) {
                 Map<String, Object> object = objectMapper.convertValue(source.getObject(), Map.class);
 
-                if (object.containsKey("@class")) {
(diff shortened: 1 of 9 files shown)


activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
```diff
@@ -0,0 +1,51 @@
...
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public String convert(ObjectValue source) {
+
+        try {
+            Map<String, Object> value = objectMapper.convertValue(source, Map.class);
+
+            if (Map.class.isInstance(value.get("object"))) {
```
erdemedeiros: It's probably better to extract a constant for `"object"`.

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
```diff
@@ -0,0 +1,51 @@
...
+                Map<String, Object> object = objectMapper.convertValue(source.getObject(), Map.class);
+
+                if (object.containsKey("@class")) {
```
erdemedeiros: Same for`@class`.

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapDeserializer.java
```diff
@@ -45,13 +46,31 @@ public ProcessVariablesMapDeserializer(ConversionService conversionService) {
...
+                        result = ObjectValue.class.cast(result)
+                                                  .getObject();
+                    }
+
+                    map.put(name, result);
+                }
+                else {
+                    Object value = null;
+                    try {
+                        value = new ObjectMapper().treeToValue(entryValue,
```
erdemedeiros: Shaw, we try to inject the `ObjectMapper` instead of creating a new one?

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapDeserializer.java
```diff
@@ -45,13 +46,31 @@ public ProcessVariablesMapDeserializer(ConversionService conversionService) {
...
+                                                               Object.class);
+                    } catch (JsonProcessingException e) {
+                        e.printStackTrace();
```
erdemedeiros: Please, use logger instead of `printStackTrace()`.

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapSerializer.java
```diff
@@ -61,17 +59,20 @@ public void serialize(ProcessVariablesMap<String, Object> processVariablesMap,
...
         throws JsonProcessingException {
         ProcessVariableValue variableValue = null;
+
         if (value != null) {
             Class<?> entryValueClass = value.getClass();
+            String entryType = resolveEntryType(entryValueClass, value);
 
-            if (!conversionService.canConvert(entryValueClass, String.class)) {
-                value = objectMapper.writeValueAsString(value);
+            if("object".equals(entryType)) {
```
erdemedeiros: Re-use constant.

erdemedeiros (changes requested):
Some small details.

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapDeserializer.java
igdianov: @erdemedeiros I am afraid to do it because it may bring side effects. We should only consider to inject global object mapper to support some test case, I think.

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapDeserializer.java
igdianov: done.

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapSerializer.java
igdianov: Done

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
igdianov: Done

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ObjectValueToStringConverter.java
igdianov: Done

activiti-core/activiti-api-impl/activiti-api-process-model-impl/src/main/java/org/activiti/api/runtime/model/impl/ProcessVariablesMapDeserializer.java
erdemedeiros: Ok. Let's keep a new instance of ObjectMapper in this case.

erdemedeiros (approved):

Issue #3356, opened 2020-06-25
Variable mapping 'pass all variables' use case does not allow multiple elements


miguelruizdev:
**Problem:**
The current mapping type 'Send all variables', represented by:
```
{
  "mappings": { }
}
```
does not allow for multiple tasks having different mappings.
In such a way, if a task sets this mapping type, all tasks have to have it, preventing the possibility of setting different mapping types for each task in the same process.

**Solution:**
Change the representation of this type of mapping to specifically reference the task that has it:
```
{
  "mappings": { 
      "taskId": { }
}
```

PR #3357, opened 2020-06-25
Update pass all variables use case

miguelruizdev:
Addressing this issue: #3356

### DIFF
--- a/activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/ProcessVariablesMapping.java
+++ b/activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/ProcessVariablesMapping.java
@@ -17,13 +17,12 @@
 
 package org.activiti.spring.process.model;
 
-import java.util.HashMap;
 import java.util.Map;
 
 public class ProcessVariablesMapping {
 
-    private Map<String, Mapping> inputs = new HashMap<>();
-    private Map<String, Mapping> outputs = new HashMap<>();
+    private Map<String, Mapping> inputs;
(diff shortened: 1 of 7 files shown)


activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/ProcessVariablesMapping.java
```diff
@@ -17,13 +17,12 @@
...
 package org.activiti.spring.process.model;
 
-import java.util.HashMap;
 import java.util.Map;
 
 public class ProcessVariablesMapping {
 
-    private Map<String, Mapping> inputs = new HashMap<>();
-    private Map<String, Mapping> outputs = new HashMap<>();
+    private Map<String, Mapping> inputs;
```
miguelruizdev, line 24: if we initialize these maps by default, this use case:
```
"mappings": {
"UserTask_1negxpm": { }
}
```
and this one
```
       "mappings": {
           "UserTask_1negxpm": {
               "inputs": { },
               "outputs": { }
           }
       }
```
are interpreted in the same way

erdemedeiros (changes requested):
As discussed with @miguelruizdev the initial assumption is wrong: `Send all` does not mean having an empty mapping section, it means having no entry in the mapping for the given task id. Which means that the following map will filter variables sent to `task1`, but will `Send all` for `task2` or any other task that is not referenced in the mapping:

```
"mappings": {
        "task1": {
          "inputs": {
            "taskVar": {
              "type": "variable",
              "value": "name"
            }
          }
}
```

PR #3359, opened 2020-06-28
Duplicate property exist.

silverskyvicto:
Hi.

There were two properties with the same content, so I deleted one.

How about this?


mteodori:

PR #3361, opened 2020-06-28
Conquered because activiti.version was still old.

silverskyvicto:
Hi.

The activiti.version in build.xml was still old, so I fixed it.

How about this?


mteodori:

Issue #3076, opened 2020-01-20
CommandContextCloseListener.closed called inside transaction


ggrnd0:
https://github.com/Activiti/Activiti/blob/8c3db4b575581aa79296a8d7f027e8b2085bb23e/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java#L965-L983
Now CommandContextInterceptor and CommandContextCloseListener.closed run inside TransactionInterceptor (SpringTransactionInterceptor in my case)

Is it correct to run AsyncJobAddedNotification.closed() inside transaction?

ggrnd0:
Ok, chain order is good. Transaction must be open while context closing...

But AsyncJobAddedNotification.closed() really must not be executed inside transaction...

erdemedeiros:
@ggrnd0, I was looking at this issue and [AsyncJobAddedNotification](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/jobexecutor/AsyncJobAddedNotification.java#L48) is using `TransactionPropagation.REQUIRES_NEW` which means that it should be using a new transaction to execute its unit of work. Could you, please, let us know which is the misbehaviour that you're noticing?

ghost:
@erdemedeiros, in cases when *current* transaction is not commit-ish enough concurrent transaction will fail

erdemedeiros:
@ggrnd0, do you have a reproducer? I guess that if you opened this issue it's because you faced an issue in one of your use cases. With a reproducer, it would easier to investigate and prove that the problem has been solved.

ghost:
no. but we have about 5 issues with stucked tasks per year.
they  successfully unstuck after lock expired

erdemedeiros:
@ggrnd0, this PR should fix the issue for `AsyncJobAddedNotification`:  https://github.com/Activiti/Activiti/pull/3367. Thank you for pointing it out!

PR #3367, opened 2020-07-07
fix(AAE-2166): CommandContext close listener being called inside the transaction

igdianov:
This PR investigated close listener for async job notifications via CommandContextListener.close method. Using CommandContext.close method may not work 100% for async executions since it schedules a task in new transaction to execute async continuation behavior while parent transaction is still active. There is a  chance for race conditions because async job transaction starts in a different thread while parent spring transaction may not be committed yet. This problem is hard to reproduce because thread scheduling is not deterministic.

The proposed async job notification implementation should use different approach to register transaction listener instead of command context listener, so that job always scheduled after parent transaction committed.


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3076

erdemedeiros (approved):
Thanks, @igdianov. LGTM: this should fix the issue for `AsyncJobAddedNotification`.

PR #3374, opened 2020-07-16
context path do not end with a '/'

### LINKED ISSUE #3373: The context path must start with a blank or'/' and end with a character other than'/'.
Hi.

It seems that the context path must start with a blank or slash and end with a non-slash character.
This seems to be due to the Servlet specifications.

- [JSR 315: JavaTM Servlet 3.0 Specification](https://jcp.org/en/jsr/detail?id=315)

https://github.com/Activiti/Activiti/blob/01813c105679e8d9a6587fd9f4c40428b73ff691/modules/activiti-ui/activiti-app/src/main/webapp/META-INF/context.xml#L21

silverskyvicto:
I fix #3373 . How about this?


mteodori:

PR #3375, opened 2020-07-16
Corresponding because there was a part where logback was not set

silverskyvicto:
Hi

There was a part where logback was not set, so I added it.

How about that?


mteodori:

PR #3378, opened 2020-07-16
Fixed deprecated methods in Spring Framework Http

silverskyvicto:
Hi.

Fixed some methods that were deprecated in Spring Framework Http.

How about that?


mteodori:

Issue #3385, opened 2020-07-20
Multi instance Result Collection should allow to collect all the output variables


erdemedeiros:
As a modeler, I should be able to collect all the output variables into Result Collection.
Currently, you can only select one output variable.

Here is the current JSON in REsult COllection
```json
[
  "Accept",
  "Change",
  "Accept",
  "Accept",
  "Accept",
  "Accept"
]
```
What is expected is all the variables in the array:
```json
[
  {
   "var1":"valueA1",
   "var2":"valueA2",
   "var3":"valueA3"
  },
 {
   "var1":"valueB1",
   "var2":"valueB2",
   "var3":"valueB3"
  }, 
  {
   "var1":"valueC1",
   "var2":"valueC2",
   "var3":"valueC3"
  }
]
```

PR #3386, opened 2020-07-20
Extract interface for variables mapping

erdemedeiros:
- bring the new interface (VariablesCalculator) closer to activiti core


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java
```diff
@@ -449,11 +452,21 @@ protected void updateResultCollection(DelegateExecution childExecution,
...
-            resultCollection.add(childExecution.getVariable(getOutputDataItem()));
+            resultCollection.add(getResultElementItem(childExecution));
             setLoopVariable(miRootExecution, getLoopDataOutputRef(), resultCollection);
         }
     }
 
+    private Object getResultElementItem(DelegateExecution childExecution) {
+        CommandContext commandContext = Context.getCommandContext();
+        if (commandContext.getCommand() instanceof CompleteTaskCmd) {
+            Map<String, Object> taskVariables = ((CompleteTaskCmd) commandContext
```
miguelruizdev, line 467: Why only for `CompleteTaskCmd` ?

miguelruizdev (approved):

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java
erdemedeiros: Good point! Actually, user tasks follow a different path because the variables are attached directly to the task entity (which is deleted during the execution of CompleteTaskCmd). For the other cases, the variables will be available in the execution. I preferred to not change the current mechanism that we have for variables on user tasks as it can cause some unwanted behavior.

erdemedeiros: I added a comment explaining this.

Issue #3368, opened 2020-07-08
Variable mapping representation changes


miguelruizdev:
**Don't send variables**

This will become the default mapping case, happening implicitly when there is no mapping entry for a given element within the "mappings" section of the extensions.json file.
No variables are going to be passed as inputs or outputs to any given element.

**Map variables**

The standard mapping case, where inputs and output variables are mapped.
It will stay unchanged

**Send all variables**

For each mapping entry with this type, in order to guarantee the independence between inputs and outputs, there will be a field stating if the 'Send all variables' affects inputs (**MAP_ALL_INPUTS**), outputs (**MAP_ALL_OUTPUTS**), or both (**MAP_ALL**).

Extensions representation:
```
"mappings": {
    "task_id": {
        "mappingType": "MAP_ALL"
    }
}
```
What we achieve:

- A more clear and intuitive representation for the person modeling the process.
- Independence of mapping type between inputs and outputs of the same element

PR #3392, opened 2020-07-24
Variable mapping representation changes

miguelruizdev:
Addressing: #3368

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
@@ -77,7 +77,7 @@ public class ExtensionsVariablesMappingProvider implements VariablesCalculator {
 
         Map<String, Object> inboundVariables;
 
-        if (extensions.hasMappingType(execution.getCurrentActivityId()) && extensions.hasMappingTypeInputs(execution.getCurrentActivityId())) {
+        if (extensions.hasMappingTypeInputs(execution.getCurrentActivityId())) {
             inboundVariables = execution.getVariables();
         } else {
             inboundVariables = calculateInputVariables(execution, extensions);
@@ -140,8 +140,7 @@ public class ExtensionsVariablesMappingProvider implements VariablesCalculator {
             return emptyMap();
         }
 
-        if (extensions.hasMappingType(mappingExecutionContext.getActivityId()) &&
-            extensions.hasMappingTypeOutputs(mappingExecutionContext.getActivityId())) {
+        if (extensions.hasMappingTypeOutputs(mappingExecutionContext.getActivityId())) {
             return (availableVariables != null ? new HashMap<>(availableVariables) : emptyMap());
         }
 
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
@@ -77,7 +77,7 @@ public class ExtensionsVariablesMappingProvider implements VariablesCalculator {
 
         Map<String, Object> inboundVariables;
 
-        if (extensions.hasMappingTypeInputs(execution.getCurrentActivityId())) {
+        if (extensions.shouldMapAllInputs(execution.getCurrentActivityId())) {
             inboundVariables = execution.getVariables();
         } else {
             inboundVariables = calculateInputVariables(execution, extensions);
@@ -140,7 +140,7 @@ public class ExtensionsVariablesMappingProvider implements VariablesCalculator {
             return emptyMap();
         }
 
-        if (extensions.hasMappingTypeOutputs(mappingExecutionContext.getActivityId())) {
+        if (extensions.shouldMapAllOutputs(mappingExecutionContext.getActivityId())) {
             return (availableVariables != null ? new HashMap<>(availableVariables) : emptyMap());
         }
 
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
@@ -71,13 +71,13 @@ public class ExtensionsVariablesMappingProvider implements VariablesCalculator {
 
         Map<String, Object> constants = calculateConstants(execution, extensions);
 
-        if (extensions.hasEmptyInputsMapping(execution.getCurrentActivityId())) {
+        if (!extensions.hasMapping(execution.getCurrentActivityId())) {
             return constants;
         }
 
         Map<String, Object> inboundVariables;
 
-        if (!extensions.hasMapping(execution.getCurrentActivityId())) {
+        if (extensions.hasMappingType(execution.getCurrentActivityId()) && extensions.hasMappingTypeInputs(execution.getCurrentActivityId())) {
             inboundVariables = execution.getVariables();
         } else {
             inboundVariables = calculateInputVariables(execution, extensions);
@@ -136,11 +136,12 @@ public class ExtensionsVariablesMappingProvider implements VariablesCalculator {
 
         Extension extensions = processExtensionService.getExtensionsForId(mappingExecutionContext.getProcessDefinitionId());
 
-        if (extensions.hasEmptyOutputsMapping(mappingExecutionContext.getActivityId())) {
+        if (!extensions.hasMapping(mappingExecutionContext.getActivityId())) {
             return emptyMap();
         }
 
-        if (!extensions.hasMapping(mappingExecutionContext.getActivityId())) {
+        if (extensions.hasMappingType(mappingExecutionContext.getActivityId()) &&
+            extensions.hasMappingTypeOutputs(mappingExecutionContext.getActivityId())) {
             return (availableVariables != null ? new HashMap<>(availableVariables) : emptyMap());
         }
 
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
@@ -88,13 +88,12 @@ public class ProcessVariablesInitiator extends ProcessInstanceHelper {
 
         if (processExtensionService.hasExtensionsFor(processDefinition)) {
 
-            processVariables = variablesCalculator.calculateOutPutVariables(MappingExecutionContext.buildMappingExecutionContext(
+            processVariables.putAll(variablesCalculator.calculateOutPutVariables(MappingExecutionContext.buildMappingExecutionContext(
                 processDefinition.getId(),
                 initialFlowElement.getId()),
-                variables);
+                variables));
 
-            processVariables = calculateVariablesFromExtensionFile(processDefinition,
-                processVariables);
+            processVariables.putAll(calculateVariablesFromExtensionFile(processDefinition, processVariables));
         }
         return processVariables;
     }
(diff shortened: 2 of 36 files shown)


activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/spring/process/ProcessVariablesInitiator.java
```diff
@@ -88,13 +88,12 @@ public ProcessVariablesInitiator(ProcessExtensionService processExtensionService
 
         if (processExtensionService.hasExtensionsFor(processDefinition)) {
 
-            processVariables = variablesCalculator.calculateOutPutVariables(MappingExecutionContext.buildMappingExecutionContext(
+            processVariables.putAll(variablesCalculator.calculateOutPutVariables(MappingExecutionContext.buildMappingExecutionContext(
```
miguelruizdev: Is there any reason why passed vars at creation of the process and vars from extensions are mutually exclusive?

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
```diff
@@ -71,13 +71,13 @@ public ExtensionsVariablesMappingProvider(ProcessExtensionService processExtensi
...
 
-        if (extensions.hasEmptyInputsMapping(execution.getCurrentActivityId())) {
+        if (!extensions.hasMapping(execution.getCurrentActivityId())) {
             return constants;
         }
 
         Map<String, Object> inboundVariables;
 
-        if (!extensions.hasMapping(execution.getCurrentActivityId())) {
+        if (extensions.hasMappingTypeInputs(execution.getCurrentActivityId())) {
```
erdemedeiros: I'm a little bit confused here. Possibly we need to rename this method. Maybe something like `shouldMapAllInputs`. What do you think?

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
```diff
@@ -136,11 +136,11 @@ public ExtensionsVariablesMappingProvider(ProcessExtensionService processExtensi
 
         Extension extensions = processExtensionService.getExtensionsForId(mappingExecutionContext.getProcessDefinitionId());
 
-        if (extensions.hasEmptyOutputsMapping(mappingExecutionContext.getActivityId())) {
+        if (!extensions.hasMapping(mappingExecutionContext.getActivityId())) {
             return emptyMap();
         }
 
-        if (!extensions.hasMapping(mappingExecutionContext.getActivityId())) {
+        if (extensions.hasMappingTypeOutputs(mappingExecutionContext.getActivityId())) {
```
erdemedeiros: In a similar way, I'd rename it to `shouldMapAllOutputs`.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeVariableMappingIT.java
```diff
@@ -149,8 +155,59 @@ public void should_passTaskOutputVariables_when_usingVariableMapping() {
     }
 
     @Test
-    public void should_passAllVariables_when_thereIsNoMapping() {
-        ProcessInstance processInstance = processBaseRuntime.startProcessWithProcessDefinitionKey(TASK_VAR_NO_MAPPING);
+    public void should_notMapTaskVariables_when_thereIsNoMapping() {
+        ProcessInstance processInstance = processBaseRuntime.startProcessWithProcessDefinitionKey(TASK_NO_MAPPING);
+
+        List<Task> tasks = taskBaseRuntime.getTasksByProcessInstanceId(processInstance.getId());
+        assertThat(tasks).isNotEmpty();
```
erdemedeiros: Tip: you don't need this check as you have already `hasSize(1)` that will complain if it's empty.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeVariableMappingIT.java
```diff
@@ -149,8 +155,59 @@ public void should_passTaskOutputVariables_when_usingVariableMapping() {
...
+            .extracting(VariableInstance::getName,
+                VariableInstance::getValue)
+            .containsOnly(tuple("process_variable_unmapped_1",
+                "unmapped1Value"),
+                tuple("process_variable_inputmap_1",
+                    "inputmap1Value"),
+                tuple("process_variable_outputmap_1",
+                    "outputmap1Value"));
+
+        processBaseRuntime.delete(processInstance.getId());
```
erdemedeiros: No need to clean it up: there is already a `@AfterEach` method taking care of it.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeVariableMappingIT.java
```diff
@@ -211,16 +268,70 @@ public void should_passAllVariables_when_thereIsNoMapping() {
         processBaseRuntime.delete(processInstance.getId());
     }
 
-    private Task checkTasks(String processInstanceId) {
-        List<Task> tasks = taskBaseRuntime.getTasksByProcessInstanceId(processInstanceId);
-        assertThat(tasks).isNotEmpty();
-        assertThat(tasks).hasSize(1);
-        return tasks.get(0);
+    @Test
+    public void should_sendOnlyInputTaskVariables_when_thereIsMappingTypeMAP_ALL_INPUTS() {
```
erdemedeiros: Suggestion: `should_sendAllInputTasksVariablesAndOnlySelectedOuputs_when_thereIsMappingTypeMAP_ALL_INPUTS`

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeVariableMappingIT.java
```diff
@@ -211,16 +268,70 @@ public void should_passAllVariables_when_thereIsNoMapping() {
...
+                    "outputMappedValue"));
+
+        processBaseRuntime.delete(processInstance.getId());
+
     }
 
     @Test
-    public void should_notHaveTaskVariables_when_mappingIsEmpty() {
-        ProcessInstance processInstance = processBaseRuntime.startProcessWithProcessDefinitionKey(TASK_EMPTY_VAR_MAPPING);
+    public void should_sendOnlyOutputTaskVariables_when_thereIsMappingTypeMAP_ALL_OUTPUTS() {
```
erdemedeiros: Suggestion:
`should_sendAllOutputTaskVariablesAndOnlySelectedInputs_when_thereIsMappingTypeMAP_ALL_OUTPUTS`

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
```diff
@@ -82,18 +83,22 @@ public VariableDefinition getPropertyByName(String name) {
...
 
-    public boolean hasEmptyInputsMapping(String elementId) {
-        ProcessVariablesMapping processVariablesMapping = mappings.get(elementId);
-        return processVariablesMapping != null && processVariablesMapping.getInputs().size() == 0;
+    public boolean hasMapping(String taskId) {
+        return mappings.get(taskId) != null;
     }
 
-    public boolean hasEmptyOutputsMapping(String elementId) {
+    public boolean hasMappingTypeInputs (String elementId) {
```
erdemedeiros: I'd rename it to `shouldMapAllInputs`.

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
```diff
@@ -82,18 +83,22 @@ public VariableDefinition getPropertyByName(String name) {
...
         ProcessVariablesMapping processVariablesMapping = mappings.get(elementId);
-        return processVariablesMapping != null && processVariablesMapping.getOutputs().size() == 0;
+        return processVariablesMapping.getMappingType() != null &&
+            (processVariablesMapping.getMappingType().equals(MappingType.MAP_ALL_INPUTS) ||
+            processVariablesMapping.getMappingType().equals(MappingType.MAP_ALL));
     }
 
-    public boolean hasMapping(String taskId) {
-        return mappings.get(taskId) != null;
+    public boolean hasMappingTypeOutputs(String elementId) {
```
erdemedeiros: I'd rename it to `shouldMapAllOutputs`.

activiti-core/activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/model/ExtensionTest.java
```diff
@@ -34,33 +35,52 @@
...
 
-        assertThat(extension.hasEmptyOutputsMapping("elementId")).isTrue();
+        assertThat(extension.hasMappingTypeInputs("elementId")).isFalse();
+        assertThat(extension.hasMappingTypeOutputs("elementId")).isTrue();
+
     }
 
     @Test
-    public void hasMappingShouldReturnTrueWhenThereIsMapping() {
+    public void should_hasMappingReturnTrue_when_thereIsMapping() {
```
erdemedeiros, line 82: Personally, I prefer `hasMapping_should_returnTrue_when_thereIsMapping`: it's more fluent and closer to the speaking language.

erdemedeiros (changes requested):
Good job @miguelruizdev! Here some concerns.
In addition, I think it's better to add a test covering the mapping of the built-in multi-instance variable `loopCounter`.

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
zoltanpalfi: I wouldn't name something that returns a boolean with `should` prefix, I believe `has`, `is` is more appropriate for those

activiti-core/activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/model/ExtensionTest.java
zoltanpalfi: There is a naming convention for tests that we all agreed to, hence this naming. If you don't like that, let's change that, otherwise we should refrain from individual variants as we end up with methods in 100s of style

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
zoltanpalfi: Again, disagree. For a boolean function, `has`/`is` prefix is a better fit

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
erdemedeiros: I agree, in most of the cases, a method returning a boolean will have the prefix `is` or `has` specially when they are simple getters. However, this is not a simple getter, there is business logic behind it. In this case, the only purpose of having this method is to check whether we should or should not map all the inputs. In this way, and for the sake of the code readability I think it makes totally sense to use `should` prefix. Questions starting by `Should` can be answered by `yes/no`, `true`/`false`. So I don't see why `should` is less appropriate than `is` or `has` if it helps to improve the code readability.
Does it make sense?

activiti-core/activiti-spring-process-extensions/src/test/java/org/activiti/spring/process/model/ExtensionTest.java
erdemedeiros: Ok. It makes sense. Let's review the naming convention for tests to clarify some specific points.

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
erdemedeiros: Same as [above](https://github.com/Activiti/Activiti/pull/3392#discussion_r476305360).

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/model/Extension.java
zoltanpalfi: Yes, now makes sense. It's difficult now to change, we should avoid to have something like this if possible as makes it hard to understand what's going on

zoltanpalfi: let's rename it and also remove that tailing space as well please @miguelruizdev

miguelruizdev:
> Good job @miguelruizdev! Here some concerns.
> In addition, I think it's better to add a test covering the mapping of the built-in multi-instance variable `loopCounter`.

Opened an issue to check multi-instance mapping as a separate PR: https://github.com/Activiti/Activiti/issues/3430

erdemedeiros (approved):

PR #3401, opened 2020-08-10
Fix the propagation of variable created event

erdemedeiros:
Only publish events when it's a process variable or a task variable. This excludes other execution variables as we are not exposing the notion of execution in the new APIs and in audit and query service.

Ref https://github.com/Activiti/Activiti/pull/3393


eromano (approved):

erdemedeiros:

PR #3402, opened 2020-08-10
Fix variable filter for standalone task

erdemedeiros:
The executionId will not be set in this case

Ref https://github.com/Activiti/Activiti/pull/3393


balsarori (approved):

erdemedeiros:

Issue #3398, opened 2020-08-05
Filter tasks by processDefinitionName


Giovanni007:
As user I would like to filter the task list by processDefinitonName using the API

query/v1/tasks?processDefinitionName={processDefinitionName}

Right now the processDefinitionName is not present in the response of the service. It needs to be added in order to allow the user to filer by processDefinitionName.

PR #3403, opened 2020-08-12
  Activiti/Activiti#3398 - added processDefinitionName to Task

Giovanni007:
Added processDefinitionNama to Task interface. 
As described in the issue [#3398](https://github.com/Activiti/Activiti/issues/3398) we need to return this value in the task api


mauriziovitale (commented):
can we add a unit test?

Issue #3418, opened 2020-08-27
Synchronize feign clients of Runtime bundle with Controllers


miguelruizdev:
The current architecture present in several services, where an interface is implemented by a controller, can be leveraged to build a feign client that can be used in external services as well as in the core acceptance test layer, avoiding the usage of manually created feign clients and guaranteeing a synchronization between controllers and clients.

PR #3428, opened 2020-09-03
Activiti/Activiti#3418 Add setter methods to Payloads

miguelruizdev:
Addressing part of: Activiti/Activiti#3418


miguelruizdev:
Not needed, better to use the payload builders instead

Issue #3439, opened 2020-09-14
Allow end-user to reassign a task


Giovanni007:
Allow an end-user to reassign a claimed task to another user that is a candidate. 
This  functionality is already used by the admin ([TaskAdminRuntime](https://github.com/Activiti/Activiti/blob/eeda8439bcd30f525f10b23f54a7cbb4819d2ef9/activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskAdminRuntime.java#L101)). To allow end-users to do the same we need a to implement a new `assign` method for [TaskRuntime](https://github.com/Activiti/Activiti/blob/eeda8439bcd30f525f10b23f54a7cbb4819d2ef9/activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java)
User must be the owner or the current assignee to assign the task to another user.
User can assign a task only to candidate users.

PR #3440, opened 2020-09-14
Activiti/Activiti#3439 - Added assign method for TaskRuntime

Giovanni007:
Ref https://github.com/Activiti/Activiti/issues/3439

### DIFF
--- a/activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java
+++ b/activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java
@@ -21,6 +21,7 @@ import org.activiti.api.model.shared.model.VariableInstance;
 import org.activiti.api.runtime.shared.query.Page;
 import org.activiti.api.runtime.shared.query.Pageable;
 import org.activiti.api.task.model.Task;
+import org.activiti.api.task.model.payloads.AssignTaskPayload;
 import org.activiti.api.task.model.payloads.CandidateGroupsPayload;
 import org.activiti.api.task.model.payloads.CandidateUsersPayload;
 import org.activiti.api.task.model.payloads.ClaimTaskPayload;
@@ -133,4 +134,12 @@ public interface TaskRuntime {
     
     List<String> userCandidates(String taskId);
     List<String> groupCandidates(String taskId);
+
+    /**
+     * Assign a task with a new user
+     *  - If there is a user assigned, reassign it to the new user
+     *  - user should be a candidate-user for the task
+     *  - after the assign the task should be in assigned status
+     */
+    Task assign(AssignTaskPayload assignTaskPayload);
 }
--- a/activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java
+++ b/activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java
@@ -136,10 +136,9 @@ public interface TaskRuntime {
     List<String> groupCandidates(String taskId);
 
     /**
-     * Assign a task with a new user
-     *  - If there is a user assigned, reassign it to the new user
-     *  - user should be a candidate-user for the task
-     *  - after the assign the task should be in assigned status
+     * Assign a task that has been claimed before to a different user
+     * - Only the current assignee can perform the action
+     * - The new assignee should be part of the candidate users for this task
      */
     Task assign(AssignTaskPayload assignTaskPayload);
 }
--- a/activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
+++ b/activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
@@ -473,12 +473,8 @@ public class TaskRuntimeImpl implements TaskRuntime {
     }
     
     private void assertCanReleaseTask(String taskId) {
-        Task task;
-        try {
-            task = task(taskId);
-        } catch (IllegalStateException ex) {
-            throw new IllegalStateException("The authenticated user cannot release task" + taskId + " due it is not a candidate for it");
-        }
+        Task task = task(taskId);
+        
         if (task.getAssignee() == null || task.getAssignee().isEmpty()) {
             throw new IllegalStateException("You cannot release a task that is not claimed");
         }
--- a/activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
+++ b/activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
@@ -47,7 +47,6 @@ import org.activiti.runtime.api.model.impl.APITaskConverter;
 import org.activiti.runtime.api.model.impl.APIVariableInstanceConverter;
 import org.activiti.runtime.api.query.impl.PageImpl;
 import org.springframework.security.access.prepost.PreAuthorize;
-import org.activiti.api.runtime.shared.NotFoundException;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -195,25 +194,9 @@ public class TaskRuntimeImpl implements TaskRuntime {
 
     @Override
     public Task release(ReleaseTaskPayload releaseTaskPayload) {
-        // Validate that the task is visible by the currently authorized user
-        Task task;
-        try {
-            task = task(releaseTaskPayload.getTaskId());
-        } catch (IllegalStateException ex) {
-            throw new IllegalStateException("The authenticated user cannot release task" + releaseTaskPayload.getTaskId() + " due it is not a candidate for it");
-        }
-        // validate the task doesn't have an assignee
-        if (task.getAssignee() == null || task.getAssignee().isEmpty()) {
-            throw new IllegalStateException("You cannot release a task that is not claimed");
-        }
-        String authenticatedUserId = securityManager.getAuthenticatedUserId();
-        // validate that you are trying to release task where you are the assignee
-        if (!task.getAssignee().equals(authenticatedUserId)) {
-            throw new IllegalStateException("You cannot release a task where you are not the assignee");
-        }
-
-        taskService.unclaim(releaseTaskPayload.getTaskId());
-        return task(releaseTaskPayload.getTaskId());
+        String taskId =  releaseTaskPayload.getTaskId();
+        releaseTask(taskId);
+        return task(taskId);
     }
 
     @Override
@@ -452,14 +435,8 @@ public class TaskRuntimeImpl implements TaskRuntime {
         String assignee = assignTaskPayload.getAssignee();
         String taskId = assignTaskPayload.getTaskId();
 
-        checkUserHasAccessToTask(taskId);
-        checkAssigneeIsACandidateUser(taskId, assignee);
-        
-        //We need to release, claim for assigned task is not working!
-        taskService.unclaim(taskId);
-        //Now assign a new user
-        taskService.claim(taskId, assignee);
-
+        assertAssigneeIsACandidateUser(taskId, assignee);
+        reassignTask(taskId, assignee);
         return reassignedTask(taskId);
     }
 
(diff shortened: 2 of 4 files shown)


activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -89,6 +91,10 @@ public TaskRuntimeConfiguration configuration() {
     public Task task(String taskId) {
         return taskConverter.fromWithCandidates(taskRuntimeHelper.getInternalTaskWithChecks(taskId));
     }
+    
+    private Task reassignedTask(String taskId) {
+        return taskConverter.fromWithCandidates(taskRuntimeHelper.getInternalTask(taskId));
+    }
```
Giovanni007, line 96: When the user reassign a task, he is not anymore the assignee, for this reason the method `getInternalTaskWithChecks` fails to find the task. I used `getInternalTask` as workaround but another solution (?) could be change the implementation of [getInternalTaskWithChecks](https://github.com/Activiti/Activiti/blob/eeda8439bcd30f525f10b23f54a7cbb4819d2ef9/activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeHelper.java#L190) with less restrictive checks.

miguelruizdev (approved):
LGTM 👍

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -441,6 +447,22 @@ public void save(SaveTaskPayload saveTaskPayload) {
...
 
+    @Override
+    public Task assign(AssignTaskPayload assignTaskPayload) {
+        String assignee = assignTaskPayload.getAssignee();
+        String taskId = assignTaskPayload.getTaskId();
+
+        checkUserHasAccessToTask(taskId);
+        checkAssigneeIsACandidateUser(taskId, assignee);
+        
+        //We need to release, claim for assigned task is not working!
```
mauriziovitale: do we need these comments ?

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -441,6 +447,22 @@ public void save(SaveTaskPayload saveTaskPayload) {
...
+        taskService.unclaim(taskId);
+        //Now assign a new user
```
mauriziovitale: in my opinion, it would be better to have a releaseAndClaim method and remove the comments

mauriziovitale (changes requested):
check the comment

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
erdemedeiros: The current assumption from the APIs is that only the assignee, candidates, or owner can see the task. A possible solution would be programmatically adding the user who is releasing the task as a candidate for it. What it makes sense IMO, we can say if a user as the assignee of this task at some point, he remains a candidate for it.

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
erdemedeiros: I agree with @mauriziovitale. It's probably better to create a private method `releaseAndClaim` or `reassign` and remove the comments.

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -455,5 +477,20 @@ public void save(SaveTaskPayload saveTaskPayload) {
...
+        if(!userCandidates.contains(assignee)){
+            throw new IllegalStateException("You cannot assign a task to " + assignee + " due it is not a candidate for it");
+        }
+    }
+    
+    private void checkUserHasAccessToTask(String taskId){
+        try {
+            taskRuntimeHelper.assertHasAccessToTask(taskId);
+        }catch (NotFoundException ex) {
+            throw new IllegalStateException("You cannot update a task where you are not the assignee");
```
erdemedeiros: Note: not only the assignee can see the task. All the candidates and the owner can also see it as well. So either the message is wrong or the method you're calling is not checking what you expected.
Who is supposed to use this new method? Only the assignee or anyone who can see the task?

Also, `assertHasAccessToTask` is throwing `NotFoundException` that you are catching here to throw a different exception. Instead of throw/catch/rethrow it's better to extract another method that is not throwing an exception but it's returning nullable or `Optional` so you can throw directly the exception you want.

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/test/java/org/activiti/runtime/api/impl/TaskRuntimeImplTest.java
```diff
@@ -15,26 +15,40 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
+import org.activiti.runtime.api.model.impl.APIVariableInstanceConverter;
 
 public class TaskRuntimeImplTest {
 
-    @InjectMocks
+    private static final String AUTHENTICATED_USER = "user";
+
```
erdemedeiros, line 48: Have you tried to add the annotation `@Spy` in addition to `@InjectMocks`? If it works it avoids instantiating it manually in the test `setup`, which means one thing less to update if the constructor changes.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeTaskAssigneeTest.java
```diff
@@ -228,5 +229,73 @@ public void createStandaloneTaskForUsersAndAdminReassignUser() {
...
+
+        String taskId = createTask("garth");
+        //assign the new task
+        userAssignTask(taskId,"garth", "dean");
+        //reassign task to candidate
+        userAssignTask(taskId,"dean", "garth");
+
+        test_shouldReturnIllegalStateExceptionWhenUserIsNotAssigneeOrOwner(taskId,"dean", "garth");
+
+        cleanUp(taskId);
```
erdemedeiros: No need to clean up here. It's already done by [the @AfterEach method](https://github.com/Activiti/Activiti/pull/3440/files#diff-9a5a1effbdd6cbccc0d0cca73867d471R51)

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeTaskAssigneeTest.java
```diff
@@ -228,5 +229,73 @@ public void createStandaloneTaskForUsersAndAdminReassignUser() {
...
+        // the owner should be able to see the created task
+        Page<Task> tasks = taskRuntime.tasks(Pageable.of(0,
+                50));
 
+        assertThat(tasks.getContent()).hasSize(1);
+
+        Task task = tasks.getContent().get(0);
+        assertThat(task.getAssignee()).isNull();
+        assertThat(task.getStatus()).isEqualTo(Task.TaskStatus.CREATED);
+        return task.getId();
```
erdemedeiros: I don't think you need these lines here. As I'm expecting this to be already covered by another test. You can return directly the result of `taskRuntime.create`.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/tasks/TaskRuntimeTaskAssigneeTest.java
```diff
@@ -228,5 +229,73 @@ public void createStandaloneTaskForUsersAndAdminReassignUser() {
...
+    @Test
+    public void userCanReassignClaimedTaskToCandidateUsers() {
+
+        String taskId = createTask("garth");
+        //assign the new task
+        userAssignTask(taskId,"garth", "dean");
+        //reassign task to candidate
+        userAssignTask(taskId,"dean", "garth");
+
+        test_shouldReturnIllegalStateExceptionWhenUserIsNotAssigneeOrOwner(taskId,"dean", "garth");
```
erdemedeiros: I'd use a separate test to cover the exception case.

erdemedeiros (changes requested):

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
Giovanni007: As discussed, the API `getInternalTaskWithChecks` works for candidate users only if the current assignee is null [(see query)](https://github.com/Activiti/Activiti/blob/2d94c7d0f3be62a1abef9da8a53f9e7dfa0f0b2c/activiti-core/activiti-engine/src/main/resources/org/activiti/db/mapping/entity/Task.xml#L602). In my case after the assign operation, the assignee is not null and the current user is still a candidate but the API doesn't return the task. Because of that I will use `getInternalTask`.

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
Giovanni007: as discussed, I will not use `assertHasAccessToTask`

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/test/java/org/activiti/runtime/api/impl/TaskRuntimeImplTest.java
Giovanni007: I will use the annotation as suggested, thanks

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
```diff
@@ -455,5 +454,38 @@ public void save(SaveTaskPayload saveTaskPayload) {
...
+        assertCanReleaseTask(taskId);
+        taskService.unclaim(taskId);
+    }
+    
+    private void assertCanReleaseTask(String taskId) {
+        Task task;
+        try {
+            task = task(taskId);
+        } catch (IllegalStateException ex) {
+            throw new IllegalStateException("The authenticated user cannot release task" + taskId + " due it is not a candidate for it");
```
erdemedeiros: IMO, there is no need to catch this exception here to rethrow it with a different message. What do you think?

activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java
```diff
@@ -133,4 +134,12 @@
     
     List<String> userCandidates(String taskId);
     List<String> groupCandidates(String taskId);
+
+    /**
+     * Assign a task with a new user
+     *  - If there is a user assigned, reassign it to the new user
+     *  - user should be a candidate-user for the task
+     *  - after the assign the task should be in assigned status
+     */
```
erdemedeiros: I'd rephrase this a little bit to make it clear that the task should be claimed before and only the assignee can do it. Also, I don't think it's necessary to mention that mention that the task will be in the assigned status:
```
Assign a task that has been claimed before to a different user. Only the current assignee can perform the action and the new assignee should be part of the candidate users for this task.
```
What do you think?

erdemedeiros (changes requested):

activiti-core/activiti-api-impl/activiti-api-task-runtime-impl/src/main/java/org/activiti/runtime/api/impl/TaskRuntimeImpl.java
Giovanni007: I think you are right, but in this case I extracted that method from the release functionality because I needed to release a task before to assign it. I would avoid to change the logic they used in order to be sure that there are no regression in the release functionality.

activiti-api/activiti-api-task-runtime/src/main/java/org/activiti/api/task/runtime/TaskRuntime.java
Giovanni007: Yes, I agree with you. Thanks!

erdemedeiros (approved):

mauriziovitale (approved):

Issue #3425, opened 2020-09-03
Exception getting ProcessDefinitions when there are more than 1 versions in 7.1.0-M9


shkhadijah:
When getting list of process definitions and there are more than 1 version of a process definition deployed, I will get an exception. The exception is similar to this: "Query return 2 results instead of max 1".

We use the process runtime as embedded library, we exposed an API to list the active process definitions which calls the method ProcessRuntime.processDefinitions(Pageable pageable). 
From here, the exception is thrown by DeploymentQuery.executeSingleResult(CommandContext commandContext) as can be seen in the image below. 

I have also attached the full stack trace. 

![image](https://user-images.githubusercontent.com/70693797/92099915-afd24500-ee0d-11ea-8063-cb22bc0fd15f.png)
[activiti7.log](https://github.com/Activiti/Activiti/files/5167686/activiti7.log)

wangjian1984:
I also have this problem

temiotu:
should this query be modified to get the latest deployment when starting a new process.
I'm not sure what the use case would be for getting older deployments when starting a new process.

erdemedeiros:
Hi @shkhadijah and @wangjian1984, are you using auto-deployment for the process definitions, or are you deploying them programmatically?

erdemedeiros:
Also, if you guys can share the current entries for the deployment table that would be great!

shkhadijah:
> Hi @shkhadijah and @wangjian1984, are you using auto-deployment for the process definitions, or are you deploying them programmatically?

@erdemedeiros I am using auto-deployment.

erdemedeiros:
@shkhadijah and @wangjian1984, we were not able to reproduce the issue. However, [this PR](https://github.com/Activiti/Activiti/issues/3425) is bringing some improvements in that area.

PR #3441, opened 2020-09-15
fix(GH-3425): Exception getting ProcessDefinitions with multple versions

igdianov:
This PR tried to address issue https://github.com/Activiti/Activiti/issues/3425. I created a project to test deployment migration from M6 -> M9, but was not able to reproduce the problem.

- [x] Implemented latestVersion criteria in DeploymentQueryBuilder Api
- [x] Reverted DeploymentQueryBuilder singleResult override to use latestVersion from DeploymentQueryBuilder in ProcessRuntime.selectLatestDeployment
- [x] Removed existing project version from latest deployment query criteria 
- [x] Replaced process definition in-memory filtering with latest version query criteria
- [x] Fixed findLatestDeploymentVersion by name query criteria

During investigation, I discovered that the implementation of DeploymentQuery.singleResult api was overriding query builder criteria to fetch latest SpringAutoDeployment with existing project version. This may have been causing issues with deployment and process definition queries since Activiti Community does not require project manifest Json. I have removed single result override and implemented latest deployment version criteria in DeploymentQuery builder to be used in selectLatestDeployment Api. I have also removed project version criteria from latest deployment query without breaking any tests.

Another problem that I discovered was using in-memory results filtering for process definitions queries to filter latest deployments based on nullable app version field. This was causing to pick process definitions by key using previous deployment version. 

I have refactored the implementation to use latest version criteria in process definition query to always select latest definitions matching latest deployment version.

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
@@ -473,6 +473,8 @@ public class ProcessRuntimeImpl implements ProcessRuntime {
         return deploymentConverter.from(
                 repositoryService
                     .createDeploymentQuery()
+                    .deploymentName("SpringAutoDeployment")
+                    .latestVersion()
                     .singleResult()
         );
     }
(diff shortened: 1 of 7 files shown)


erdemedeiros (commented):
It looks much better!
@igdianov, from what I understood from our conversation in some cases it would be necessary to filter by `deploymentId`, but I don't see it in the PR. Does it means that it's no longer needed?

igdianov:
> It looks much better!
> @igdianov, from what I understood from our conversation in some cases it would be necessary to filter by `deploymentId`, but I don't see it in the PR. Does it means that it's no longer needed?

It is not needed as there could be several different deployments, so we cannot use deployment id to filter process definitions.

erdemedeiros (approved):
LGTM!

PR #3466, opened 2020-10-21
New analysis

FAnett1997:
"QualityGate maintainability is also added to the Current status block. Here you can find the Activiti_Activiti project in our source code management system: https://quality-gate.com/dashboard/branches/137137#overview


mteodori:
thanks but we would rather use tools that are integrated with github

FAnett1997:
What type of tools would you use that are integrated with GitHub?
Just for clairvoyance.

Marcello Teodori <notifications@github.com> (időpont: 2021. jan. 24., V,
10:05) ezt írta:

PR #3468, opened 2020-10-22
Add previous value to update event

### LINKED ISSUE #3467: VARIABLE_UPDATED event should contain previous value
VariableEvent of VARIABLE_UPDATED should contain along the current value the previous value too

zoltanpalfi:
Add previous value to update event, tests, minor cleanup

closes #3467

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-model-shared-impl/src/main/java/org/activiti/api/runtime/event/impl/VariableCreatedEventImpl.java
+++ b/activiti-core/activiti-api-impl/activiti-api-model-shared-impl/src/main/java/org/activiti/api/runtime/event/impl/VariableCreatedEventImpl.java
@@ -26,10 +26,16 @@ public class VariableCreatedEventImpl extends VariableEventImpl implements Varia
 
     public VariableCreatedEventImpl(VariableInstance entity) {
         super(entity);
+        setProcessInstanceId(entity.getProcessInstanceId());
     }
 
     @Override
     public VariableEvent.VariableEvents getEventType() {
         return VariableEvent.VariableEvents.VARIABLE_CREATED;
+    }
+
+    @Override
+    public String getProcessInstanceId() {
+        return super.getProcessInstanceId();
     }
 }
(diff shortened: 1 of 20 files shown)


activiti-core/activiti-api-impl/activiti-api-model-shared-impl/src/main/java/org/activiti/api/runtime/event/impl/VariableCreatedEventImpl.java
```diff
@@ -26,10 +26,16 @@ public VariableCreatedEventImpl() {
...
+        setProcessInstanceId(entity.getProcessInstanceId());
     }
 
     @Override
     public VariableEvent.VariableEvents getEventType() {
         return VariableEvent.VariableEvents.VARIABLE_CREATED;
     }
+
+    @Override
+    public String getProcessInstanceId() {
```
erdemedeiros: I guess we don't need this as it's only calling `super`, right?

activiti-core/activiti-api-impl/activiti-api-runtime-shared-impl/src/test/java/org/activiti/runtime/api/event/impl/ToVariableDeletedConverterTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.activiti.engine.impl.variable.VariableType;
+import org.junit.jupiter.api.Test;
+
+class ToVariableDeletedConverterTest {
+
+    private ToVariableDeletedConverter converter = new ToVariableDeletedConverter();
+
+    @Test
+    void should_convertToVariableDeletedEvent() {
+        ActivitiVariableEventImpl internalEvent = new ActivitiVariableEventImpl(ActivitiEventType.VARIABLE_CREATED);
```
erdemedeiros: I guess this should be `ActivitiEventType.VARIABLE_DELETED`.

erdemedeiros (changes requested):
Hi @zoltanpalfi, the general approach LGTM. There are two small details to be updated and I think it would be better to add an integration test covering the change in as well. Maybe [here](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeEventsIT.java). My impression is that we don't have any IT covering variable update yet. You can start an instance of any process having a user task then update a variable and check the triggered event. You can use [LocalEventSource](https://github.com/Activiti/Activiti/blob/develop/activiti-core-common/activiti-core-test/activiti-core-test-local-runtime/src/main/java/org/activiti/test/LocalEventSource.java) to get the events. 

BTW, thank you for the new tests you added for the converters!

erdemedeiros (approved):
LGTM! Thanks for the clean up in the tests!

PR #3488, opened 2020-11-22
Reset counter after each test in RetryInterceptorTest

lzx404243:
The test `org.activiti.engine.test.cfg.RetryInterceptorTest.testRetryInterceptor` is not idempotent and fails if run twice in the same JVM, because it pollutes state shared among tests. It may be good to clean this state pollution so that some other tests do not fail in the future due to the shared state polluted by this test.

Details
--
When running RetryInterceptorTest.testRetryInterceptor twice, the second run would fail with the following assertion error:
```
java.lang.AssertionError: Expected: 4. Actual: 8 (RetryInterceptorTest:80 expected:<[4]> but was:<[8]>)
```
The root cause is that the assertion is made to a static counter, which is incremented by each of the two test runs. The fix for this is to reset the static counter to `0` in `shutdownProcessEngine()`.

With the proposed fix, the test does not pollute the shared state (and passes when run twice in the same JVM).


erdemedeiros (approved):
Thanks for the contribution @lzx404243 !

Issue #3548, opened 2021-01-31
mysql-connector-java 8.0.16 can't create table


shootercheng:
When I use MySQL connect Java 8.0.16, I can't create tables

activiti version ：7.1.0.M2

project address : https://github.com/shootercheng/SpringBoot2-Activiti7

I have debugged the code and found the problem line  `org.activiti.engine.impl.db.DbSqlSession.isEngineTablePresent()`.I didn't create a table under the database.When the driver is 8.0.16, the method returns true, but when the driver is 5.1.47, the method returns false

mteodori:
please use a more recent milestone and let us know if it fixes: https://activiti.gitbook.io/activiti-7-developers-guide/releases/7.1.0-m11

shootercheng:
@mteodori 
        Activiti 7.1.0-m11 failed to download. I tried the code of the development branch, but the same problem still occurred.
Please visit my code branch [ https://github.com/shootercheng/Activiti/tree/mysql-jdbc-version]( https://github.com/shootercheng/Activiti/tree/mysql-jdbc-version), then execute `MySqlConnectionPoolTest `.

There is the same problem in this code  `org.activiti.engine.impl.db.DbSqlSession.isEngineTablePresent()`  

Exception :
```
01:10:03 [main] ERROR org.activiti.engine.impl.interceptor.CommandContext  - Error while closing command context
org.apache.ibatis.exceptions.PersistenceException: 
### Error querying database.  Cause: java.sql.SQLSyntaxErrorException: Table 'activiti_7.act_ge_property' doesn't exist
### The error may exist in org/activiti/db/mapping/entity/Property.xml
### The error may involve org.activiti.engine.impl.persistence.entity.PropertyEntityImpl.selectProperty-Inline
### The error occurred while setting parameters
### SQL: select * from ACT_GE_PROPERTY where NAME_ = ?
### Cause: java.sql.SQLSyntaxErrorException: Table 'activiti_7.act_ge_property' doesn't exist
	at org.apache.ibatis.exceptions.ExceptionFactory.wrapException(ExceptionFactory.java:30)
	at org.apache.ibatis.session.defaults.DefaultSqlSession.selectList(DefaultSqlSession.java:149)
	at org.apache.ibatis.session.defaults.DefaultSqlSession.selectList(DefaultSqlSession.java:140)
	at org.apache.ibatis.session.defaults.DefaultSqlSession.selectOne(DefaultSqlSession.java:76)
[...]

PR #3549, opened 2021-01-31
mysql driver version problem, issues #3548 

shootercheng:
fix issues #3548


activiti-core/activiti-engine/pom.xml
```diff
@@ -154,6 +154,10 @@
       <artifactId>hibernate-core</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
```
balsarori: why add mysql connector here? You should add that in your application's pom.xml if you're using mysql.

shootercheng:
forget to revert it



---Original---

shootercheng (commented):
checked

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
```diff
@@ -1169,6 +1169,10 @@ public boolean isTablePresent(String tableName) {
                 catalog = null;
             }
 
+            if (catalog == null && ProcessEngineConfigurationImpl.DATABASE_TYPE_MYSQL.equalsIgnoreCase(databaseType)) {
```
balsarori, line 1172: Did you try setting the catalog value by using `ProcessEngineConfiguration.setDatabaseCatalog` ? Will this change still work with previous MySQL versions i.e. 5.x?

shootercheng: Thanks for you anwser. just config databaseCatalog like this `<property name="databaseCatalog" value="activiti_7" />`,but  the value sholud config the same as jdbcUrl, Why not get catalog from connection?

I have tested MySQL 5.x and 8.x, please visit this
[https://github.com/shootercheng/SpringBoot2-Activiti7/blob/master/src/test/java/db/mysql/JdbcTest.java](https://github.com/shootercheng/SpringBoot2-Activiti7/blob/master/src/test/java/db/mysql/JdbcTest.java)

PR #3562, opened 2021-02-16
feat(GH-3565): add new section in Extension json for templates

### LINKED ISSUE #3565: New section in Extension json for templates
We need to provide a way to store the email template mapping for process definition. Each user task can be mapped to an email template.

```json
{
  "templates": {
    "myTaskId1": {
        "type": "file",
        "value": "https://github.com/leemunroe/responsive-html-email-template/blob/master/email.html"
    },
    "myTaskId2": {
        "type": "variable",
        "value": "myTemplateVariable"
    }
  }
}
```

User tasks can be mapped to:

- Process variables (string)
- Files (.html)
- New template (This will be stored as a project file, same as in the previous case)
- File urls (It can be located as a local resource, or external resource => http?

igdianov:
This PR adds support to store the email templates mapping of a process, i.e. each user task can be mapped to a template. We will also provide email template configuration per process definition using special mapping, i.e. `@`. The mapping should fallback to default template mapping in Rb if there are no matching binding in the process definition extension .

```json
{ 
  "templates": {
      "myTaskId1": {
          "type": "file",
          "value": "https://github.com/leemunroe/responsive-html-email-template/blob/master/email.html"
      },
      "myTaskId2": {
          "type": "variable",
          "value": "myTemplateVariable"
      },
     "@": {
          "type": "file",
          "value": "classpath:templates/email.html"
      }
  }
}
```

Fixes https://github.com/Activiti/Activiti/issues/3565


erdemedeiros (approved):
LGTM!
Note: this is covering the runtime, but we'll also need to update the modeling backend [here](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/process/Extensions.java) and update the json schema [here](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/resources/schema/process-extensions-schema.json).

mteodori:
private jira issue numbers should not be used

davidcanonieto (approved):

Issue #3571, opened 2021-02-25
BpmnXMLConverter throw a SAXParseException when URL contains UTF-8 characters 


Giovanni007:
This is an edge case that happens when the classpath contains `UTF-8` character, for example when using Tomcat installed in a directory like _`C:\Program Files\Apache Software Foundation\你好\Tomcat8.5`_. 
The error happens when parsing XSD files, and the problem is in **Xerces** library of java SDK [XMLEntityManager](https://github.com/openjdk/jdk11/blob/37115c8ea4aff13a8148ee2b8832b20888a5d880/src/java.xml/share/classes/com/sun/org/apache/xerces/internal/impl/XMLEntityManager.java#L2222-L2237) that doesn't handle the case where you have an XSD file with a UTF-8 path and that imports/include other XSDs. 

In addition, this error happens on Tomcat because the classLoader [WebappClassLoaderBase](https://github.com/apache/tomcat/blob/8.5.x/java/org/apache/catalina/loader/WebappClassLoaderBase.java#L932) has an implementation of the method `findResource` that doesn't encode the UTF-8 characters, while is not directly reproducible on SprinBoot locally because the [BuiltinClassLoader](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java#L264) encode the URL ([code](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/sun/net/www/ParseUtil.java#L49-L62)).

PR #3572, opened 2021-02-25
force path to be UTF-8 encoded when parsing BPMN20.xsd

Giovanni007:
Ref <https://github.com/Activiti/Activiti/issues/3571> The issue happened when using ACS with Tomcat so it is also in Activiti 5.x and not only in Activiti 7. 
I tried to add a unit test that replicates the behavior we have on Tomcat. The test itself is more related to the Xerces library than to the Activiti code, but I tried to add them to keep track of the error.


andrea-ligios (approved):
Great job, @Giovanni007 ! 

LGTM

erdemedeiros (approved):
LGTM!

PR #3575, opened 2021-03-01
replace all  use JavaScript Engine to goovy ；because open jdk remove the Nashorn JavaScript Engine(open jdk 14 && 14+)

cnlpj:
…tp://openjdk.java.net/jeps/372

so use goovy replace JavaScript in all testCase and process bpm xml


cnlpj:
open jdk remove the Nashorn JavaScript Engine(open jdk 14 && 14+) 
this is openjdk note website
http://openjdk.java.net/jeps/372

PR #3576, opened 2021-03-02
use goovy replace JavaScript in all testCase and process bpm xml

cnlpj:
open jdk remove the Nashorn JavaScript Engine
here is website note http://openjdk.java.net/jeps/372
so use goovy replace JavaScript in all testCase and process bpm xml


mteodori:
no need to make such a change until new LTS is out and probably we would just add rhino rather than replace with Groovy

cnlpj:
> no need to make such a change until new LTS is out and probably we would just add rhino rather than replace with Groovy

ok! but now i need use openjdk15; i try add rhino to resolve this！

PR #3598, opened 2021-03-19
fix: Change 'initMocks' method to 'openMocks'

silverskyvicto:
`MockitoAnnotations.initMocks` method is deprecated since Mockito 3.4.0.

Ref: [Class MockitoAnnotations](https://www.javadoc.io/doc/org.mockito/mockito-core/3.4.0/org/mockito/MockitoAnnotations.html#initMocks-java.lang.Object-)


erdemedeiros (changes requested):
Hi @silverskyvicto, thank you for the proposal. However, as the project upgraded to Junit 5, I think it's better to completely remove `initMocks` or `openMocks` and annotate the class with `@ExtendWith(MockitoExtension.class)`.
Would you mind updating your PR accordingly?

PR #3602, opened 2021-03-20
Fix deprecated method newinstance

silverskyvicto:
`Class.newInstance()` method is deprecated since Java 9.

Ref: [newInstance - Class Class<T>](https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--)


erdemedeiros (approved):
Thank you for the contribution, @silverskyvicto !

PR #3603, opened 2021-03-22
Use 2 threads to build project

erdemedeiros:


erdemedeiros:
It doesn't seem to decrease the build time.

PR #3633, opened 2021-04-23
Prevent circular dependency from happening

miguelruizdev:
Prevent application error context initialization errors like:

`Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.activiti.api.runtime.conf.impl.ProcessModelAutoConfiguration': Unsatisfied dependency expressed through field 'converters'; nested exception is org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'stringToMapConverter': Requested bean is currently in creation: Is there an unresolvable circular reference?`


mmartinadanx (approved):

erdemedeiros (approved):
LGTM!

Issue #3671, opened 2021-06-09
Make EL Resolvers reusable and generic


mmartinadanx:
Resolving Expression Language is useful across all activiti-cloud ecosystem (not only Activiti runtime). The current ExpressionManager implementation is coupled to [activiti-engine](https://github.com/Activiti/Activiti/tree/develop/activiti-core/activiti-engine), which is not a dependency we want to have for example in activiti-cloud connectors. 

**Scope**

We need a way to reuse some of the JUEL utilities that [activiti-engine](https://github.com/Activiti/Activiti/tree/develop/activiti-core/activiti-engine) provides, i.e: EL Resolvers. In order to do that, we can do the following steps:

Create a new maven module named `activiti-expression` under `[core-common](https://github.com/Activiti/Activiti/tree/develop/activiti-core-common)`

Move the following classes and its tests:

- `DynamicBeanPropertyELResolver`
- `ELResolverDecorator`
- `ELResolverReflectionBlockerDecorator`
- `ExpressionFactoryResolver`
- `JsonNodeELResolver`
- `ReadOnlyMapELResolver`

Rename packages: `org.activiti.engine.impl.el` with `org.activiti.core.el`

Create a generic version of `ExpressionManager` that doesn’t depend on the engine.

Reuse the new module in `activiti-engine` and remove duplicates.

PR #3675, opened 2021-06-15
Fixes Make EL Resolvers reusable and generic #3671

mmartinadanx:


activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/ELContextBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+import javax.el.CompositeELResolver;
+import javax.el.ELContext;
+import javax.el.ELResolver;
+
+/**
+ * Builder of {@link javax.el.ELContext} instances.
+ */
+public class ELContextBuilder {
```
mmartinadanx, line 30: This class can be used in [ExpressionManager](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/el/ExpressionManager.java).

I want to do it as part of a further PR.

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CommonELResolvers.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.el.BeanELResolver;
+import javax.el.ELContext;
+import javax.el.ELResolver;
+import javax.el.ListELResolver;
+import javax.el.MapELResolver;
+
+/**
+ * Builder of {@link ELContext} instances.
+ */
+public final class CommonELResolvers {
```
mmartinadanx: This class can be used in [ExpressionManager](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/el/ExpressionManager.java).

I want to do that as part of a further PR.

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/ExpressionResolver.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.core.el;
+
+import java.util.Map;
+
+public interface ExpressionResolver {
```
mmartinadanx, line 20: This won't be used by Activiti Engine, since it has its own ExpressionManager. It's intended for activiti-cloud connectors and any other Activiti extensions usage.

nazarethjim (approved):

activiti-core-common/activiti-expression-language/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>provided</scope>
```
erdemedeiros: Is it normal that you're using `provided` scope instead of `test` scope for `junit` and `assertj`?

activiti-core-common/activiti-expression-language/pom.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.activiti</groupId>
+    <artifactId>activiti-core</artifactId>
+    <version>7.1.0-SNAPSHOT</version>
+    <relativePath>../../activiti-core</relativePath>
+  </parent>
```
erdemedeiros: I'd prefer to have `activiti-core-common-dependencies` as parent, in order to keep consistency with other maven submodules under `activiti-core-common`.
```xml
  <parent>
    <artifactId>activiti-core-common-dependencies</artifactId>
    <version>7.1.0-SNAPSHOT</version>
    <relativePath>../activiti-core-common-dependencies</relativePath>
  </parent>
```
In order to do so, you'll need to move some of the dependency management from [activiti-core pom](https://github.com/Activiti/Activiti/blob/develop/activiti-core/pom.xml#L50-L64) to [activiti-core-common pom](https://github.com/Activiti/Activiti/blob/develop/activiti-core-common/pom.xml). Does it make sense?

activiti-core-common/activiti-expression-language/src/main/test/org/activiti/core/el/ELResolverReflectionBlockerDecoratorTest.java
```diff
@@ -0,0 +1,69 @@
+/*
```
erdemedeiros: The tests ended up in the wrong folder: `src/main/test` instead of `src/test/java`.

activiti-core-common/activiti-expression-language/src/main/test/org/activiti/core/el/ELResolverReflectionBlockerDecoratorTest.java
```diff
@@ -0,0 +1,69 @@
...
+        //given
+        Map<String, Object> availableVariables = Collections.singletonMap("name", "jon doe");
+        String expressionString = "${name.getClass().getName()}";
+        ExpressionResolver expressionResolver = new JuelExpressionResolver();
+
+        //then
+        assertThatExceptionOfType(IllegalArgumentException.class)
+            .as("Using Native Method: getClass in an expression")
+            .isThrownBy(() -> expressionResolver.resolveExpression(expressionString, availableVariables))
+            .as("Illegal use of Native Method in a JUEL Expression");
```
erdemedeiros: You probably meant `withMessage` instead of `as` here.

activiti-core-common/activiti-expression-language/src/main/test/org/activiti/core/el/ELResolverReflectionBlockerDecoratorTest.java
```diff
@@ -0,0 +1,69 @@
...
+        //given
+        Map<String, Object> availableVariables = Collections.singletonMap("class", String.class);
+        String expressionString = "${class.forName(\"java.lang.Runtime\").getMethods()[6].invoke()}";
+        ExpressionResolver expressionResolver = new JuelExpressionResolver();
+
+        //then
+        assertThatExceptionOfType(IllegalArgumentException.class)
+            .as("Using Reflection in an expression")
+            .isThrownBy(() -> expressionResolver.resolveExpression(expressionString, availableVariables))
+            .as("Illegal use of Reflection in a JUEL Expression");
```
erdemedeiros: `withMessage`

activiti-core-common/activiti-expression-language/src/main/test/org/activiti/core/el/JuelResolverTest.java
```diff
@@ -0,0 +1,81 @@
...
+        //given
+        Map<String, Object> availableVariables = Collections.singletonMap("name", "jon doe");
+        String expressionString = "${nameeee}";
+        ExpressionResolver expressionResolver = new JuelExpressionResolver();
+
+        //then
+        assertThatExceptionOfType(PropertyNotFoundException.class)
+            .as("Referencing an unknown variable")
+            .isThrownBy(() -> expressionResolver.resolveExpression(expressionString, availableVariables))
+            .as("Cannot resolve identifier 'nameeee'");
```
erdemedeiros: `withMessage`

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/JsonNodeELResolver.java
```diff
@@ -241,6 +241,14 @@ public Object getValue(ELContext context, Object base, Object property) {
 		return result;
 	}
 
+    /**
+     * Returns the {@link ObjectMapper} used internally to convert {@link List}
+     * properties. Subclasses may override this method to provide a specific one
+     */
+	protected ObjectMapper getObjectMapper() {
```
erdemedeiros, line 248: Instead of having subclasses only to be able to override the object mapper, I'd prefer to inject it via the constructor: Object mapper becomes a mandatory constructor parameter for JsonNodeElResolver and JuelExpressionResolver:
```java
protected ELContext buildContext (Map<String, Object> variables) {
        return new ELContextBuilder()
            .withResolvers(
                arrayResolver(),
                listResolver(),
                mapResolver(),
                jsonNodeResolver(objectMapper),
                beanResolver()
            )
            .withVariables(variables)
            .build();
    }```

erdemedeiros (changes requested):
It's looking good! It's great to have basic expression code generic enough to be used in other places as well.
Here are some small details and some questions.

activiti-core-common/activiti-expression-language/pom.xml
mmartinadanx: It was a copy paste from activiti-engine, but I realise it doesn't make sense here.

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/JsonNodeELResolver.java
mmartinadanx: I prefer composition over inheritance too, and in fact, it was my first approach. The problem is the expression:

`Context.getProcessEngineConfiguration().getObjectMapper().convertValue(resultNode, List.class);`

If we pass this to a constructor, the instance will keep whatever this returned when the instance was created.  The process engine configuration depends on ThreadLocal variables so I'm not sure the result will be the same in both cases.

We could fix this using a `Supplier<ObjectMapper>` but I thought it would be more readable this way.

erdemedeiros: It makes sense. Let's keep it then, as there are technical constraints.

erdemedeiros (approved):
LGTM!

Issue #3674, opened 2021-06-15
Split messages containing CloudProcessDeployedEvent


erdemedeiros:
Currently, [CloudProcessDeployedProducer](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-events/src/main/java/org/activiti/cloud/services/events/listeners/CloudProcessDeployedProducer.java#L42) is sending a message containing a `CloudProcessDeployedEvent` for all the process definition (and all versions) deployed in the runtime bundle. Taking into account that this event is also including the associated process definition file, the message can become very large.

In order to avoid large messages, it's better to split ProcessDeployedEvents into several messages (each message will contain a subset of process definitions).
In addition to that, [ProcessDeployedEventProducer](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/ProcessDeployedEventProducer.java#L54) should filter only the latest version of a given process definition: the previous versions were already covered.

PR #3680, opened 2021-06-23
fix(GH-3674): publish only latest process definitions for deployed processes

igdianov:
Part of https://github.com/Activiti/Activiti/issues/3674


erdemedeiros (approved):
LGTM

PR #3738, opened 2021-10-12
5.x supportin7.x

AnkurChauhan-131:
Add Activiti 5 compatibility code with Activiti5 Engine compatible code


balsarori (changes requested):
@AnkurChauhan-131 can you remove the org.activiti.form package and all classes in it? That is not required.

AnkurChauhan-131:
> @AnkurChauhan-131 can you remove the org.activiti.form package and all classes in it? That is not required.

Deleted engine/form package. But it seems Activit5 code that we have added in APS do refer to the files within engine/form package. So should i add them in APS itself

balsarori:
@AnkurChauhan-131 they should be deleted from there too

activiti-core/activiti-engine/src/main/java/org/activiti/engine/FormService.java
```diff
@@ -0,0 +1,113 @@
+/*
```
balsarori: This file can be removed

erdemedeiros:
Covered by https://github.com/Activiti/Activiti/pull/3740

PR #3745, opened 2021-10-27
Allow expressions in process variables

rbarroso:
Related issue: https://github.com/Activiti/Activiti/issues/3744

### DIFF
--- a/activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/VariableType.java
+++ b/activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/VariableType.java
@@ -25,7 +25,7 @@ import org.activiti.engine.ActivitiException;
  */
 public abstract class VariableType {
 
-    private static String expressionPattern = "^\\$\\{(.*)[\\}]$";
+    private static final Pattern EXPRESSION_PATTERN = Pattern.compile("^\\$\\{(.|\\n)*[\\}]$");
     private String name;
 
     public String getName() {
@@ -43,9 +43,6 @@ public abstract class VariableType {
     }
 
     protected boolean isExpression(Object var) {
-        if (Objects.nonNull(var) && var.getClass().isAssignableFrom(String.class) && Pattern.compile(expressionPattern).matcher((CharSequence) var).matches()) {
-            return true;
-        }
-        return false;
+        return Objects.nonNull(var) && var.getClass().isAssignableFrom(String.class) && EXPRESSION_PATTERN.matcher((CharSequence) var).matches();
     }
 }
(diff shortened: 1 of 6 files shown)


activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/VariableType.java
```diff
@@ -16,15 +16,16 @@
...
 import org.activiti.engine.ActivitiException;
 
 /**
- * Base variable type for types as defined in extension json files.
- * Used to validate variables against definition.
+ * Base variable type for types as defined in extension json files. Used to validate variables against definition.
  */
 public abstract class VariableType {
 
+    private static String expressionPattern = "^\\$\\{(.*)[\\}]$";
```
pmartinezga:
```suggestion
    private static String expressionPattern = "^\\$\\{(.|\\n)*[\\}]$";
```
If we want to support multiline expressions, then we should change the pattern like this

mmartinadanx: Use here a compiled pattern instead
```suggestion
    private static final Pattern EXPRESSION_PATTERN = Pattern.compile("^\\$\\{(.|\\n)*[\\}]$");
```

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/VariableType.java
```diff
@@ -40,4 +41,11 @@ public void setName(String name) {
     public Object parseFromValue(Object value) throws ActivitiException {
         return value;
     }
+
+    protected boolean isExpression(Object var) {
+        if (Objects.nonNull(var) && var.getClass().isAssignableFrom(String.class) && Pattern.compile(expressionPattern).matcher((CharSequence) var).matches()) {
```
mmartinadanx: NO if required
```suggestion
        return (Objects.nonNull(var) && var.getClass().isAssignableFrom(String.class) && EXPRESSION_PATTERN.matcher((CharSequence) var).matches());
```

mmartinadanx (approved):
Nice! just 2 minor comments

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/VariableType.java
rbarroso: Done

activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/variable/types/VariableType.java
rbarroso: Thanks. Simplification addressed in last commit

PR #3755, opened 2021-11-10
Developing jump commands

zhudunfeng:
Hi . This is a command I developed using the underlying API of Activiti, and I hope to get your suggestions,  I hope it will be adopted.
Currently, it has the following limitations
(1) If the process involves gateways,Nodes outside the gateway cannot jump to nodes inside the gateway.
(2) If the process involves gateways,Nodes in the same execution instance within the gateway can jump to each other.
(3) If the process involves gateways,Nodes outside the gateway can jump to each other.
 In other cases, the process does not end properly.

Waiting for you!


zhudunfeng:
A little flaw

Issue #3765, opened 2021-11-25
Add category field to process models


miguelruizdev:
As a modeler, I need to be able to define categories to group my modeled processes.

miguelruizdev:
Ref #3765

miguelruizdev:
We are going to leverage targetNamespace of Bpmn process models to achieve it:
![image](https://user-images.githubusercontent.com/32025234/145365684-1c2f935a-5521-4113-8802-f3d6de7eb455.png)
Resulting in:
![image](https://user-images.githubusercontent.com/32025234/145365796-fdf88c1f-ec09-4089-8eb2-142af46098cb.png)

PR #3784, opened 2021-12-15
Activiti/Activiti#3765 Propagate category field from internal process definitions

miguelruizdev:
Ref: Activiti/Activiti#3765


erdemedeiros (approved):
LGTM!

PR #3792, opened 2021-12-29
Remove deprecated Mergify strict mode

erdemedeiros:
Use merging queues instead

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -16,4 +16,4 @@ pull_request_rules:
     actions:
       queue:
         method: squash
-        strict: default
+        name: default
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -17,6 +14,6 @@ pull_request_rules:
       - status-success=license/cla
       - label!=do-not-merge
     actions:
-      merge:
+      queue:
         method: squash
-        strict: true
+        strict: default
```
gicappa: Looking at the blog post I think this should use `name` instead of `strict`

```yaml
actions:
      queue:
        method: merge
        name: default
```

gicappa (changes requested):

.mergify.yml
erdemedeiros: Thanks, I missed that one. Fixed.

gicappa (approved):

andrea-ligios (approved):
LGTM!

PR #3796, opened 2022-01-10
fix(GH-3798): update spring-boot.version to 2.6.2

igdianov:
This PR updates `spring-boot.version` to 2.6.2 release.

- [x] Removed circular dependencies in several auto configurations because Spring Boot 2.6+ started prohibiting circular references by default: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.6-Release-Notes#circular-references-prohibited-by-default

Part of https://github.com/Activiti/Activiti/issues/3798

### DIFF
--- a/activiti-examples/activiti-api-web-example/src/main/java/org/activiti/examples/DemoApplicationConfiguration.java
+++ b/activiti-examples/activiti-api-web-example/src/main/java/org/activiti/examples/DemoApplicationConfiguration.java
@@ -31,6 +31,8 @@ import org.springframework.security.core.authority.SimpleGrantedAuthority;
 import org.springframework.security.core.userdetails.User;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+import org.springframework.security.crypto.factory.PasswordEncoderFactories;
+import org.springframework.security.crypto.password.DelegatingPasswordEncoder;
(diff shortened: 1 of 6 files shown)


andrea-ligios (approved):
LGTM!

erdemedeiros (approved):
LGTM!

activiti-examples/activiti-api-web-example/src/main/java/org/activiti/examples/DemoApplicationConfiguration.java
```diff
@@ -83,12 +88,10 @@ protected void configure(HttpSecurity http) throws Exception {
...
                 .and()
                 .httpBasic();
-
-
     }
 
     @Bean
-    public PasswordEncoder passwordEncoder() {
+    public static PasswordEncoder passwordEncoder() {
         return new BCryptPasswordEncoder();
```
gicappa: IIRC adding a static would return a new `BCryptPasswordEncoder`  for every method call. Is this desired?

gicappa (approved):
LGTM!

activiti-examples/activiti-api-web-example/src/main/java/org/activiti/examples/DemoApplicationConfiguration.java
igdianov: Spring will instantiate this bean as singleton only once using static method. This is done to break circular dependency between bean instance and configuration class instance that uses it.

Issue #3787, opened 2021-12-20
Variables updated while completing tasks are not taking into account by expressions


erdemedeiros:
Given the following process definition and extension file:
![image](https://user-images.githubusercontent.com/26007058/146790007-1a559bf4-9d66-467b-861f-deae95a78d3a.png)

```xml
<?xml version="1.0" encoding="UTF-8"?>
<bpmn2:definitions xmlns:activiti="http://activiti.org/bpmn" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI" xmlns:dc="http://www.omg.org/spec/DD/20100524/DC" xmlns:di="http://www.omg.org/spec/DD/20100524/DI" id="model-fdee92a0-5a2b-48be-8696-245eff49aa07" name="variable-mapping-with-loop" targetNamespace="http://bpmn.io/schema/bpmn" xsi:schemaLocation="http://www.omg.org/spec/BPMN/20100524/MODEL BPMN20.xsd">
  <bpmn2:process id="Process_N4qkN051N" name="variable-mapping-with-loop" isExecutable="true">
    <bpmn2:documentation />
    <bpmn2:startEvent id="Event_1">
      <bpmn2:outgoing>SequenceFlow_1anr2ek</bpmn2:outgoing>
    </bpmn2:startEvent>
    <bpmn2:sequenceFlow id="SequenceFlow_1anr2ek" sourceRef="Event_1" targetRef="Task_125yjke" />
    <bpmn2:userTask id="Task_125yjke" name="Enter values" activiti:formKey="form-29e73efe-5bd6-48c3-8c8d-598cb0ade731" activiti:assignee="user">
      <bpmn2:incoming>SequenceFlow_1anr2ek</bpmn2:incoming>
      <bpmn2:incoming>SequenceFlow_0g8zsg1</bpmn2:incoming>
      <bpmn2:outgoing>SequenceFlow_0urignn</bpmn2:outgoing>
    </bpmn2:userTask>
    <bpmn2:sequenceFlow id="SequenceFlow_0urignn" sourceRef="Task_125yjke" targetRef="Task_0od1320" />
    <bpmn2:exclusiveGateway id="ExclusiveGateway_1mah1i8" default="SequenceFlow_0g8zsg1">
[...]

PR #3799, opened 2022-01-13
fix: update variable propagation logic

erdemedeiros:
Simplify the propagation logic and make it consistent for user tasks, service tasks and call activities:
- Execution variables are only set in the context of a multi-instance, so that they can be collected into the result collection by the execution of the multi-instance root
- In other cases, task variables, connector outputs and subprocess variables (from call activities) are updating directly the process instance variables based on the defined mapping

Possible side effects:
- previously connector outputs were made available to the current execution even in the case where no mapping was defined, meaning that connector outputs could be directly used in sequence flow expressions. After this change this will be no longer the case. In order to be able to use connector outputs in further expressions you'll need to define a mapping for it (either explicit or using one of `MAP_ALL_OUTPUTS` or `MAP_ALL` options).

Ref https://github.com/Activiti/Activiti/issues/3787


igdianov (approved):
LGTM

Issue #3825, opened 2022-02-15
Use prereleases for internal releases


erdemedeiros:
Currently, internal releases are using the pattern `MAJOR.MINOR.PATCH` for internal releases. To avoid confusion it's better to use preleases for that, i.e. using suffixes `alpha.x`.

PR #3826, opened 2022-02-15
ci: set tag pattern to be watched when generating next version

erdemedeiros:
This avoids generating internal releases with the pattern `7.x.y` while we migrate to the new versioning system.

Part of https://github.com/Activiti/Activiti/issues/3825


andrea-ligios (approved):
LGTM!

PR #3828, opened 2022-02-16
ci: change the pattern used by internal releases

erdemedeiros:
Previously, Activiti was creating internal releases (created every the CI builds on the `develop` branch) with the pattern
`MAJOR.MINOR.PATCH`. However, these are actually pre-releases, so better to use prereleases suffixes, such as `alpha`.
At the same time we are moving from Semver to Calver: the next main releases are going to have the pattern `YY.MINOR.PATCH` and so the internal releases are going to have the pattern `YY.MINOR.PATCH-alpha.COUNTER`

Part of https://github.com/Activiti/Activiti/issues/3825

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,12 +36,6 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      - uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


andrea-ligios (approved):
LGTM!

PR #3839, opened 2022-02-22
When creating, add the tenantId input parameter to the start method at startup 

licong0007:
When creating, add the tenantId input parameter to the start method at startup


licong0007:
All committers have signed the CLA.

PR #3888, opened 2022-03-23
Mnt 22933 migration timer events from aps1

vrmoreira:


vrmoreira:
This was a working-in-progress and followed a wrong investigation path

Issue #3887, opened 2022-03-23
Show processes to all involved users


balsarori:
`ProcessRuntime.processInstances` should return all processes a user is involved (initiator or involved in a task).

This include RB and Query service endpoints
`rb/v1/process-instances/`
`query/v1/process-instances/`

**Note**: Members of candidate groups will not be considered involved users for this ticket.

PR #3892, opened 2022-03-28
return processes to involved users

balsarori:
#3887

- Candidate groups:
    -  Identity links are not added to groups at runtime same way they are added to user i.e.  [task assignee](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManagerImpl.java#L166)
    - Process instance query doesn’t have a way to filter results by involved groups similar to what available to [users](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/runtime/ProcessInstanceQuery.java#L127)
- [Fixed] Process instance query `involvedUser()` doesn’t work for subprocesses. Seems that an identitylink isn't added at runtime to subprocesses like normal [parent processes](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/ExecutionEntityManagerImpl.java#L250) (workaround to use `or()` - startedBy or involved user)
- Or startedBy clause was missing in mybatis XML - fixed in this PR

### DIFF
--- a/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/api/runtime/ProcessInstanceQueryTest.java
+++ b/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/api/runtime/ProcessInstanceQueryTest.java
@@ -1836,4 +1836,21 @@ public class ProcessInstanceQueryTest extends PluggableActivitiTestCase {
 
     assertThat(processInstances).hasSize(1);
   }
+
+  @Deployment(resources = { "org/activiti/engine/test/api/oneTaskProcess.bpmn20.xml" })
+  public void testOrQueryByStartedBy() throws Exception {
+    final String authenticatedUser = "user1";
+    Authentication.setAuthenticatedUserId(authenticatedUser);
+    ProcessInstance expectedProcessInstance = runtimeService.startProcessInstanceByKey("oneTaskProcess");
+
+    ProcessInstance actualProcessInstance = runtimeService.createProcessInstanceQuery()
+        .or()
+        .startedBy(authenticatedUser)
+        .processDefinitionKey("notTheProcessDefinitionKey")
+        .endOr()
+        .singleResult();
+
+    assertThat(actualProcessInstance.getProcessInstanceId())
+        .isEqualTo(expectedProcessInstance.getProcessInstanceId());
+    }
 }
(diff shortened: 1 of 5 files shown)


Giovanni007 (changes requested):
Do we need a test for the case when the user is part of a candidate group?
Can you specify who is the involved users in the description of the related Github issue? Thx 

Update:
As discussed with @balsarori we won't consider candidate groups for process instances

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/api/runtime/ProcessInstanceQueryTest.java
```diff
@@ -1836,4 +1836,22 @@ public void testQueryStartedBy() throws Exception {
 
     assertThat(processInstances).hasSize(1);
   }
+
+  @Deployment(resources = { "org/activiti/engine/test/api/oneTaskProcess.bpmn20.xml" })
+  public void testOrQueryByStartedBy() throws Exception {
+    final String authenticatedUser = "user1";
+    Authentication.setAuthenticatedUserId(authenticatedUser);
+    ProcessInstance expectedProcessInstance = runtimeService.startProcessInstanceByKey("oneTaskProcess");
```
erdemedeiros, line 1844: I think you should start a second instance of the process using another user, so you make sure the query is actually filtering the results and not returning all of them.

erdemedeiros (changes requested):
One small detail in the test. LGTM otherwise.

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/api/runtime/ProcessInstanceQueryTest.java
balsarori: The query uses `singleResult` so it should return only one result otherwise it will fail. Does it make sense?

balsarori: Ok I think I got your point now :) I'll start another process instance using a different user.

Giovanni007 (approved):
LGTM

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/api/runtime/ProcessInstanceQueryTest.java
erdemedeiros: Exactly, in this way, you have two instances in the database, but one is not matching the filter and so it should not be returned.

erdemedeiros (approved):
LGTM!

PR #3904, opened 2022-04-07
[#3920] caching process extension service and remove variables from process definition

jsokolowskii:
- moving the variable definitions to activiti-cloud CloudProcessDefinition,
- adding caching to to ProcessExtensionService to cache reading from JSON file
Ref: https://github.com/Activiti/Activiti/issues/3920


jesty (approved):
LGTM

PR #3908, opened 2022-04-19
Fix #3909 - Validate should check Model processes id to avoid having duplicated 

### LINKED ISSUE #3909: Raise an error whenever in the same BPMN model two process have the same ids
This can happen on copy pasting and it's not a blocker when it can create problem on deploying the app.

VitoAlbano:
…PMN have the same Ids

### DIFF
--- a/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
+++ b/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
@@ -33,14 +33,20 @@ public class TestProcessUtil {
    */
   public static BpmnModel createOneTaskBpmnModel() {
     BpmnModel model = new BpmnModel();
-    model.addProcess(createOneTaskProcess(1));
+    model.addProcess(createOneTaskProcess());
     return model;
   }
 
-  public static org.activiti.bpmn.model.Process createOneTaskProcess(int index) {
+  public static org.activiti.bpmn.model.Process createOneTaskProcessWithId(String id) {
+    org.activiti.bpmn.model.Process process = createOneTaskProcess();
+    process.setId(id);
+    return process;
+  }
+
+  public static org.activiti.bpmn.model.Process createOneTaskProcess() {
     org.activiti.bpmn.model.Process process = new org.activiti.bpmn.model.Process();
 
-    process.setId("oneTaskProcess".concat("-"+index));
+    process.setId("oneTaskProcess");
     process.setName("The one task process");
 
     StartEvent startEvent = new StartEvent();
--- a/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
+++ b/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
@@ -33,14 +33,14 @@ public class TestProcessUtil {
    */
   public static BpmnModel createOneTaskBpmnModel() {
     BpmnModel model = new BpmnModel();
-    model.addProcess(createOneTaskProcess());
+    model.addProcess(createOneTaskProcess(1));
     return model;
   }
 
-  public static org.activiti.bpmn.model.Process createOneTaskProcess() {
+  public static org.activiti.bpmn.model.Process createOneTaskProcess(int index) {
     org.activiti.bpmn.model.Process process = new org.activiti.bpmn.model.Process();
 
-    process.setId("oneTaskProcess");
+    process.setId("oneTaskProcess".concat("-"+index));
     process.setName("The one task process");
 
     StartEvent startEvent = new StartEvent();
--- a/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
+++ b/activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
@@ -22,6 +22,8 @@ import org.activiti.bpmn.model.SequenceFlow;
 import org.activiti.bpmn.model.StartEvent;
 import org.activiti.bpmn.model.UserTask;
 
+import java.util.UUID;
+
 /**
 
  */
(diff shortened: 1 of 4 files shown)


activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/util/TestProcessUtil.java
```diff
@@ -33,14 +33,14 @@ public class TestProcessUtil {
    */
   public static BpmnModel createOneTaskBpmnModel() {
     BpmnModel model = new BpmnModel();
-    model.addProcess(createOneTaskProcess());
+    model.addProcess(createOneTaskProcess(1));
     return model;
   }
 
-  public static org.activiti.bpmn.model.Process createOneTaskProcess() {
+  public static org.activiti.bpmn.model.Process createOneTaskProcess(int index) {
```
erdemedeiros: I think it's better to extract a method that takes the entire processId as a parameter instead of passing an index to concatenate it. You can keep one method that takes no parameter and call the new one passing `UUID.randomUUID().toString()` as process id, so there is no need to pass an index in this case.

activiti-core/activiti-engine/src/test/java/org/activiti/standalone/validation/DefaultProcessValidatorTest.java
```diff
@@ -304,27 +299,36 @@ public void testAllNonExecutableProcesses() {
...
+        org.activiti.bpmn.model.Process firstProcess = TestProcessUtil.createOneTaskProcess(1);
+        firstProcess.setExecutable(true);
+        bpmnModel.addProcess(firstProcess);
+
+        org.activiti.bpmn.model.Process secondProcess = TestProcessUtil.createOneTaskProcess(1);
+        secondProcess.setExecutable(true);
+        bpmnModel.addProcess(secondProcess);
+
+        List<ValidationError> errors = processValidator.validate(bpmnModel);
+        assertThat(errors).hasSize(1);
```
erdemedeiros, line 310: It's better to check the content of the error, not only the size.

activiti-core/activiti-engine/src/test/java/org/activiti/standalone/validation/DefaultProcessValidatorTest.java
```diff
@@ -286,16 +286,11 @@ public void testInvalidMessageFlow() throws Exception {
...
-   * If all processes in a deployment are not executable, throw an exception as this doesn't make sense to do.
-   */
   @Test
-  public void testAllNonExecutableProcesses() {
+  public void should_raiseAValidationError_when_noProcessIsExecutable() {
     BpmnModel bpmnModel = new BpmnModel();
     for (int i = 0; i < 5; i++) {
-      org.activiti.bpmn.model.Process process = TestProcessUtil.createOneTaskProcess();
+      org.activiti.bpmn.model.Process process = TestProcessUtil.createOneTaskProcess(i);
       process.setExecutable(false);
```
erdemedeiros: In order to improve the test readability, you could extract the methods `createExecutableProcess` and `createNonExecutableProcess`. The entire `for` block becomes:
```java
for (int i = 0; i < 5; i++) {
	bpmnModel.addProcess(createNonExecutableProcess());
}
```
BTW, I don't think we need 5 processes here, one should be enough, so no `for` block at all, only `bpmnModel.addProcess(createNonExecutableProcess())`. What do you think?

activiti-core/activiti-process-validation/src/main/java/org/activiti/validation/validator/impl/BpmnModelValidator.java
```diff
@@ -91,4 +100,17 @@ protected boolean validateAtLeastOneExecutable(BpmnModel bpmnModel, List<Validat
...
+  protected List<Process> getProcessesWithSameName(final List<Process> processes) {
+            List<Process> filteredProcesses = processes.stream()
+                .filter(process -> process.getName() != null).collect(Collectors.toList());
+          return getDuplicatesMap(filteredProcesses).values().stream()
+              .filter(duplicates -> duplicates.size() > 1)
+              .flatMap(Collection::stream)
+              .collect(Collectors.toList());
+  }
+
+  private static Map<String, List<Process>> getDuplicatesMap(List<Process> personList) {
```
erdemedeiros: The name of the parameter looks wrong: `processes`?

activiti-core/activiti-process-validation/src/main/java/org/activiti/validation/validator/impl/BpmnModelValidator.java
```diff
@@ -91,4 +100,17 @@ protected boolean validateAtLeastOneExecutable(BpmnModel bpmnModel, List<Validat
 		return nrOfExecutableDefinitions > 0;
   }
 
+  protected List<Process> getProcessesWithSameName(final List<Process> processes) {
```
erdemedeiros: Same id, not same name.

activiti-core/activiti-process-validation/src/main/java/org/activiti/validation/validator/impl/BpmnModelValidator.java
```diff
@@ -34,6 +37,12 @@ public class BpmnModelValidator extends ValidatorImpl {
   @Override
   public void validate(BpmnModel bpmnModel, List<ValidationError> errors) {
 
+    List<Process> processesDuplicated = getProcessesWithSameName(bpmnModel.getProcesses());
+    if(processesDuplicated.size() > 0 ) {
```
erdemedeiros: You can use `isEmpty()` method:
`if (!processesDuplicated.isEmpty()) {`

activiti-core/activiti-process-validation/src/main/java/org/activiti/validation/validator/Problems.java
```diff
@@ -72,6 +72,7 @@ public interface Problems {
 
   String BPMN_MODEL_TARGET_NAMESPACE_TOO_LONG = "activiti-bpmn-model-target-namespace-too-long";
 
+  String PROCESS_DEFINITION_NAME_NOT_UNIQUE = "activiti-process-definition-name-duplicated";
```
erdemedeiros: Please, rename it to match what is actually being checked (`id`, not `name`).

activiti-core/activiti-process-validation/src/main/java/org/activiti/validation/validator/impl/BpmnModelValidator.java
```diff
@@ -34,6 +37,12 @@ public class BpmnModelValidator extends ValidatorImpl {
...
+        addError(errors, Problems.PROCESS_DEFINITION_NAME_NOT_UNIQUE, processesDuplicated.get(0),
+            "The name of the process definition must be unique");
```
erdemedeiros: `The id`

erdemedeiros (changes requested):

erdemedeiros (approved):
LGTM!

PR #3910, opened 2022-04-23
Clean Activiti Code 

gicappa:
Uncle Bob said it clearly: 

> [The only way to go fast is to go well](https://twitter.com/unclebobmartin/status/1163789159309434880?lang=en). 

So let’s apply this principle to Activiti, let’s Clean Activiti Code!

The idea is to make a week of refactoring to improve the design of the code so that working on activiti can be more fun and effective. 

**Goals**:

- **Improve SRP**: we want to be obsessed with naming correctly the classes, splitting classes/methods too long, delete dead/useless code. Removing code gets a reward! 
- **Improve DIP**: we need to separate domain from implementation. Modules with the higher abstraction (business rules) must not depend on the ones with the concrete implementation (like database and message queuing).
- **Improve Testing**: follow the testing pyramid and be sure that the tests are covering the whole application but not more than that. We need to trust our tests and avoid too many integration tests to make them faster.

Just as an example [this is a class](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java) I would start from.


gicappa:
This was just an initial try I did before the hackathon. Closing in favor of the official PR https://github.com/Activiti/Activiti/pull/3912

PR #3296, opened 2020-05-20
fix: add support for simple process variable types in IntegrationContext

igdianov:
This PR fixes variable type names for IntegrationContext variables passed between Rb and Connectors to use simple names instead of full Java class name for data types in the raw json to match types in the ProcessVariables entity stored inside Query and Audit databases, i.e.

The current format contains full Java class name:
```
          "outBoundVariables": {
            "processDefinitionId": {
              "type": "java.lang.String",
              "value": "HeadersConnectorProcess:1:dc32c0bc-9a56-11ea-afc3-92581787ab92"
            },
            "processDefinitionVersion": {
              "type": "java.lang.Integer",
              "value": "1"
            },
            "processDefinitionKey": {
              "type": "java.lang.String",
              "value": "HeadersConnectorProcess"
            }
            "amount": {
              "type": "java.lang.BigDecimal",
              "value": "10.00"
            }
          }
```
The updated format, will contain simple type names:

```
          "outBoundVariables": {
            "processDefinitionId": {
              "type": "string",
              "value": "HeadersConnectorProcess:1:dc32c0bc-9a56-11ea-afc3-92581787ab92"
            },
            "processDefinitionVersion": {
              "type": "integer",
              "value": "1"
            },
            "processDefinitionKey": {
              "type": "string",
              "value": "HeadersConnectorProcess"
            }
            "amount": {
              "type": "bigdecimal",
              "value": "10.00"
            }
         }

```

PR #3925, opened 2022-05-10
#3296 add processDefinitionId to variable created event

jsokolowskii:
Adding process definition id to variable created event to later use it to retrieve variable definition id in order to store it on query service database.
Ref: https://github.com/Activiti/Activiti/issues/3926


jesty (approved):
LGTM

miguelruizdev (approved):

PR #3965, opened 2022-06-15
Activiti/Activiti#3964 Fix variable mapping when no defined process variable matches

### LINKED ISSUE #3964: The input mapping is not working when previous task is MAP_OUTPUTS
In a service task, for example, we can use implicit mappings like "Map all inputs/outputs", "Map inputs" or "Map outputs" but it is not working as expected because in Activiti when calculating the input mapping, it only maps variables that have been defined in the process extensions as you can see here:
https://github.com/Activiti/Activiti/blob/4cc9d1c6c2f461d49ee7a26b4ca34dac887b0cc8/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java#L60

If we have a process where the first service task set its mapping to "Map all outputs" and we want to use one of those outputs in a subsequent service task this will prevent our inputs in the second task to work, because the outputs have not been created in the process instance scope as they are not mapped to an existing variable.

We need to create the variables in the scope of the process instance when using the implicit mapping provided or not the name of the output matches a process variable defined in the process extensions

pmartinezga:
closes https://github.com/Activiti/Activiti/issues/3964


miguelruizdev (approved):
LGTM!
Amazing job and troubleshooting

PR #3967, opened 2022-06-15
3966 fix context in output mapping expression resolution

### LINKED ISSUE #3966: The output mapping is resolving expressions using only the task context
The output mapping is using only the context of the task to resolve expressions as for this:
https://github.com/Activiti/Activiti/blob/4cc9d1c6c2f461d49ee7a26b4ca34dac887b0cc8/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java#L177

This means that we can not use process variables to be mixed with task variables to build output, and we can neither use execution resolvers like `${initiator}` or `${assignee}`

pmartinezga:
closes https://github.com/Activiti/Activiti/issues/3966


miguelruizdev (approved):
LGTM

PR #3981, opened 2022-07-01
fix(GH-3980): Add support for analytics variable extension property in Activiti engine

igdianov:
This PR adds `analytics` Boolean property for variable definition extension.

Part of https://github.com/Activiti/Activiti/issues/3980


andrea-ligios (approved):
LGTM!

Issue #4048, opened 2022-08-22
Align version in POM files with the current release


erdemedeiros:
The version in POM files should be updated from `7.1.0-SNAPSHOT` to `7.5.0-SNAPHOT`, the release currently under development.

PR #4049, opened 2022-08-22
Update version to `7.5.0-SNAPSHOT`

erdemedeiros:
- update pom files to version 7.5.0-SNAPSHOT
- update build to use action that calculates the next version based on the current pom version

Ref: #4048


igdianov (approved):
LGTM!

Issue #3940, opened 2022-05-23
Wrong VERSION field in ProcessEngine interface in 7.3.0 release


coldsmirk:
## Here is the ProcessEngine source code in 7.3.0 version
```java
package org.activiti.engine;


import org.activiti.engine.api.internal.Internal;

/**
 * Provides access to all the services that expose the BPM and workflow operations.
 *
 * <ul>
 * <li>
 * <b>{@link org.activiti.engine.RuntimeService}: </b> Allows the creation of {@link org.activiti.engine.repository.Deployment}s and the starting of and searching on
 * {@link org.activiti.engine.runtime.ProcessInstance}s.</li>
 * <li>
 * <b>{@link org.activiti.engine.TaskService}: </b> Exposes operations to manage human (standalone) {@link org.activiti.engine.task.Task}s, such as claiming, completing and assigning tasks</li>
 * <li>
 * <b>{@link org.activiti.engine.ManagementService}: </b> Exposes engine admin and maintenance operations</li>
 * <li>
 * <b>{@link org.activiti.engine.HistoryService}: </b> Service exposing information about ongoing and past process instances.</li>
 * </ul>
 *
 * Typically, there will be only one central ProcessEngine instance needed in a end-user application. Building a ProcessEngine is done through a {@link ProcessEngineConfiguration} instance and is a
 * costly operation which should be avoided. For that purpose, it is advised to store it in a static field or JNDI location (or something similar). This is a thread-safe object, so no special
 * precautions need to be taken.
 *
 */
@Internal
public interface ProcessEngine {

  /** the version of the activiti library */
  public static String VERSION = "7.1.0-M6"; // Note the extra -x at the end. To cater for snapshot releases with different database changes

  /**
   * The name as specified in 'process-engine-name' in the activiti.cfg.xml configuration file. The default name for a process engine is 'default
   */
  String getName();

  void close();

  RepositoryService getRepositoryService();

  RuntimeService getRuntimeService();

  TaskService getTaskService();

  HistoryService getHistoryService();

  ManagementService getManagementService();

  DynamicBpmnService getDynamicBpmnService();

  ProcessEngineConfiguration getProcessEngineConfiguration();

}
```

## We can see the wrong version literal '7.1.0-M6' in the class above. Can it be fixed? Thanks in advance.

PR #4057, opened 2022-09-04
Fix issue: wrong VERSION field in ProcessEngine interface.

pandaapo:
Fix the issue #3940

* Change the hardcoding method of writing version in `ProcessEngine` class to fetch from the pom file automatically.
   Use `<resources>` configuration provided by maven and `@project.version@` in projectVersion.properties file to get project version specified in `<version>` in the pom.

* Use PL/SQL or SQL variable, string replacement  to modify the SQL fragment in the scripts that initializes the Table `act_ge_property` data, so that it can apply the process engine's version in the application when the script is executed. The SQL scripts modified includes Db2, H2, HSQL, MSSQL, MySQL, Oracle, PostgreSQL.

### DIFF
--- a/activiti-core/activiti-engine/pom.xml
+++ b/activiti-core/activiti-engine/pom.xml
@@ -118,6 +118,42 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.ibm.db2.jcc</groupId>
+      <artifactId>db2jcc</artifactId>
+      <version>db2jcc4</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hsqldb</groupId>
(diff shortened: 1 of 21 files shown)


pandaapo:
The SQL statements I modified all run correctly in the application by JDBC. Why does Codacy regard them as syntax error?

pandaapo:
> The SQL statements I modified all run correctly in the application by JDBC. Why does Codacy regard them as syntax error?

This makes me feel uncertain. I add a PR #4058 as a second solution.

pandaapo:
> The SQL statements I modified all run correctly in the application by JDBC. Why does Codacy regard them as syntax error?

I connacted with Codacy. This is the reply. So I think that is not my problem.
codacy

PR #4058, opened 2022-09-04
Fix hard coding VERSION in ProcessEngine. Make it read version info from pom.

### LINKED ISSUE #3940: Wrong VERSION field in ProcessEngine interface in 7.3.0 release

pandaapo:
Fixes #3940. 

* Change the hardcoding of version in `ProcessEngine` to fetch from the pom file automatically.
  Use `<resources>` configuration provided by maven and `@project.version@` in projectVersion.properties file to get project version from `<version>` in the pom.

* Use string replacement based on regex to modify the SQL fragment in the scripts that initializes the Table `act_ge_property` data, so that it can apply the process engine's version in the application when the SQL script is executed by Activiti.

### DIFF
--- a/activiti-core/activiti-engine/pom.xml
+++ b/activiti-core/activiti-engine/pom.xml
@@ -118,6 +118,42 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.ibm.db2.jcc</groupId>
+      <artifactId>db2jcc</artifactId>
+      <version>db2jcc4</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hsqldb</groupId>
+      <artifactId>hsqldb</artifactId>
+      <version>2.2.7</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.microsoft.sqlserver</groupId>
+      <artifactId>sqljdbc4</artifactId>
+      <version>4.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
+      <version>5.1.46</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.oracle.database.jdbc</groupId>
+      <artifactId>ojdbc10</artifactId>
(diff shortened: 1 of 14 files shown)


pandaapo:
@igdianov Could you review this PR?

Issue #4062, opened 2022-09-08
Bpmn process add subProcess sequence flow validate in version 6.x


guankai:
When using Bpmn to build an activiti process, when a subprocess appears in the process and use the function Process.addFlowElement(subprocess), the SequenceFlowValidator will prompt "activiti-seq-flow-invalid-src" or "activiti-seq-flow-invalid-target" because it cannot find the SequenceFlow in the subprocess.

PR #4063, opened 2022-09-08
Update Process.java

guankai:
resolve the issue #4062 Bpmn process add subProcess sequence flow validate
when process add flow element subprocess, put the subprocess's flow element to process.


guankai:
please check this pull request

PR #4100, opened 2022-10-07
add CustomFunctionProvider to ExpressionManager

jsokolowskii:
https://github.com/Activiti/Activiti/issues/4101

Adding custom function providers to expression manager.

### DIFF
--- a/activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CustomFunctionProvider.java
+++ b/activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CustomFunctionProvider.java
@@ -17,5 +17,5 @@ package org.activiti.core.el;
 
 public interface CustomFunctionProvider {
 
-    void addCustomFunctions(ActivitiElContext elContext) throws NoSuchMethodException;
+    void addCustomFunctions(ActivitiElContext elContext);
 }
--- a/activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CustomFunctionProvider.java
+++ b/activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CustomFunctionProvider.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2010-2020 Alfresco Software, Ltd.
+ *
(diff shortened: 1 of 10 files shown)


activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CustomFunctionProvider.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.core.el;
+
+public interface CustomFunctionProvider {
+
+    void addCustomFunctions(ActivitiElContext elContext) throws NoSuchMethodException;
```
erdemedeiros: Why do we need to throw `NoSuchMethodException ` here?

erdemedeiros (changes requested):
Apart from my question related to `NoSuchMethodException` it would also be great to add a test covering this. You could define a test bean implementing a very simple `CustomFunctionProvider ` and evaluate an expression that uses it, for example.

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/CustomFunctionProvider.java
jsokolowskii: It's thrown by `class.getMethod()` which is used to provide the method. I can catch it inside the implementation though.

jsokolowskii: Removed.

jsokolowskii:
> Apart from my question related to `NoSuchMethodException` it would also be great to add a test covering this. You could define a test bean implementing a very simple `CustomFunctionProvider ` and evaluate an expression that uses it, for example.

@erdemedeiros I added a simple test. More robust ones should be added in the future.

erdemedeiros (approved):
LGTM!

PR #4102, opened 2022-10-08
DeployCmd  orderby bug

LoveMyOrange:
When I use StrongUuidGenerator to generate uuid,orderByDeploymentId() may be wrong.
You used orderByDeploymenTime() in [if] at line 64,but used orderByDeploymentId() in [else] at line 70.


LoveMyOrange:
@erdemedeiros  Excuse me, Is it convenient for you to see this PR when you are free

PR #4103, opened 2022-10-08
ReceiveTask.java typo bug

LoveMyOrange:
Here is a typing error
This is different from common typing errors. The setValues () here will result in calling the methods of the parent class

Although it is intended to call the method of the parent class

eg: In the ManualTask, it calls setValues() of this class first, then uses setValues of this class to call setValues() of the parent class


LoveMyOrange:
@erdemedeiros  Is it convenient for you to see this PR when you are free

Issue #4089, opened 2022-09-26
User task assignee should be captured for usage in variable mapping.


balsarori:
It should be possible to map the user task assignee to a process variable. When map all (or map all out) mapping type is used the user task assignee should be mapped to `sys_task_assignee` variable.

PR #4114, opened 2022-10-18
Revert "Activiti/Activiti#3855 Add Task Assignee in expression resolver"

balsarori:
Reverts Activiti/Activiti#3856
Changes don't work. Capturing task assignee is possible using mapping see #4089


Giovanni007 (approved):
LGTM!

PR #4181, opened 2022-12-19
Restrict starting processes to candidate starters

### LINKED ISSUE #4180: Restrict who can start a process using candidate starters
Use candidate starters to restrict who can start a process instance. In order to maintain backwards compatibility, auto insert an identityLink between process definition and a group called `*` when candidate starters don't exist in the process definition XML file.

Note: this should be handled at the API level not at the core engine level.

balsarori:
fixes #4180

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
@@ -68,7 +68,6 @@ import org.activiti.runtime.api.model.impl.APIProcessDefinitionConverter;
 import org.activiti.runtime.api.model.impl.APIProcessInstanceConverter;
 import org.activiti.runtime.api.model.impl.APIVariableInstanceConverter;
 import org.activiti.runtime.api.query.impl.PageImpl;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.transaction.annotation.Transactional;
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
@@ -15,6 +15,7 @@
  */
 package org.activiti.runtime.api.impl;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -99,9 +100,6 @@ public class ProcessRuntimeImpl implements ProcessRuntime {
 
     private final SecurityManager securityManager;
 
-    @Value("${activiti.candidateStarters.enabled:false}")
-    private boolean candidateStartersEnabled;
-
     public ProcessRuntimeImpl(RepositoryService repositoryService,
                               APIProcessDefinitionConverter processDefinitionConverter,
                               RuntimeService runtimeService,
@@ -147,11 +145,9 @@ public class ProcessRuntimeImpl implements ProcessRuntime {
     }
 
     private ProcessDefinitionQuery createProcessDefinitionQueryWithAccessCheck() {
-        ProcessDefinitionQuery processDefinitionQuery = repositoryService.createProcessDefinitionQuery();
-        if (candidateStartersEnabled) {
-            processDefinitionQuery.startableByUser(securityManager.getAuthenticatedUserId());
-        }
-        return processDefinitionQuery;
+        return repositoryService.createProcessDefinitionQuery()
+                                .startableByUser(securityManager.getAuthenticatedUserId())
+                                .startableByGroups(getCurrentUserGroupsIncludingEveryOneGroup());
     }
 
     private Optional<org.activiti.engine.repository.ProcessDefinition> findLatestProcessDefinition(ProcessDefinitionQuery processDefinitionQuery) {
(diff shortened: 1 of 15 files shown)


activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
```diff
@@ -555,4 +550,9 @@ private boolean isATaskAssigneeOrACandidate(String processInstanceId) {
         return taskQuery.count() > 0;
     }
 
+    private List<String> getCurrentUserGroupsIncludingEveryOneGroup() {
+        List<String> groups = new ArrayList<>(securityManager.getAuthenticatedUserGroups());
+        groups.add("*");
```
Giovanni007: is "*" a group? Can we make it more explicit? for example use another name or set a constant with a meaningful name? Also, do we cover this case with tests?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/repository/ProcessDefinitionQuery.java
```diff
@@ -140,10 +141,16 @@ public interface ProcessDefinitionQuery extends Query<ProcessDefinitionQuery, Pr
...
 
     /**
-     * Only selects process definitions which given userId is authoriezed to start
+     * Only selects process definitions which given userId is authorized to start
      */
     ProcessDefinitionQuery startableByUser(String userId);
 
+    /**
+     * Only selects process definitions which given group members are authorized to start
+     * If not set and startableByUser is set, the members of that user will be used
```
Giovanni007: I don't understand  what this means: `the members of that user will be used` can you explain?

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
balsarori: I'm trying to avoid a specific group name since its not a real group that exists.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/repository/ProcessDefinitionQuery.java
balsarori: it should be `the groups of that user` :) I will update it

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ProcessRuntimeImpl.java
balsarori: Regarding tests `nonCandidateStarter_should_getProcessDefinitionWithNoCandidatesSet` covers that. Additionally, by default now no one can get a process definition or start a process unless the user is either a candidate or the process definition has an identitylink to the `*` group

Giovanni007:
The code is ok but I don't like the fact that we are going to introduce this immaginary group `*`. 
I don't understand if it is possible ,instead of using this * group,  to change the query to return all the process where user is a starter + all the process where there are not user/group starters (that would mean that anyone can start them).
If we need to restrict some process so that can only be started programmatically, I will introduce something to distinguish them and not relay on the fact that there are no user/group selected as candidate starters.
Happy to approve this if the introduction of "*" is not a big deal in your opinion

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
```diff
@@ -274,6 +276,22 @@ public void setCandidateStarterGroups(List<String> candidateStarterGroups) {
     this.candidateStarterGroups = candidateStarterGroups;
   }
 
+  public boolean isCandidateStarterUsersExists() {
+    return candidateStarterUsersExists;
+  }
```
erdemedeiros: Would you mind to rename the field to `candidateStarterUsersDefined`? I think the getter name will sound better in this case: `isCandidateStarterUsersDefined()`. What do you think?

activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/boot/CandidateStartersDeploymentConfigurer.java
```diff
@@ -0,0 +1,46 @@
...
+    public class CandidateStartersDeploymentHelper extends BpmnDeploymentHelper {
+        @Override
+        public void addAuthorizationsForNewProcessDefinition(Process process, ProcessDefinitionEntity processDefinition) {
+            super.addAuthorizationsForNewProcessDefinition(process, processDefinition);
+            if (process != null &&
+                !process.isCandidateStarterUsersExists() &&
+                !process.isCandidateStarterGroupsExists()) {
+                addAuthorizationsFromIterator(Context.getCommandContext(), List.of(EVERYONE_GROUP), processDefinition, ExpressionType.GROUP);
+            }
+        }
```
erdemedeiros: We need to double check what happens when an application is upgraded or have the runtime updated: I think the processes will get redeployed with a new version, but please, make sure that this is covered during the verification of this ticket.

erdemedeiros (changes requested):
I have only one naming suggestion and I added a small note about the coverage of some specific use cases. LGTM, otherwise.

activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/boot/CandidateStartersDeploymentConfigurer.java
balsarori: Yes thats what I think too. The process will be redeployed and identity links will be inserted. That should sync the query service too. But you are right we need to double check that.

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
balsarori: Not sure how will that sound in case of `activiti:candidateStarterUsers="" activiti:candidateStarterGroups=""` because in that case the value should be set to true.

erdemedeiros: Yes, in this case it's defined to empty, a bit controversial maybe.

erdemedeiros (approved):
LGTM!

PR #4197, opened 2023-01-03
Methods to check for error boundary events

LorenzoDiGiuseppe:
This PR introduces two new methods:

- `BoundaryEvent.hasErrorEventDefinition()`: checks if the boundary event contains an `ErrorEventDefinition`
- `ServiceTask.hasBoundaryErrorEvents()`: checks if the service task contains a boundary error event

It is part of the solution for Activiti/Activiti#4184.


erdemedeiros (approved):
LGTM!

PR #4208, opened 2023-01-13
AAE-12101 publish only one event for the latest version

jesty:
In the previous version the RB was producing 1 event for each deployed version, while we need to produce only one for the latest version.

Ref: https://github.com/Activiti/Activiti/issues/4209


Giovanni007 (approved):
LGTM!

PR #4210, opened 2023-01-13
Migration fix for Initiator issue

AnkurChauhan-131:


balsarori (changes requested):
We need this for all supported databases not just MySQL

balsarori (approved):

PR #4223, opened 2023-01-26
MNT-23269 - Fixed the database error that happens when we try to end …

antoniojfelix:
…a task with an empty 'currentDistributionList' variable

https://alfresco.atlassian.net/browse/MNT-23269

### DIFF
--- a/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java
+++ b/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java
@@ -42,6 +42,7 @@ import org.activiti.engine.impl.el.UelExpressionCondition;
 import org.activiti.engine.impl.interceptor.CommandContext;
 import org.activiti.engine.impl.persistence.entity.ExecutionEntity;
 import org.activiti.engine.impl.persistence.entity.ExecutionEntityManager;
+import org.activiti.engine.impl.persistence.entity.TaskEntity;
 import org.activiti.engine.impl.util.CollectionUtil;
 import org.activiti.engine.impl.util.condition.ConditionUtil;
 import org.slf4j.Logger;
@@ -299,7 +300,15 @@ public class TakeOutgoingSequenceFlowsOperation extends AbstractOperation {
     }
 
     protected void cleanupExecutions(FlowElement currentFlowElement) {
-        if (execution.getParentId() != null && execution.isScope()) {
+        if (execution.getParentId() != null && execution.isScope() && execution.hasInstances()) {
+            for (TaskEntity taskEntity : commandContext.getTaskEntityManager().findTasksByProcessInstanceId(execution.getProcessInstanceId()))
+            {
+                //If there's a task with the current execution associated, this execution cannot be deleted yet or else it will throw a
+                //database error
+                if(taskEntity.getExecutionId().equals(execution.getId())){
+                    return;
+                }
+            }
 
             // If the execution is a scope (and not a process instance), the scope must first be
             // destroyed before we can continue and follow the sequence flow
--- a/activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/DelegateExecution.java
+++ b/activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/DelegateExecution.java
@@ -175,5 +175,14 @@ public interface DelegateExecution extends VariableScope {
    * @param isMultiInstanceRoot
    */
   void setMultiInstanceRoot(boolean isMultiInstanceRoot);
+  /**
+   * Returns whether this execution has instances or not.
+   */
+  Boolean hasInstances();
+  /**
+   * Changes whether this execution has instances or not.
+   * @param hasInstances
+   */
+  void setInstances(Boolean hasInstances);
 
 }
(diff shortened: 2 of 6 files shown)


balsarori:
Can you provide a unit test that reproduces the issue? Or is this issue limited to activiti engine 5 to activiti 7 upgrade?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java
```diff
@@ -299,7 +300,15 @@ protected void cleanupCompensation() {
     }
 
     protected void cleanupExecutions(FlowElement currentFlowElement) {
-        if (execution.getParentId() != null && execution.isScope()) {
+        if (execution.getParentId() != null && execution.isScope() && execution.hasInstances()) {
+            for (TaskEntity taskEntity : commandContext.getTaskEntityManager().findTasksByProcessInstanceId(execution.getProcessInstanceId()))
```
balsarori, line 304: what about other types of tasks such as manual task, integration tasks, etc! Shouldn't that apply to them?

antoniojfelix: I believe that this problem will only happen with UserTask types because it origins in the [insert(taskEntity,executionEntity) of the class TaskEntityManagerImpl.java](https://github.com/Activiti/Activiti/blob/eeda8439bcd30f525f10b23f54a7cbb4819d2ef9/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManagerImpl.java#L75) that is only being called in the [UserTaskActivityBehavior](https://github.com/Activiti/Activiti/blob/096fb2ceefa005dddf7b9a613dd0e0b1416d2981/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java#L212). There's another reference to this method in the [SaveTaskCmd.java](https://github.com/Activiti/Activiti/blob/407f917f216c03580cf99757c5eeab0d946bf5f2/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SaveTaskCmd.java#L54) but since the `executionEntity` is being passed as `null`, it's not really a problem. The issue happened when adding an execution id to a `taskEntity` from an execution that was already deleted.

balsarori: I removed the above changes to debug the test and better understand the issue. However, the test was green. The test doesn't seem to replicate the issue.

Additionally, can you remove any unrelated data/steps to the issue from the process definition? That can help us better understand debug and understand the issue.

antoniojfelix: Yes, there was a problem in the test. I was setting the variable "currentDistributionList" with `Arrays.asList("")` making it with the size 1. It should be with size 0 in order to replicate the problem, so I changed it to `Arrays.asList()` and now the test is fixed. I've also added two other tasks in the test (reviewTask and rejectionTask) to simulate the workflow. 
Also, I tried to simplify the process definition the best I could. But since this is a very specific case, I'm not sure how I can simplify it further. I hope that it is good enough. Thanks

balsarori: The `for` loop block doesn't seem to get executed. When the `currentDistributionList` list is empty the `else` block is always executed.

Also, I noticed that process has 2 sequences leading to the subprocess (2 parallel gateways after Distribution Task). Which could cause issues.

balsarori: Seems that the problem is due the fact user task insert is executed after the process `execution` is being deleted. I'm not sure why the insert is happening in the first place.

antoniojfelix: This was working in APS 1.11.4 that is using Activiti 5.23.0.1 and stopped working in APS 2.3.4 that is using Activiti 7.4.0. 
If the `nrOfInstances` is 0 (that only happens if the `currentDistributionList` is empty) in the MultiInstanceActivityBehavior.java it leaves the execution without executing anything. Therefore I added the `execution.hasInstances()` in this method `cleanupExecutions()` to avoid the execution record to be deleted from the database and to be linked with the task. That's what was being done in the previous version of the Activiti where the case was working. 
The problem was that the execution was being created in memory, was being associated with the task and then the execution was being deleted causing a database error when storing the task record in the database (because the execution was not created in the database).
To fix this, I compared both Activiti versions and after the fix I tested the process to check if workflow was the same between the APS 1.11.4 and APS 2.3.4.

balsarori: Seems parallel and inclusive gateways aren't properly used here. Parallel gateway having one outcome to the subprocess and another to inclusive gateway. And another outcome from parallel gateway to subprocess (only single outcome from a parallel gateway). Thats not how parallel gateways should be used. Please refer to https://www.activiti.org/userguide/#bpmnParallelGateway

balsarori:
An inclusive gateway can’t be used to join a parallel gateway. Replacing the inclusive gateway with parallel gateway would solve the issue.

PR #4235, opened 2023-02-02
fix/MNT-23416_boundary_event_sub_subprocess

vrmoreira:
…own by a subprocess within a subprocess

The end error event raised in simple subprocess is not propagating to parent call activity execution because of the same error id used in both processes. This PR fixes the edge case to try propagating the error event for call activity.

image

image


andrea-ligios (approved):
LGTM

erdemedeiros (approved):
LGTM!

Issue #4255, opened 2023-02-21
Support rollback at engine level


miguelruizdev:
In order to avoid conflicts and deployment of already existing deployments during a rollback, we need to support the retrieval of the currently enforced version deployment, and cleanup from the persistence layer the previously rolled back deployment, which will cause disruption for all those parts in the code we were fetch 'the latest deployment' as the current one.

PR #4256, opened 2023-02-21
#4255 - Support rollback at engine level

miguelruizdev:
Ref: #4255


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/RollbackEnvironment.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.activiti.engine.impl.cfg;
+
+public class RollbackEnvironment {
+
+    public static final String ACTIVITI_DEPLOY_AFTER_ROLLBACK_VARIABLE_NAME = "ACTIVITI_DEPLOY_AFTER_ROLLBACK";
+
+    public boolean isRollbackEnabled() {
+        return Boolean.parseBoolean(System.getenv(ACTIVITI_DEPLOY_AFTER_ROLLBACK_VARIABLE_NAME));
```
balsarori: I think this should be a primitive boolean (default false) with set/get. Reading the env var and setting the value should be from a spring property somewhere outside the core engine code (maybe activiti-cloud).

miguelruizdev: It makes sense, thanks for the feeback.
We can do it from the `SpringProcessEngineConfiguration`, from there we can inject the property via `@Value` annotation into the `ProcessEngineConfiguration`

balsarori: Yeah, `@Value` can be injected here [ApplicationUpgradeContextAutoConfiguration](https://github.com/Activiti/Activiti/blob/4255-support-rollback-engine-level/activiti-core-common/activiti-spring-project/src/main/java/org/activiti/core/common/spring/project/conf/ApplicationUpgradeContextAutoConfiguration.java#L38) to [ApplicationUpgradeContextService](https://github.com/Activiti/Activiti/blob/4255-support-rollback-engine-level/activiti-core-common/activiti-spring-project/src/main/java/org/activiti/core/common/spring/project/ApplicationUpgradeContextService.java) which `SpringProcessEngineConfiguration` can use to set the flag. Seems to me ApplicationUpgradeContextService is the right place for it as we are doing an upgrade here.

balsarori: `SpringProcessEngineConfiguration ` can read the property from `ApplicationUpgradeContextService` [here](https://github.com/Activiti/Activiti/blob/4255-support-rollback-engine-level/activiti-core/activiti-spring/src/main/java/org/activiti/spring/SpringProcessEngineConfiguration.java#L69)

miguelruizdev: alright, I will implement it using the `ApplicationUpgradeContextService`

activiti-core/activiti-spring/src/main/java/org/activiti/spring/SpringProcessEngineConfiguration.java
```diff
@@ -45,44 +40,42 @@
...
-  private DefaultAutoDeploymentStrategy defaultAutoDeploymentStrategy;
+    protected PlatformTransactionManager transactionManager;
+    protected String deploymentName = "SpringAutoDeployment";
+    protected Resource[] deploymentResources = new Resource[0];
+    protected String deploymentMode = "default";
+    protected ApplicationContext applicationContext;
+    protected Integer transactionSynchronizationAdapterOrder = null;
+    private Collection<AutoDeploymentStrategy> deploymentStrategies = new ArrayList<>();
+    private DefaultAutoDeploymentStrategy defaultAutoDeploymentStrategy;
+    private boolean isRollbackDeployment;
```
balsarori: this can be removed since its in the parent class

activiti-core/activiti-spring/src/main/java/org/activiti/spring/SpringProcessEngineConfiguration.java
```diff
@@ -181,6 +174,10 @@ public void setDeploymentMode(String deploymentMode) {
     this.deploymentMode = deploymentMode;
   }
 
+    public boolean isRollbackDeployment() {
```
balsarori: same here, can be removed since its in the parent class

balsarori (changes requested):
Will be good to add a setter for `isRollbackDeployment` in `ProcessEngineConfigurationImpl` as its a configuration class and should be possible to set configurations by using setters.

miguelruizdev:
> Will be good to add a setter for `isRollbackDeployment` in `ProcessEngineConfigurationImpl` as its a configuration class and should be possible to set configurations by using setters.

Addressed in latest commit

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/RollbackEnvironment.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.engine.impl.cfg;
+
+public class RollbackEnvironment {
```
balsarori: This class can be deleted. No longer used

balsarori (changes requested):
I wonder if some integration tests can be added. For example, deploy some process definitions with `isRollbackEnabled` false (maybe run some process instances) then set to true and deploy again (recheck the process instances). This help us better understand the impact of a rollback

miguelruizdev:
> I wonder if some integration tests can be added. For example, deploy some process definitions with `isRollbackEnabled` false (maybe run some process instances) then set to true and deploy again (recheck the process instances). This help us better understand the impact of a rollback

the flow to check if everything is 'business as usual' after a rollback is being tested in a e2e way, in the acceptance test enterprise repo (that I won't link here)
that being said, maybe I can come up with some test to add a bit of coverage at this level to guarantee compatibility with other usages beyond APA

balsarori (approved):
LGTM

Issue #4294, opened 2023-04-05
Expression resolving should be done once a variable is created 


balsarori:
When creating a process variable either initially from definition or from mapping, the expression should be resolved once. In case the expression isn't resolvable is should be replaced with `null`. The variable values should not contain the expression of the definition in case the expression couldn't be resolved.

balsarori:
Fixed in https://github.com/Activiti/Activiti/pull/4295

PR #4295, opened 2023-04-05
Expressions should be evaluated once a variable is created

balsarori:
- Updated `expressionResolver` to resolve expressions and not leave expressions unresolved. Resulting to no expressions being saved in variables.
- Fixed `Expressions are not allowed as variable values in the output mapping` not being thrown when `ALL` mapping type is used.
- Process initial variables are resolved before starting the process.
- Expressions in process instance variables value are no longer resolved.

Note while I don't think we still need the exception throwing when an expression text is submitted by user i.e. start process payload or as outbound variables of tasks/connectors, I left them as they are for now. In such cases submitted text will handled as simple text and never evaluated.

refs: #4294 , AAE-11481

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExpressionResolver.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExpressionResolver.java
@@ -114,9 +114,8 @@ public class ExpressionResolver {
             return expressionEvaluator.evaluate(expressionManager.createExpression(sourceString), expressionManager,
                 delegateInterceptor);
         } catch (final Exception e) {
-            logger.warn("Unable to resolve expression in variables, keeping original value",
-                        e);
-            return sourceString;
+            logger.warn("Unable to resolve expression in variables", e);
+            return null;
         }
     }
 
@@ -133,8 +132,8 @@ public class ExpressionResolver {
                 matcher.appendReplacement(sb,
                                           Objects.toString(value));
             } catch (final Exception e) {
-                logger.warn("Unable to resolve expression in variables, keeping original value",
-                            e);
+                logger.warn("Unable to resolve expression in variables", e);
+                matcher.appendReplacement(sb, "");
             }
         }
         matcher.appendTail(sb);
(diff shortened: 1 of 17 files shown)


activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExpressionResolver.java
```diff
@@ -114,9 +114,8 @@ private Object resolveObjectPlaceHolder(ExpressionEvaluator expressionEvaluator,
             return expressionEvaluator.evaluate(expressionManager.createExpression(sourceString), expressionManager,
                 delegateInterceptor);
         } catch (final Exception e) {
-            logger.warn("Unable to resolve expression in variables, keeping original value",
-                        e);
-            return sourceString;
+            logger.warn("Unable to resolve expression in variables", e);
+            return null;
```
mauriziovitale, line 118: we should never return null

balsarori: Not sure what else to return since the method returns an object that can't be resolved. `null` will be assigned to the variable with expression value that can't be resolved.

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java
```diff
@@ -166,15 +166,26 @@ private Map<String, Object> calculateOutPutVariables(MappingExecutionContext map
...
+                        new SimpleMapExpressionEvaluator(availableVariables),
+                        new VariableScopeExpressionEvaluator(mappingExecutionContext.getExecution())),
+                        outboundVariables);
+            }
+            return expressionResolver.resolveExpressionsMap(
+                new VariableScopeExpressionEvaluator(mappingExecutionContext.getExecution()), outboundVariables);
+        } else {
+            return expressionResolver.resolveExpressionsMap(
+                new SimpleMapExpressionEvaluator(availableVariables), outboundVariables);
         }
```
Giovanni007, line 180: Maybe we can extract some method here to make it more readable?

balsarori: I moved execution expression resolving to a different method https://github.com/Activiti/Activiti/pull/4295/commits/263b1c4808f2a17db3b0b7cd2878fb1ef69ac9d9

Giovanni007 (approved):

PR #4406, opened 2023-08-24
Create codeql.yml

ciairuzz:
custom workflow required

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -62,7 +62,7 @@ jobs:
       with:
           java-version: '17'
           distribution: 'temurin'
-
+    
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,97 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "develop", "5.x", "6.x", "7.*.x" ]
```
erdemedeiros: The branches `5.x` and `6.x` are no longer maintained, so I'm not sure they should be included here. In a similar way, only the latest branch of `7.*.x` is maintained, i.e. `7.11.x`,  so it should be enough `develop` and `7.11.x`.

erdemedeiros (commented):
See my comment about the branches coverage. LGTM otherwise.

erdemedeiros (approved):

PR #4413, opened 2023-09-06
MNT-23813: endEventListener of an userTask is triggered if attached b…

vrmoreira:
…oundaryEvent is executed


erdemedeiros (approved):
LGTM!

balsarori (approved):

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/timer/BoundaryTimerEventTest.java
```diff
@@ -54,6 +54,14 @@ public void notify(DelegateExecution execution) {
     }
   }
 
+    @Override
+    protected void setUp() throws Exception
+    {
```
balsarori: can you move the curly brace to the same line after the Exception? So that it matches the code style used

PR #4434, opened 2023-09-27
MNT-23965: added .getEngineServices() to execution to allow Activiti5…

vrmoreira:
… inflight cases

### DIFF
--- a/activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/DelegateExecution.java
+++ b/activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/DelegateExecution.java
@@ -21,7 +21,7 @@ import java.util.List;
 
 import org.activiti.bpmn.model.ActivitiListener;
 import org.activiti.bpmn.model.FlowElement;
-import org.activiti.engine.impl.cfg.ProcessEngineConfigurationImpl;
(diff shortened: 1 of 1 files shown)


activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/DelegateExecution.java
```diff
@@ -182,4 +184,8 @@ public interface DelegateExecution extends VariableScope {
    */
   void setMultiInstanceRoot(boolean isMultiInstanceRoot);
 
+  default ProcessEngineConfigurationImpl getEngineServices() {
```
balsarori: can we return `ProcessEngineConfiguration` instead and have the implementation classes implement it? We should not expose `impl` classes here.

balsarori (changes requested):

activiti-core/activiti-engine/src/main/java/org/activiti/engine/delegate/DelegateExecution.java
vrmoreira: You're right, I should be using `ProcessEngineConfiguration` instead.. noob mistake. Fixed.

balsarori (approved):
LGTM

PR #4444, opened 2023-10-09
AAE-16959 Extend IdentityLink Api with user identity details attribute

igdianov:
This PR extends IdentityLink api to be able to store and query implementation specific user identity details in the process and task entity contexts. 

Part of: https://alfresco.atlassian.net/browse/AAE-16959

### DIFF
--- a/activiti-core/activiti-engine/pom.xml
+++ b/activiti-core/activiti-engine/pom.xml
@@ -13,6 +13,10 @@
   <artifactId>activiti-engine</artifactId>
   <name>Activiti :: Engine</name>
 
+  <properties>
+    <activiti-engine.unpack.version>7.2.0</activiti-engine.unpack.version>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.activiti</groupId>
@@ -232,7 +236,7 @@
                 <artifactItem>
                   <groupId>org.activiti</groupId>
                   <artifactId>activiti-engine</artifactId>
-                  <version>7.2.0</version>
+                  <version>${activiti-engine.unpack.version}</version>
                   <overWrite>true</overWrite>
                   <outputDirectory>${project.build.directory}/activiti-engine</outputDirectory>
                 </artifactItem>
--- a/activiti-core/activiti-engine/pom.xml
+++ b/activiti-core/activiti-engine/pom.xml
@@ -177,6 +177,21 @@
       <artifactId>hibernate-core</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>postgresql</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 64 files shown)


balsarori:
I'm not sure it's a good idea to save such info in the engine level. The engine's concern should concentrate on process execution. Can't this logic and data live somewhere outside of the engine?

igdianov:
> I'm not sure it's a good idea to save such info in the engine level. The engine's concern should concentrate on process execution. Can't this logic and data live somewhere outside of the engine?

The info is needed to be attached to the process execution and task entities via identity links, so that we can generate audit events with actor details from identity service. The identity link is the best fit for it by extending the details attribute, so we don't have to introduce any dependency on Hibernate/JPA/Liquibase in the Activiti Cloud Process Runtime.

activiti-core/activiti-engine/pom.xml
```diff
@@ -201,6 +216,33 @@
...
+            <phase>compile</phase>
+            <goals>
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.activiti</groupId>
+                  <artifactId>activiti-engine</artifactId>
+                  <version>7.2.0</version>
```
graric: What do you think about using a variable with a meaningful name(e.g. activiti.previous.version)?

graric (approved):

activiti-core/activiti-engine/src/main/resources/org/activiti/db/upgrade/activiti.mysql55.upgradestep.800.to.810.history.sql
```diff
@@ -0,0 +1 @@
+alter table ACT_HI_IDENTITYLINK add column DETAILS_ LONGBLOB;
```
erdemedeiros, line 1: Missing migration from `mysql`? I only see `mysql55`

activiti-core/activiti-engine/src/test/java/org/activiti/engine/impl/db/DbSchemaUpdateTest.java
```diff
@@ -0,0 +1,82 @@
...
+        CommandExecutor commandExecutor = processEngine.getProcessEngineConfiguration().getCommandExecutor();
+        CommandConfig config = new CommandConfig().transactionNotSupported();
+
+        // and when
+        PropertyEntity schemaVersion = commandExecutor.execute(config,
+            commandContext -> commandContext.getDbSqlSession()
+                .selectById(PropertyEntity.class,"schema.version"));
+
+        // then
+        assertEquals(ProcessEngine.VERSION, schemaVersion.getValue());
```
erdemedeiros: Please, use `AssertJ` assertions

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/history/HistoricTaskInstanceTest.java
```diff
@@ -605,6 +604,62 @@ public void testHistoricIdentityLinksOnTask() throws Exception {
...
+                    assertThat(link.getDetails()).isEqualTo("details".getBytes());
+                    foundCustom = true;
+                }
+            }
+        }
+
+        assertThat(foundAssignee).isTrue();
+        assertThat(foundCandidateGroup).isTrue();
+        assertThat(foundCandidateUser).isTrue();
+        assertThat(foundCustom).isTrue();
```
erdemedeiros: All this block can be simplified to
```suggestion
        assertThat(historicIdentityLinks)
            .extracting(
                HistoricIdentityLink::getType,
                HistoricIdentityLink::getGroupId,
                HistoricIdentityLink::getUserId,
                HistoricIdentityLink::getDetails)
            .containsExactlyInAnyOrder(
                tuple("candidate", "sales", null, null),
                tuple("candidate", null, "fozzie", null),
                tuple("assignee", null, "kermit", null),
                tuple("actor", null, "gonzo", "details".getBytes())
            );
```

erdemedeiros (changes requested):
Apart from some small details, LGTM

activiti-core/activiti-engine/src/main/resources/org/activiti/db/upgrade/activiti.mysql55.upgradestep.800.to.810.history.sql
igdianov: Mysql updgrade exists here: https://github.com/Activiti/Activiti/pull/4444/files#diff-00b1b0e8c4ec776b678c0eea023a4e9faad83407da3057d95e46344d37323df3R1

erdemedeiros: The one you mentioned is `mssql`, not `mysql` :D

erdemedeiros (approved):

PR #3947, opened 2022-05-24
Handle Boundary Event Properly

AnkurChauhan-131:

PR #4459, opened 2023-10-24
MNT-22737: Handle Boundary Event Properly (#3947)

vrmoreira:


balsarori (approved):
LGTM. Although, I recommend removing extra noise from the process BPMN XML such as diagram info and other unneeded extensions.

balsarori (approved):

balsarori (approved):

PR #4474, opened 2023-11-13
MNT-23915: errors thrown in called activities get caught by parent process

vrmoreira:


balsarori:
Can you update the title of the PR?

Is this a regression? Does Activiti 5 behave similarly ?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -60,7 +60,7 @@ public static void propagateError(String errorRef, DelegateExecution execution)
             isCatchExecutedForProcess = executeCatch(eventMap, execution, errorRef);
         }
 
-        if (isCallActivity(execution)) {
+        if (isCallActivity(execution) && !isCatchExecutedForProcess) {
```
balsarori: We don't need to call `isCallActivity` if `isCatchExecutedForProcess` is `true`

```suggestion
        if (!isCatchExecutedForProcess && isCallActivity(execution)) {
```

vrmoreira:
> Can you update the title of the PR?

Done

> 
> Is this a regression? Does Activiti 5 behave similarly ?

Activiti 5 behaves correctly: I've replicated the same childProcess (sub-process, boundary-event on sub-process, user-task after boundary-event), changed the unit-test (to handle the new user-task) and the event is caught on the child-process (correct behaviour).

Analyzing the code, if there are local eventHandlers, they get executed and the that's it. If there aren't any, continues to search on super execution.

Source code: https://github.com/Activiti/Activiti/blob/5.x/modules/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java#L66

erdemedeiros (approved):
Good catch! LGTM!

PR #4485, opened 2023-11-21
AAE-18108 Use SHA for GitHub actions

Giovanni007:


.github/dependabot.yml
```diff
@@ -5,9 +5,14 @@ updates:
...
+      interval: "weekly"
 
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+
+  - package-ecosystem: "github-actions"
+    directory: "/.github/actions/check-ext-build"
```
atchertchian, line 16: This action does not reference any other action so it is not needed to add it here (but it does not hurt either)

atchertchian (approved):

.github/dependabot.yml
Giovanni007: you are right. I'm adding this kind of configuration for all the actions for 2 reason:

1. if we modify it in future adding reference to other actions, we won't have to remember to add the dependabot configuration
2. It's was faster for me adding all the actions in the configuration then checking one by one which one is referencing other actions

I guess this should not be a big issue. But if you think this can have any negative side effect on dependabot, let me know and I will review the configuration.

atchertchian: It's not a problem, it's safer to keep it as is for the reasons you mentioned, ideally we should have the same logic for checking this than on https://github.com/Alfresco/alfresco-build-tools as it's also checked as part of pre-commit.
And I thought this script checked if adding the action was needed or not, but I'm wrong, so you can disregard this comment :+1:

PR #4506, opened 2023-12-06
MNT-24080: cache is updated if the new processDefinition is for the s…

vrmoreira:
…ame activiti version


vrmoreira:
@erdemedeiros is right, updating the cache based on the class of the already-existing item is weird. 

Anyway, because of one of the APS PR (https://github.com/Alfresco/activiti-bpm-suite/pull/5950) fixes the execution engine to be used (Activiti-5 instead of Activiti-7), there's no need to change Activiti-7 codebase.

PR #4529, opened 2024-01-12
[AAE-19314] New CI workflows to allow external contributions

graric:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

We split `main.yml` into `main_push.yml` and `main_pull.yml` which handle, respectively, pushes on the `develop` branch and events on PR branches against the `develop` branch.

For PRs opened by external contributors (i.e. coming from forks) we do not use any secrets in the workflow and the "Preview" label is not available.

We ran several examples with different configurations against a dummy release branch:

- PR from external contributor (from fork) with org access: https://github.com/Activiti/Activiti/pull/4535
- PR from external contributor (from fork) without org access: https://github.com/Activiti/Activiti/pull/4536
- Push on release branch: https://github.com/Activiti/Activiti/actions/runs/7529909194

- Issue Link: https://alfresco.atlassian.net/browse/AAE-19314

> - [ ] Yes
> - [x] No

### DIFF
--- a/.github/workflows/main_pull.yml
+++ b/.github/workflows/main_pull.yml
@@ -1,0 +1,106 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - develop
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+
+  # Allows to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+
+  pre-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Check dependabot build
+        uses: ./.github/actions/check-ext-build
+      - name: pre-commit
+        uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit@f2259c2bb5ff4a5579d1a425cb130175a6459665 # v5.10.0
+        with:
+          skip_checkout: true
+      - name: Ensure SHA pinned actions
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@ba37328d4ea95eaf8b3bd6c6cef308f709a5f2ec # v3.0.3
+
+  build:
+    runs-on: ubuntu-latest
+    needs: pre-checks
+    steps:
+
+      - name: Checkout repository
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          token: ${{ secrets.BOT_GITHUB_TOKEN }}
+
+      - uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Setup Java JDK 21
+        uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # 4.0.0
+        with:
+          java-version: 21
+          distribution: 'temurin'
+
+      - name: Set preview version
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'preview') }}
+        run: |
+          GITHUB_PR_NUMBER=PR-${{ github.event.pull_request.number }}
+          echo 0.0.1-${GITHUB_PR_NUMBER}-${GITHUB_RUN_NUMBER}-SNAPSHOT > VERSION
+
+      - name: Set VERSION env variable
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'preview') }}
+        run: |
+          VERSION=$(cat VERSION)
+          echo set VERSION=$VERSION
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+
+      - name: Update pom files to the new version
+        if: ${{ contains(github.head_ref, 'preview') || contains(github.event.pull_request.labels.*.name, 'preview') }}
+        run: mvn -B versions:set -DnewVersion=$VERSION -DprocessAllModules=true -DgenerateBackupPoms=false
+
+      - name: Define Maven Command
+        id: define_maven_command
+        shell: bash
+        run: |
+          if [ $DO_PUSH == 'true' ]
+          then
+            echo "command=deploy" >> $GITHUB_OUTPUT
+          else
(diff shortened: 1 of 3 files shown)


.github/workflows/main_pull.yml
```diff
@@ -0,0 +1,106 @@
+name: CI
```
erdemedeiros: Let's find a specific name for this: `PR`, `Validate PR`, `Build PR` ?

.github/workflows/main_pull.yml
```diff
@@ -0,0 +1,106 @@
...
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+
+  # Allows to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name || github.run_id }}
```
erdemedeiros: The expression can be simplified. As this is running on  pull requests only, `github.head_ref` is always set and there is no need for fallbacks:
```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref }}
```

.github/workflows/main_pull.yml
```diff
@@ -0,0 +1,106 @@
...
+          MAVEN_CLI_OPTS: --show-version --no-transfer-progress --settings settings.xml
+          MAVEN_USERNAME: ${{ secrets.NEXUS_USERNAME }}
+          MAVEN_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+
+      - name: Build and Test with Maven for external contributors
+        if: github.event.pull_request.head.repo.full_name != 'Activiti/Activiti'
+        shell: bash
+        run: mvn verify ${{ env.MAVEN_CLI_OPTS}}
+        env:
+          MAVEN_CLI_OPTS: --show-version --no-transfer-progress --settings settings.xml
```
erdemedeiros: I'm not sure this distinction is needed. I think you can combine the two steps in a single one (the one that is setting maven credentials). In the case of a fork, the secrets will not be available, but it should not impact the build because we are not going to deploy in this case (preview not supported) and, hence, it will not try to use the credentials. Could you give it a try, please?

.github/workflows/main_push.yml
```diff
@@ -1,17 +1,11 @@
 name: CI
 
+permissions: write-all
```
erdemedeiros: Maybe a bit too permissive here. It's better to set only the strict needed permissions.

.github/workflows/main_push.yml
```diff
@@ -60,46 +54,20 @@ jobs:
 
       - uses: Alfresco/alfresco-build-tools/.github/actions/update-pom-to-next-pre-release@f2259c2bb5ff4a5579d1a425cb130175a6459665 # v5.10.0
```
erdemedeiros, line 56: The step: `Check dependabot build` can be removed as this is targeting push only

.github/workflows/main_push.yml
```diff
@@ -110,34 +78,29 @@ jobs:
         run: find . -name TEST-*.xml -exec grep -h testcase {} \; | awk -F '"' '{printf("%s#%s() - %.3f s\n", $4, $2, $6); }' | sort -n -k 3 | tail -20
```
erdemedeiros: It's maybe worth to extract this code inside an action to avoid duplicating this complex expression

.github/workflows/main_push.yml
```diff
@@ -1,17 +1,11 @@
 name: CI
```
erdemedeiros: Let's rename the workflow to have a unique name, please

erdemedeiros (changes requested):
Great job!
Here are some suggestions of improvements:

.github/workflows/main_pull.yml
matteo4diani: If a requested secret is not found the job will fail

.github/workflows/main_push.yml
```diff
@@ -1,17 +1,11 @@
...
-    branches:
-      - develop
-    types:
-      - opened
-      - reopened
-      - synchronize
-      - labeled
 
   # Allows to run this workflow manually from the Actions tab
   workflow_dispatch:
```
erdemedeiros, line 14: The concurrency group can be simplified to ` ${{ github.workflow }}-${{ github.ref_name }}`

.github/workflows/main_pull.yml
matteo4diani: That is why we splitted it, and that is why it was failing before


ETA: maybe I misunderstood the original issue, we'll try going this way

.github/workflows/main_push.yml
matteo4diani: What would be for you the strictly needed permissions?

.github/workflows/main_push.yml
matteo4diani: not even the fallback on `github.run-id`?

matteo4diani: https://github.com/Activiti/Activiti/pull/4529/commits/d5f96261875a6604378c59939a4d890b91061612

.github/workflows/main_push.yml
matteo4diani: https://github.com/Activiti/Activiti/pull/4529/commits/d5f96261875a6604378c59939a4d890b91061612

.github/workflows/main_push.yml
matteo4diani: https://github.com/Activiti/Activiti/pull/4529/commits/d5f96261875a6604378c59939a4d890b91061612

.github/workflows/main_pull.yml
matteo4diani: https://github.com/Activiti/Activiti/pull/4529/commits/d5f96261875a6604378c59939a4d890b91061612

.github/workflows/main_pull.yml
matteo4diani: https://github.com/Activiti/Activiti/pull/4529/commits/d5f96261875a6604378c59939a4d890b91061612

.github/workflows/main_pull.yml
matteo4diani: Done in https://github.com/Activiti/Activiti/pull/4529/commits/182962f4f6fa24436652db1c0d526d15c3209fda

.github/workflows/main_push.yml
matteo4diani: Done in https://github.com/Activiti/Activiti/pull/4529/commits/d464bc01a1c4b968523c5e2b43d9f1f878314bd1

erdemedeiros (approved):
LGTM!

.github/workflows/main_push.yml
matteo4diani: Done in https://github.com/Activiti/Activiti/pull/4529/commits/26f530c995e018031d562fa585dd8b65c12c7de0

PR #4596, opened 2024-02-21
MNT-24219: local variables before calling subprocess remains after en…

vrmoreira:
…d of subprocess

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://alfresco.atlassian.net/browse/MNT-24219

> - [ ] Yes
> - [X] No

### DIFF
--- a/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/EndExecutionOperation.java
+++ b/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/EndExecutionOperation.java
@@ -30,6 +30,7 @@ import org.activiti.engine.delegate.event.ActivitiEventType;
 import org.activiti.engine.delegate.event.impl.ActivitiEventBuilder;
 import org.activiti.engine.impl.bpmn.behavior.MultiInstanceActivityBehavior;
 import org.activiti.engine.impl.bpmn.helper.ScopeUtil;
+import org.activiti.engine.impl.bpmn.helper.SubProcessVariableSnapshotter;
 import org.activiti.engine.impl.context.Context;
 import org.activiti.engine.impl.delegate.ActivityBehavior;
 import org.activiti.engine.impl.delegate.SubProcessActivityBehavior;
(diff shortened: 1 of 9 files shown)


vrmoreira:
Tried to make the fix as clear as possible.
Anyway, think this diagram might shed some light. If this solution is approved, then I'll update the JIRA ticket
diagram

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/EndExecutionOperation.java
```diff
@@ -229,6 +231,21 @@ protected ExecutionEntity handleSubProcessEnd(ExecutionEntityManager executionEn
...
+        // {@code executionEntityManager.findChildExecutionsByParentExecutionId()} will not return that execution
+        Optional<? extends ExecutionEntity> first = parentExecution.getParent().getExecutions().stream()
+            .filter(executionEntity -> {
+                // scoped execution of the current subprocess should not be considered
+                return !executionEntity.getId().equals(parentExecution.getId());
+            } )
+            .findFirst();
+        if( first.isPresent() ) {
+            new SubProcessVariableSnapshotter().setVariablesSnapshots(first.get(), executionToContinue);
+        }
```
erdemedeiros: This logic only works for some specific cases, but it cannot work in general: `parentExecution.getParent().getExecutions()` may return several executions depending on the process design and peeking the first one could return the wrong one. Also, there is no guarantee that the execution we are looking for is still in the database at this point, as it's marked to be deleted in [ContinueProcessOperation.createChildExecutionForSubProcess](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/ContinueProcessOperation.java#L142) and depending on the process design it may or may not be there. For instance, if the subprocess has a user task or any flow node marked as async, the execution will be already deleted from the database when we reach this point.

erdemedeiros (changes requested):

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/agenda/EndExecutionOperation.java
vrmoreira: I've improved the fetch of the execution but it fails when a UserTask is inside of the subprocess: the execution was already deleted. 
Pushed modifications in order to continue the discussion

PR #4600, opened 2024-02-27
AAE-20402 Fix MAP_ALL variables mapping

Giovanni007:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

Right now, when using MAP_ALL  with json variables, if the variable doesn't contain any expression, it doesn't get converted in a Map and remain of type JsonNode.

- Issue Link: https://alfresco.atlassian.net/browse/AAE-20402

> - [ ] Yes
> - [ ] No

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/test/java/org/activiti/spring/process/ProcessVariablesInitiatorIT.java
+++ b/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/test/java/org/activiti/spring/process/ProcessVariablesInitiatorIT.java
@@ -17,7 +17,9 @@ package org.activiti.spring.process;
 
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonMap;
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+import static org.assertj.core.api.Assertions.entry;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.mock;
 
@@ -166,7 +168,7 @@ public class ProcessVariablesInitiatorIT {
     }
 
     @Test
-    public void calculateVariablesFromExtensionFileShouldReturnMapWhen() throws Exception {
+    public void calculateVariablesFromExtensionFileShouldReturnMapWhenVariableIsJson() throws Exception {
         try (InputStream inputStream = Thread.currentThread().getContextClassLoader().getResourceAsStream("processes/variable-mapping-all-extensions.json")) {
 
             ProcessExtensionModel extension = reader.read(inputStream);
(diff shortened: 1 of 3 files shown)


Giovanni007:
I'm not sure if this is the right way to fix this. Some points that can be useful to understand the issue:

1. When we read the extension all the Json variable are represented as JsonNode by purpose https://github.com/Activiti/Activiti/blob/eeda8439bcd30f525f10b23f54a7cbb4819d2ef9/activiti-core/activiti-spring-process-extensions/src/main/java/org/activiti/spring/process/ProcessExtensionResourceReader.java#L57
2. I'm not sure if the conversion  from JsonNode to map , for variables that do NOT contain expression should happen. If  when we init process as I did or If  where we check the inputVariables and the mapping type https://github.com/Activiti/Activiti/blob/263b1c4808f2a17db3b0b7cd2878fb1ef69ac9d9/activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExtensionsVariablesMappingProvider.java#L75
3. This old fix is related https://github.com/Activiti/Activiti/pull/4295/files
4. In tests, when reading test resources directly with ObjectMapping we loose the conversion  we do in point 1 and that is executed in real world. I fixed it for the test class I used but we should keep this in mind in the future.

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/test/java/org/activiti/spring/process/ProcessVariablesInitiatorIT.java
```diff
@@ -17,15 +17,14 @@
 
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonMap;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.catchThrowable;
+import static org.assertj.core.api.Assertions.*;
```
wojciech-piotrowiak: is this correct to add import in asterisk ?
Not sure which settings are set in activiti

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/test/java/org/activiti/spring/process/ProcessVariablesInitiatorIT.java
```diff
@@ -169,4 +164,31 @@ public void calculateVariablesFromExtensionFileShouldThrowExceptionWhenProvidedV
                     .hasMessageContaining("as variables fail type validation - age");
         }
     }
+
+    @Test
+    public void calculateVariablesFromExtensionFileShouldReturnMapWhen() throws Exception {
```
wojciech-piotrowiak: is there missing name on the end?

wojciech-piotrowiak (approved):

PR #4701, opened 2024-06-27
Activiti-5195 : Update Email2-Jakarta and Core Version

AnkurChauhan-131:
Update Email2-Jakarta and Core


balsarori (approved):

balsarori:
fixed in https://github.com/Activiti/Activiti/pull/4696

PR #4702, opened 2024-06-28
MNT-24335: improve historical tasks performance

vrmoreira:
**What kind of change does this PR introduce?**

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Code style update (formatting, local variables)
> - [X] Refactoring (no functional changes, no api changes)
> - [ ] Build related changes
> - [ ] Documentation
> - [ ] Other... Please describe:

## Description

- Issue Link: https://hyland.atlassian.net/browse/MNT-24335

**Does this PR introduce a breaking change?** (check one with "x")

> - [ ] Yes
> - [X] No

PR #4716, opened 2024-07-11
MNT-24335: improve historical tasks performance (#4702)

vrmoreira:
(cherry picked from commit 84b6e253bc3631e45769c2750edbf939949def42)

> - [ ] Bugfix
> - [ ] Feature
> - [X] Refactoring (no functional changes, no api changes)
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/MNT-24335

> - [ ] Yes
> - [X] No


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
```diff
@@ -146,6 +146,7 @@ public class DbSqlSession implements Session {
```
balsarori: We'll need to add `7.11.1`
`ACTIVITI_VERSIONS.add(new ActivitiVersion("7.11.1"));`
before 
`ACTIVITI_VERSIONS.add(new ActivitiVersion(LAST_V7_VERSION));`

Also need to add the update steps

balsarori (changes requested):
We'll need this commit to fix native query issue if still exists --> https://github.com/Activiti/Activiti/pull/4675/commits/ff69ebc4df9940981bbb8b6751f9f1ef68b9159c

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/db/DbSqlSession.java
vrmoreira: In order to make sure that update (indexes) are run if we upgrade from a previous Activiti 8.x version, I've added schema version 8.1.1 and add the steps. This won't stop to re-execute the addition of those indexes if we upgrade from the latest Activiti 7.11.x version.

I'll add the steps for 7.11.1 version but need to verify if there's any problem executing those steps twice.

vrmoreira: Found a couple of problems, which think I've nailed it.
Also, on "testcontainers" branch, I've updated the `DbSchemaUpdateTest` in order to test it against all databases. I couldn't see any problems but, locally, on certain databases, if I change the updatestep for 8.1.1 schema by adding an error, everything runs ok, which is weird.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java
```diff
@@ -1146,18 +1147,31 @@ public void initDatabaseType() {
...
+      databaseType = databaseVendor;
+
+      switch (databaseVendor) {
+          case DATABASE_TYPE_MYSQL:
+              String databaseProductVersion = databaseMetaData.getDatabaseProductVersion();
+              // MariaDB has performance penalty when using "withoutJoins" SQL scripts
+              // example of MariaDB 10.5.4
+              //     input: vendor=MySQL, version=5.5.5-10.5.24-MariaDB-1:10.5.24+maria~ubu2004
+              //     output: databaseType=mariadb
+              if (databaseProductVersion.contains("MariaDB")) {
```
balsarori: maybe we should do a case insensitive comparison?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java
```diff
@@ -1146,18 +1147,31 @@ public void initDatabaseType() {
...
+      databaseType = databaseVendor;
+
+      switch (databaseVendor) {
+          case DATABASE_TYPE_MYSQL:
+              String databaseProductVersion = databaseMetaData.getDatabaseProductVersion();
+              // MariaDB has performance penalty when using "withoutJoins" SQL scripts
+              // example of MariaDB 10.5.4
+              //     input: vendor=MySQL, version=5.5.5-10.5.24-MariaDB-1:10.5.24+maria~ubu2004
+              //     output: databaseType=mariadb
+              if (databaseProductVersion.toLowerCase().contains("mariadb")) {
```
balsarori:
```suggestion
              if (databaseProductVersion.toLowerCase().contains(DATABASE_TYPE_MARIADB)) {
```

balsarori (approved):

PR #4725, opened 2024-07-26
AAE-24223 Log CommandContext exceptions as warning if app is in shutdown

jesty:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

Log the exception as warning if it happens during a shutdown.

- Issue Link:

https://hyland.atlassian.net/browse/AAE-24223

> - [ ] Yes
> - [x] No

### DIFF
--- a/activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/boot/ShutdownListener.java
+++ b/activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/boot/ShutdownListener.java
@@ -20,16 +20,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.event.ContextClosedEvent;
-import org.springframework.stereotype.Component;
 
-@Component
 public class ShutdownListener implements ApplicationListener<ContextClosedEvent> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ShutdownListener.class);
-
-    ShutdownListener() {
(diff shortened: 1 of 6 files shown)


activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/boot/ShutdownListener.java
```diff
@@ -0,0 +1,41 @@
...
+package org.activiti.spring.boot;
+
+import org.activiti.engine.ApplicationStatusHolder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.stereotype.Component;
+
+@Component
```
erdemedeiros: It's preferable to declare it as a `Bean` inside an auto-configuration to make sure it will get instantiated in other projects declaring `activiti-spring-boot-starter` as a dependency.  When declared as `@Component`, it may be out of scope of scanned packages in the target application, meaning that they will not be instantiated.
You can declare as a bean inside [ProcessEngineAutoConfiguration](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-spring-boot-starter/src/main/java/org/activiti/spring/boot/ProcessEngineAutoConfiguration.java)

erdemedeiros (changes requested):
I added one small comment related to the bean initialization. LGTM, otherwise

erdemedeiros (approved):
LGTM

PR #4761, opened 2024-08-21
Correcting the oracle adding column script for develop branch

moberoi2:
In oracle upgrade script we have a "column" keyword which is not required in Oracle.
Earlier it was
alter table ACT_RU_IDENTITYLINK add column DETAILS_ BLOB;
now it is
alter table ACT_RU_IDENTITYLINK add DETAILS_ BLOB;


balsarori (approved):

Agither (commented):
abc

PR #4811, opened 2024-10-21
MNT-24219: cant resolve identifier if theres a subprocess (solution2)

vrmoreira:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/MNT-24219

> - [ ] Yes
> - [X] No


vrmoreira:
To whom is going to review this PR:
- The goal of this PR is to evaluate if this is a viable solution to the problem:
1. Before entering a subprocess, all local variables are stored on the ACT_HI_VARINST table.
2. Subprocess execution will have a new local variable with the executionId of it's parent
3. When subprocess ends and the continuation execution is created, it will fetch the executionId of it's parent and restores all the local variables of the parent's execution

- it's not complete, there are quite a few problems that need to be resolved:
1. When saving the local variables, events are triggered when they shouldn't trigger
2. If a subprocess doesn't split into transactions (async, usertask inside, etc), local variables aren't stored on the DB

If the solution is admissable, I'll improve it and solve these problems. Thanks

PR #4898, opened 2025-01-28
MNT-24839: task name set by task listener don't show in interfaces behavior change from aps 1.11 to 2.4.3

vrmoreira:
**Notes**
* This implementation adds the update on the HistoricalTasks when there's a change on the name of the task at hand. It's possible to move that code to the higher abstraction layer (updating the entire `TaskEntity` on the database) but I thought that should have a bigger impact on performance than the beneficts from this fix
* There are other places where HistoricTasks are being updated (`SaveTaskCmd`) but I haven't refactored all that much, since there are logic behind the AUDIT HistoricLevel and some other logic which doesn't depend on it. Therefore, I consider that to be out-of-scope
* If necessary, I can raise a new issue to refactor that.

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/MNT-24839

> - [ ] Yes
> - [X] No


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManager.java
```diff
@@ -59,4 +61,7 @@ public interface TaskEntityManager extends EntityManager<TaskEntity> {
 
   void deleteTask(TaskEntity task, String deleteReason, boolean cascade, boolean cancel);
 
+  TaskInfo getHistoricTask(TaskEntity task);
```
balsarori: TaskEntityManager should be responsible for retrieving historical tasks.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SaveTaskCmd.java
```diff
@@ -61,15 +61,7 @@ public Task execute(CommandContext commandContext) {
       }
 
     } else {
-
```
balsarori, line 64: Probably the logic should remain in SaveTaskCmd similar to SetTaskPriorityCmd and SetTaskDueDateCmd.
What happens if the listener changes the due date, priority, or category of a task?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManager.java
vrmoreira: Sorry but I don't understand: it's the `TaskEntityManager` that's responsible for retrieving historical tasks.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SaveTaskCmd.java
vrmoreira: Two good questions, hope that I can provide two great answers
**SaveTaskCmd**
* `SaveTaskCmd` has the same logic, only changed the way we fetch the historical task (now on the `TaskEntityManager`). If the comment was to make `SaveTaskCmd` more like `SetTaskPriorityCmd`, think we need to compare with historical task since there are a lot of fields to check/update. 
* The lesser fields we need to update, the better.
* Also, seems like each `*.record*` method checks inside for the `AUDIT` level, therefore we can ditch that check on this class.

**Other operations**
I've only fixed what was described in the ticket since there are a lot fields to cover. Which fields should I consider? (I'm going to change the ones on your comment)

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/persistence/entity/TaskEntityManager.java
balsarori: I meant shouldn't :)

vrmoreira: Ah ok! Maybe on a higher abstraction level.. Or I just do what you said, don't refactor that code and use directly `HistoryManager`, without the getHistoricalTask. Do you agree?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/cmd/SaveTaskCmd.java
balsarori: A runtime task could be retrieved and modified (multiple modifications such as name, due date, priority, etc) and then saved using `taskService.save` which will call the `SaveTaskCmd` to save. Any changes should be captured.

vrmoreira: I've got a few questions from my solution which I would like to discuss with you @balsarori , namely the `*UpdaterHelper` classes

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/usertask/MakeTaskChangesUsingScriptsTests.java
```diff
@@ -0,0 +1,89 @@
...
+import org.activiti.engine.runtime.ProcessInstance;
+import org.activiti.engine.task.Task;
+import org.activiti.engine.test.Deployment;
+
+import java.util.Date;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class MakeTaskChangesUsingScriptsTests extends PluggableActivitiTestCase {
```
balsarori: how about `ScriptTaskListenerChangesTest` for the name?

balsarori (approved):
LGTM

PR #4939, opened 2025-03-11
AAE-31361 Set Previous Value To Null While Sending UpdateEvent If Variable Is Ephemeral

revatilimaye:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/AAE-31361

> - [ ] Yes
> - [x] No

### DIFF
--- a/activiti-core/activiti-api-impl/activiti-api-runtime-shared-impl/src/main/java/org/activiti/runtime/api/event/impl/ToVariableUpdatedConverter.java
+++ b/activiti-core/activiti-api-impl/activiti-api-runtime-shared-impl/src/main/java/org/activiti/runtime/api/event/impl/ToVariableUpdatedConverter.java
@@ -39,10 +39,8 @@ public class ToVariableUpdatedConverter implements EventConverter<VariableUpdate
 
         VariableInstanceImpl<Object> variableInstance = createVariableInstance(internalEvent, isEphemeral);
 
-        if(isEphemeral) {
-            ((ActivitiVariableUpdatedEventImpl) internalEvent).setVariablePreviousValue(null);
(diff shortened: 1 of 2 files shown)


activiti-core/activiti-api-impl/activiti-api-runtime-shared-impl/src/main/java/org/activiti/runtime/api/event/impl/ToVariableUpdatedConverter.java
```diff
@@ -38,6 +39,10 @@ public Optional<VariableUpdatedEvent> from(ActivitiVariableUpdatedEvent internal
 
         VariableInstanceImpl<Object> variableInstance = createVariableInstance(internalEvent, isEphemeral);
 
+        if(isEphemeral) {
+            ((ActivitiVariableUpdatedEventImpl) internalEvent).setVariablePreviousValue(null);
+        }
+
         return Optional.of(new VariableUpdatedEventImpl<>(variableInstance, internalEvent.getVariablePreviousValue()));
```
erdemedeiros: I think it's better to use temporary variable to store the value to be passed as previous value, this will avoid modifying the original internal event and also avoid the usage of casting.

```suggestion
       Object previousValue = isEphemeral? null : internalEvent.getVariablePreviousValue();
        return Optional.of(new VariableUpdatedEventImpl<>(variableInstance, previousValue));
```

erdemedeiros (changes requested):

erdemedeiros (approved):

PR #5002, opened 2025-04-30
Fix XSS vulnerability in quote method by escaping HTML special character

simei2k:

> - [ x] Bugfix
> - [ ] Feature
> - [x] Build related changes
> - [ ] Documentation

## Description
This PR addresses a potential XSS vulnerability in the Activiti/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/json/JSONObject.java" file. The method previously did not escape HTML special characters such as <, >, &, etc., which could allow malicious input to inject HTML or JavaScript into JSON strings.

- Issue Link:
This issue was also documented in [https://github.com/micromata/projectforge-webapp/commit/5a6a25366491443b76e528a04a9e4ba26f08a83c](url) and fixed in the Git Commit.

> - [ ] Yes
> - [ x] No

### DIFF
--- a/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/json/JSONObject.java
+++ b/activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/json/JSONObject.java
@@ -1035,7 +1035,7 @@ public class JSONObject {
    *          A String
    * @return A String correctly formatted for insertion in a JSON text.
    */
-  public static String quote(String string) {
+  public static String quote(String string, boolean escapeHtml) {
     if (string == null || string.length() == 0) {
       return "\"\"";
     }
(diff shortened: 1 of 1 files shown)


NileshBhatt:
- Consider reviewing the requirements for HTML escaping to ensure that all potentially problematic characters are handled. Additionally, it might be helpful to document the reasoning behind the choice of characters being escaped for clarity and future maintainability.

- The addition of the escapeHtml parameter to the quote method introduces valuable functionality for escaping HTML characters when needed. However, changing the method signature directly could break backward compatibility for existing code that relies on the original quote(String string) method. To maintain backward compatibility, consider overloading the method instead of modifying the existing signature.

P.S. I am new to OpenSource community.

PR #5032, opened 2025-06-03
AAE-34880 Update runtime to evaluate variables

revatilimaye:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/AAE-34880

> - [ ] Yes
> - [x] No

### DIFF
--- a/activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/ELContextBuilder.java
+++ b/activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/ELContextBuilder.java
@@ -17,6 +17,8 @@ package org.activiti.core.el;
 
 import static org.activiti.core.el.DateResolverHelper.addDateFunctions;
 import static org.activiti.core.el.ListResolverHelper.addListFunctions;
+
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -37,11 +39,14 @@ public class ELContextBuilder {
 
     private static final Logger logger = LoggerFactory.getLogger(ELContextBuilder.class);
 
-    private List<ELResolver> resolvers;
+    private List<ELResolver> resolvers ;
     private Map<String, Object> variables;
 
     public ELContextBuilder withResolvers(ELResolver... resolvers) {
-        this.resolvers = List.of(resolvers);
+        if (this.resolvers == null) {
+            this.resolvers = new ArrayList<>();
+        }
+        this.resolvers.addAll(List.of(resolvers));
         return this;
     }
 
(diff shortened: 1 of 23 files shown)


activiti-core/activiti-spring-boot-starter/pom.xml
```diff
@@ -10,6 +10,9 @@
   </parent>
   <artifactId>activiti-spring-boot-starter</artifactId>
   <name>Activiti :: Spring Boot Starter</name>
+  <properties>
```
erdemedeiros: Please, declare this property in the root pom to avoid the need of declaring it in several places and getting them miss aligned.

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/ELContextBuilder.java
```diff
@@ -52,6 +52,7 @@ public ELContextBuilder withVariables(Map<String, Object> variables) {
 
     public ELContext build() {
         CompositeELResolver elResolver = createCompositeResolver();
+        elResolver.add(new EnvironmentVariableELResolver());
```
erdemedeiros: I order to make it easier to modify in the future and to make it possible to add resolvers from outside (Open/Closed principle),  I think it's better add an extension point to provide extra resolvers. You could then, move `EnvironmentVariableELResolver` inside `activiti-spring-boot-starter` and instantiate it as a bean conditional on a property (`spring.activiti.env-var-el-resolver.enabled`) that is disabled by default to avoid potential conflicts on processes that have a process variable named `vars`. 

Have a look on how it was done for `CustomFunctionProvider`: you can have something similar for resolvers (`public ELContextBuilder withCustomResolvers (List<ELResolver> customResolvers)`).

activiti-core-common/activiti-expression-language/src/main/java/org/activiti/core/el/EnvironmentVariableELResolver.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public Object getValue(ELContext context, Object base, Object property) {
+        if (base == null && VAR_PREFIX.equals(property)) {
+            context.setPropertyResolved(true);
+            return new EnvVar();
+        }
+        if (base instanceof EnvVar && property != null) {
+            String env = System.getenv(VAR_PREFIX_WITH_DOT + property);
+            if(env != null) {
```
erdemedeiros: I think it should always set the property as resolved, even when the env is null or not set. Said it differently, I think you can remove this condition.

erdemedeiros (changes requested):
Apart from the refactoring to allow injecting extra resolvers, it looks good to me. Good job!

erdemedeiros (approved):
LGTM!

PR #5098, opened 2025-08-01
ACTIVITI-5503: added candidateUser/candidateGroups to orQuery parameters

vrmoreira:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/ACTIVITI-5503

> - [ ] Yes
> - [X] No


activiti-core/activiti-engine/src/main/resources/org/activiti/db/mapping/entity/HistoricTaskInstance.xml
```diff
@@ -1903,4 +1958,17 @@
     )
   </sql>
 
+  <sql id="withoutJoins_selectExistsNotIn">
+    exists (
+      select 1
+      from ${externalTable}
+      where
+        ${join}
+        and ${field} not in
```
CaptainGlac1er, line 1967: `not in` would result in the indexes being probably ignored

vrmoreira: Nice catch! Since there are quite a few places where we're using that, I would propose to open an issue just for that. That way, we have an action-item to study this, make the changes and test it.

CaptainGlac1er: @vrmoreira  thanks for the response. I'll do that. Side note I had an idea for FTS in the query side if you don't mind me pinging you about it.  https://github.com/Activiti/activiti-cloud/pull/1948

CaptainGlac1er: Okay I stand corrected. I just skimmed the cases of NOT IN they all appear to be related to explicit NOT IN queries so it basically can't be avoided. Then I realized that this was a case of that too.

vrmoreira: @CaptainGlac1er, thanks for your info. Indeed, the "not in" is only used for those cases - glad that's one less thing to do! Regarding your changes for the FTS on activiti-cloud, I must be honest and say that I have some knowledge on the activiti project but have no knowledge on the activiti-cloud project.

PR #5165, opened 2025-10-01
ACTIVITI-5557: on parallel multi instance tasks, loop and collections are evaluated when a task is closed (they should be evaluated only on task creation)

vrmoreira:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/ACTIVITI-5557

It's fixing a behaviour that was introduced in Activiti 6.x. Activiti 5.x only evaluates loops/collections when creating parallel tasks, while on Activiti 6.x it evaluates every time a task is completed.
Because of the behaviour change, it might break some expected behaviour. I would recommend to introduce a property to select which behaviour the customer wants, where the default is the behaviour implemented by this issue.

Let me know if this property should be implemented or not.

> - [X] Yes
> - [ ] No


activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/multiinstance/MultiInstanceTest.java
```diff
@@ -16,9 +16,13 @@
 package org.activiti.engine.test.bpmn.multiinstance;
 
 import static java.util.Arrays.asList;
+import static java.util.Collections.EMPTY_LIST;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singleton;
 import static java.util.Collections.singletonMap;
+import static org.activiti.engine.impl.bpmn.behavior.MultiInstanceActivityBehavior.NUMBER_OF_ACTIVE_INSTANCES;
```
vrmoreira, line 23: I haven't noticed that this class has `NR_ACTIVE_INSTANCES` constant that is a duplicate of the `MultiInstanceActivityBehavior`.
Is there a strong reason that, originally, `MultiInstanceActivityBehavior` constant are protected?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/ParallelMultiInstanceBehavior.java
```diff
@@ -112,15 +112,6 @@ protected int createInstances(DelegateExecution execution) {
      * Called when the wrapped {@link ActivityBehavior} calls the {@link AbstractBpmnActivityBehavior#leave(ActivityExecution)} method. Handles the completion of one of the parallel instances
      */
     public void leave(DelegateExecution execution) {
-        boolean zeroNrOfInstances = false;
-        if (resolveNrOfInstances(execution) == 0) {
-            // Empty collection, just leave.
-            zeroNrOfInstances = true;
-            removeLocalLoopVariable(execution, getCollectionElementIndexVariable());
-            super.leave(execution); // Plan the default leave
-            execution.setMultiInstanceRoot(false);
```
balsarori, line 121: I wonder how important are these lines? Don't we need to remove the local loop var and set multi instance to false on the last loop index? I see these executed later but is there a case where they might not be called?

vrmoreira: Instead of having the process done after the multi-instance tasks, I can add a new usertask and check on the tests if everything is ok. It gives a little more confidence.

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/multiinstance/MultiInstanceTest.java
balsarori: We either use the ones in the test class or remove them and just use the ones in MultiInstanceActivityBehavior class

balsarori (approved):
LGTM

PR #5226, opened 2025-12-05
AAE-39499 Establish catchAll error is processed after specific errors are exhausted

supratikdebnath:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link: https://hyland.atlassian.net/browse/AAE-39499
Due to the areas of change, also handled fix of https://hyland.atlassian.net/browse/AAE-41531

> - [ ] Yes
> - [x] No


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,65 +276,98 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
         String errorRef
     ) {
-        Map<String, List<Event>> eventMap = new HashMap<String, List<Event>>();
+        Map<String, List<Event>> eventMap = new HashMap<>();
         Process process = ProcessDefinitionUtil.getProcess(processDefinitionId);
         BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(processDefinitionId);
 
         String compareErrorCode = retrieveErrorCode(bpmnModel, errorRef);
 
+        // Process event sub-processes
```
erdemedeiros: This comment means that the following block can be extracted in a method with similar name. In this case you can extract `findCatchingEventSubprocesses`

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,65 +276,98 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
-                            eventMap.put(eventSubProcess.getId(), startEvents);
+                        if (isErrorCodeMatching(eventErrorCode, compareErrorCode)) {
+                            eventMap.put(eventSubProcess.getId(), Collections.singletonList(startEvent));
                         }
                     }
                 }
             }
         }
 
+        // Process boundary events
```
erdemedeiros: Same here, you can extract a method `findCatchingBoundaryEvents`

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,65 +276,98 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
         return eventMap;
     }
 
+    /**
+     * Helper method to add a boundary event to the event map.
+     * Uses computeIfAbsent for cleaner code (Java 8+).
+     *
+     * @param eventMap the map to add the event to
+     * @param boundaryEvent the boundary event to add
+     */
```
erdemedeiros: No need to maintain Javadoc for private methods

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,65 +276,98 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
+     * @param event the event to extract from
+     * @return the ErrorEventDefinition if found, null otherwise
+     */
+    private static ErrorEventDefinition getErrorEventDefinition(Event event) {
+        if (CollectionUtil.isNotEmpty(event.getEventDefinitions()) &&
+            event.getEventDefinitions().get(0) instanceof ErrorEventDefinition) {
+            return (ErrorEventDefinition) event.getEventDefinitions().get(0);
+        }
+        return null;
+    }
```
erdemedeiros: I think this method can be added directly to Event class (maybe returning an `Optional<ErrorEventDefinition>`), so it can also be reused in other places where we need to retrieve the error event definition from the event. Then we can simply call `event.getErrrorEventDefinition()`. WDTY?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,65 +276,98 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
+
+    /**
+     * Extracts the ErrorEventDefinition from an event's event definitions list.
+     *
+     * @param event the event to extract from
+     * @return the ErrorEventDefinition if found, null otherwise
+     */
+    private static ErrorEventDefinition getErrorEventDefinition(Event event) {
+        if (CollectionUtil.isNotEmpty(event.getEventDefinitions()) &&
+            event.getEventDefinitions().get(0) instanceof ErrorEventDefinition) {
```
erdemedeiros: Consider implementing the suggestion from sonar: "Replace this instanceof check and cast with 'instanceof ErrorEventDefinition erroreventdefinition'"

erdemedeiros (changes requested):
Here are the first comments

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Taken care of.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Taken care of.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Removed for the new additions.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Makes sense. Taken care of.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Taken care of.

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Event.java
```diff
@@ -51,4 +52,11 @@ public boolean isLinkEvent() {
         }
         return false;
     }
+
+    public Optional<ErrorEventDefinition> getErrorEventDefinition() {
+        return eventDefinitions != null && !eventDefinitions.isEmpty() &&
+               eventDefinitions.getFirst() instanceof ErrorEventDefinition errorEventDefinition
+            ? Optional.of(errorEventDefinition)
+            : Optional.empty();
+    }
```
erdemedeiros, line 61: Please add unit test coverage for this

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,63 +275,77 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
         String errorRef
     ) {
-        Map<String, List<Event>> eventMap = new HashMap<String, List<Event>>();
+        Map<String, List<Event>> eventMap = new HashMap<>();
         Process process = ProcessDefinitionUtil.getProcess(processDefinitionId);
         BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(processDefinitionId);
 
         String compareErrorCode = retrieveErrorCode(bpmnModel, errorRef);
 
+        findCatchingEventSubprocesses(eventMap, process, bpmnModel, compareErrorCode);
```
erdemedeiros: Instead of passing the map as parameter, prefer to instantiate it inside the the method and use it as return type. You can then aggregate the new entries to the exiting on by using `eventMap.putAll` method

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,63 +275,77 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
+
+        findCatchingBoundaryEvents(eventMap, process, bpmnModel, compareErrorCode);
```
erdemedeiros: Same here. Prefer to use return type

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,63 +275,77 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
+                            eventMap.put(eventSubProcess.getId(), Collections.singletonList(startEvent));
                         }
-                    }
+                    });
                 }
             }
         }
+    }
 
+    private static void findCatchingBoundaryEvents(
```
erdemedeiros: This could potentially also be moved to Process as a public method with independent code coverage, but it's out of the scope of this PR, so it's fining keeping it here.

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventProcessingOrderTest.java
```diff
@@ -0,0 +1,175 @@
...
+        // The specific NOT_FOUND handler should execute, not the catch-all
+        Task task = taskService.createTaskQuery().singleResult();
+        assertThat(task).isNotNull();
+        assertThat(task.getName()).isEqualTo("Not Found Handler Task");
+        assertThat(task.getTaskDefinitionKey()).isEqualTo("notFoundTask");
+
+        // Completing the task will end the process instance
+        taskService.complete(task.getId());
+        assertProcessEnded(procId);
+    }
```
erdemedeiros, line 131: I'm not sure this test case is needed. It sounds similar to the previous one, the only difference is that here it's catching the one in the middle and not the latest. Am I missing something?

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventProcessingOrderTest.java
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * BPMN Structure:
+     * - Subprocess throws error with code "123"
+     * - Boundary event with NO error code (catch-all) - defined FIRST in XML
+     * - Boundary event with error code "123" (specific) - defined SECOND in XML
+     *
+     * Expected: Specific error handler executes, NOT the catch-all
+     */
+    @Deployment
+    public void testSpecificErrorHandlerTakesPrecedenceOverCatchAllRegardlessOfXMLOrder() {
```
erdemedeiros: Isn't this scenario similar to the one defined in `testXMLOrderDoesNotAffectPrecedence`?

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventProcessingOrderTest.java
```diff
@@ -0,0 +1,175 @@
...
+     * - Subprocess throws error with code "ERROR_2"
+     * - Boundary event with error code "ERROR_1" (specific) - won't match
+     * - Boundary event with error code "ERROR_2" (specific) - SHOULD match
+     * - Boundary event with error code "ERROR_3" (specific) - won't match
+     * - Boundary event with NO error code (catch-all) - should not execute
+     *
+     * Expected: Only ERROR_2 handler executes
+     */
+    @Deployment
+    public void testOnlyMatchingSpecificHandlerExecutes() {
```
erdemedeiros: Isn't this scenario already covered by `testXMLOrderDoesNotAffectPrecedence`?

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+
+        assertThat(processInstance).isNotNull();
+
+        // Verify that ERROR_CODE_1 is caught by the specific handler (catchError1), not the catch-all
+        checkProcessAndTask(processInstance.getId(), "Task 1");
+
+        // Validate error events received
+        assertThat(listener.getErrorReceivedEvents())
+            .isNotEmpty()
+            .hasSize(1)
```
erdemedeiros: Both `isNotEmpty()` and `hasSize(1)` checks are redundant: `containsExactly` will already fail if the collection is empty or have a different size than 1.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+
+        assertThat(processInstance).isNotNull();
+
+        // Verify that the catch-all handler caught the unhandled error
+        checkProcessAndTask(processInstance.getId(), "Task Any");
+
+        // Validate that only the catch-all error event was received
+        assertThat(listener.getErrorReceivedEvents())
+            .isNotEmpty()
+            .hasSize(1)
```
erdemedeiros: Redundant

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+            )
+            .containsExactly(
+                Tuple.tuple("catchError1", "ERROR_CODE_1")
+            );
+
+        // Verify that the specific error handler was triggered, not the catch-all
+        assertThat(listener.getErrorReceivedEvents())
+            .extracting(event -> event.getEntity().getElementId())
+            .doesNotContain("catchErrorAny")
+            .contains("catchError1");
```
erdemedeiros: This is already covered by the previous assertion: if it contains `catchErrorAny` the assertion `containsExactly` at line `99` will fail

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+
+        // Additional validation: error code should not be null (proving it's not the catch-all)
+        assertThat(listener.getErrorReceivedEvents())
+            .allSatisfy(event -> {
+                assertThat(event.getEntity().getErrorCode())
+                    .as("Error code should not be null for specific error handlers")
+                    .isNotNull()
+                    .isEqualTo("ERROR_CODE_1");
+            });
```
erdemedeiros: Same for this one: it's already covered by the block `92-101`

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+
+        // Verify the catch-all handler was NOT executed
+        assertThat(listener.getErrorReceivedEvents())
+            .noneMatch(event -> event.getEntity().getElementId().equals("catchErrorAny"));
```
erdemedeiros: Same here

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+        // Verify that the catch-all handler was executed
+        assertThat(listener.getErrorReceivedEvents())
+            .extracting(event -> event.getEntity().getElementId())
+            .contains("catchErrorAny")
+            .doesNotContain("catchError1");
+
+        // Verify the specific handler (catchError1) was NOT executed
+        assertThat(listener.getErrorReceivedEvents())
+            .noneMatch(event -> event.getEntity().getElementId().equals("catchError1"));
+    }
```
erdemedeiros: Already covered by the previous assertions

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+                    .as("Only the specific error handler should be triggered")
+                    .isEqualTo("catchError1");
+                assertThat(event.getEntity().getErrorCode())
+                    .as("Error code should match the specific handler")
+                    .isEqualTo("ERROR_CODE_1");
+            });
+    }
+
+    @Test
+    public void should_NotExecuteMultipleBoundaryEvents_When_OneMatchesError() {
```
erdemedeiros: Isn't this covering the same scenario as the test above?

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
```diff
@@ -0,0 +1,221 @@
...
+            .extracting(
+                event -> event.getEntity().getElementId(),
+                event -> event.getEntity().getErrorCode()
+            )
+            .containsExactly(
+                Tuple.tuple("catchErrorAny", "UNHANDLED_ERROR")
+            );
+
+        // Verify that the catch-all handler was executed
+        assertThat(listener.getErrorReceivedEvents())
```
erdemedeiros: Already covered by the previous assertions

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
erdemedeiros: We could consider moving this method inside `Process` class as a public method with independent code coverage, but this is probably out of the scope of this PR, so it can be skipped.

erdemedeiros (changes requested):

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Event.java
supratikdebnath: Added coverage to a concrete implementation of Event.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Done

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Done

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventProcessingOrderTest.java
supratikdebnath: Removed. You are right, the test was redundant.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed redundant checks

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed redundant checks

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed redundant checks

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed redundant checks

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed redundant checks

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed redundant checks

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Yes, additional tests resulted to ensure specific negative cases mentioned in the ticket do not occur.

activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ErrorBoundaryEventProcessingOrderIT.java
supratikdebnath: Removed

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventProcessingOrderTest.java
supratikdebnath: Removed

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/error/BoundaryErrorEventProcessingOrderTest.java
supratikdebnath: Removed

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -276,65 +275,82 @@ protected static Map<String, List<Event>> findCatchingEventsForProcess(
...
+                        if (isErrorCodeMatching(eventErrorCode, compareErrorCode)) {
+                            eventMap.put(eventSubProcess.getId(), Collections.singletonList(startEvent));
                         }
-                    }
+                    });
                 }
             }
         }
+        return eventMap;
+    }
```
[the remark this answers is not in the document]
supratikdebnath: The suggestion is being explored.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Leaving it as is for this change.

supratikdebnath: Leaving it as is for this change.

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Event.java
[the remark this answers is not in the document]
supratikdebnath: Seems redundant and will only impact if StartEvent overrides the method. However implementing the suggestions seems a safer bet.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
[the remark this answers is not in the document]
supratikdebnath: Has been implemented.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -53,11 +52,11 @@ public static void propagateError(String errorRef, DelegateExecution execution)
         try {
             Map<String, List<Event>> eventMap = findCatchingEventsForProcess(
                 execution.getProcessDefinitionId(),
-                errorRef
-            );
+                errorRef,
+                null);
```
erdemedeiros: [nitpick]: wouldn't it be better to add specific constructors to `Error` class (in addition to the default one) and pass an error object as parameter instead of having `errorRef` and `errorCode`? `findCatchingEventsForProcess(execution.getProcessDefinitionId(), new Error(errorRef))`

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -107,7 +106,7 @@ protected static boolean executeCatch(
 
                 for (String refId : eventMap.keySet()) {
                     List<Event> events = eventMap.get(refId);
-                    if (CollectionUtil.isNotEmpty(events) && events.get(0) instanceof StartEvent) {
+                    if (matchingEvent == null && CollectionUtil.isNotEmpty(events) && events.get(0) instanceof StartEvent) {
```
erdemedeiros: Redundant null check?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -181,6 +181,24 @@ protected static boolean findCatchingEventsAndExecuteCatchForCallActivity(
...
+
+        ExecutionEntity executionEntity = (ExecutionEntity) execution;
+        BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(executionEntity.getProcessDefinitionId());
+
+        if (bpmnModel != null && bpmnModel.getErrors() != null) {
+            Error error = bpmnModel.getErrors().get(errorRef);
+            if (error != null) {
+                return error.getErrorCode();
+            }
+        }
```
erdemedeiros: [nitpick]: you can add a method `getErrorCode(errorRef)` directly in BpmnModel as it may also be useful in other places and it will avoid doing all this logic again.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -219,17 +237,18 @@ protected static void executeEventHandler(
...
         ) {
             BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(parentExecution.getProcessDefinitionId());
             if (bpmnModel != null) {
-                String errorCode = Optional.ofNullable(bpmnModel.getErrors().get(errorId))
-                    .map(Error::getErrorCode)
-                    .orElse(errorId);
+                String resolvedErrorCode = errorCode != null ? errorCode :
+                    Optional.ofNullable(bpmnModel.getErrors().get(errorId))
+                        .map(Error::getErrorCode)
+                        .orElse(errorId);
```
erdemedeiros: This, for instance, can reuse the method `getErrorCode` defined in BpmnModel mentioned above: `Optional.ofNullable(bpmnModel.getErrorCode(errorId)).orElse(errorId)`. BTW, this logic seems to be the same as calling `retrieveErrorCode` method. So I think you can simplify the logic of `retrieveErrorCode` as above and use it here.

erdemedeiros (commented):
I added some final recommendations. LGTM, otherwise

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Implemented.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: Implemented and also used pattern in executeCatch and another method call.

erdemedeiros (approved):

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -53,11 +52,10 @@ public static void propagateError(String errorRef, DelegateExecution execution)
         try {
             Map<String, List<Event>> eventMap = findCatchingEventsForProcess(
                 execution.getProcessDefinitionId(),
-                errorRef
-            );
+                new Error(errorRef));
 
             if (!eventMap.isEmpty()) {
-                isCatchExecutedForProcess = executeCatch(eventMap, execution, errorRef);
+                isCatchExecutedForProcess = executeCatch(eventMap, execution, new Error(errorRef));
```
souvikpanda99, line 58: [nitpick] instead of creating the Error object twice, we can create it once and reuse, which avoids a redundant allocation and makes the intent clearer

activiti-core/activiti-bpmn-model/src/test/java/org/activiti/bpmn/model/BoundaryEventTest.java
```diff
@@ -50,4 +53,72 @@ public void hasErrorEventDefinition_should_returnFalse_when_doesNotContainErrorE
...
+    @Test
+    public void getErrorEventDefinition_should_returnErrorEventDefinition_when_firstEventDefinitionIsErrorEvent() {
+        BoundaryEvent boundaryEvent = new BoundaryEvent();
+        ErrorEventDefinition errorEventDefinition = new ErrorEventDefinition();
+        boundaryEvent.setEventDefinitions(List.of(errorEventDefinition));
+
+        Optional<ErrorEventDefinition> result = boundaryEvent.getErrorEventDefinition();
+
+        assertThat(result).isPresent();
+        assertThat(result.get()).isSameAs(errorEventDefinition);
```
souvikpanda99: [nitpick] Use assertThat(actual).containsSame(expected) instead, see the sonarcloud recommendation here

activiti-core/activiti-bpmn-model/src/test/java/org/activiti/bpmn/model/BoundaryEventTest.java
```diff
@@ -50,4 +53,72 @@ public void hasErrorEventDefinition_should_returnFalse_when_doesNotContainErrorE
...
+    public void getErrorEventDefinition_should_returnErrorEventDefinition_when_firstOfMultipleEventDefinitionsIsErrorEvent() {
+        BoundaryEvent boundaryEvent = new BoundaryEvent();
+        ErrorEventDefinition errorEventDefinition = new ErrorEventDefinition();
+        MessageEventDefinition messageEventDefinition = new MessageEventDefinition();
+        boundaryEvent.setEventDefinitions(List.of(errorEventDefinition, messageEventDefinition));
+
+        Optional<ErrorEventDefinition> result = boundaryEvent.getErrorEventDefinition();
+
+        assertThat(result).isPresent();
+        assertThat(result.get()).isSameAs(errorEventDefinition);
```
souvikpanda99: [nitpick] Use assertThat(actual).containsSame(expected) instead, see the sonarcloud recommendation here

activiti-core/activiti-bpmn-model/src/test/java/org/activiti/bpmn/model/StartEventTest.java
```diff
@@ -0,0 +1,71 @@
...
+    public void getErrorEventDefinition_should_returnErrorEventDefinition_when_firstOfMultipleEventDefinitionsIsErrorEvent() {
+        StartEvent startEvent = new StartEvent();
+        ErrorEventDefinition errorEventDefinition = new ErrorEventDefinition();
+        MessageEventDefinition messageEventDefinition = new MessageEventDefinition();
+        startEvent.setEventDefinitions(List.of(errorEventDefinition, messageEventDefinition));
+
+        Optional<ErrorEventDefinition> result = startEvent.getErrorEventDefinition();
+
+        assertThat(result).isPresent();
+        assertThat(result.get()).isSameAs(errorEventDefinition);
```
souvikpanda99: [nitpick] Use assertThat(actual).containsSame(expected) instead, see the sonarcloud recommendation here

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -274,67 +275,108 @@ protected static void executeEventHandler(
...
+                }
+            }
+        }
+
+        for (EventSubProcess eventSubProcess : eventSubprocessesWithoutErrorCode) {
+            for (FlowElement flowElement : eventSubProcess.getFlowElements()) {
+                if (flowElement instanceof StartEvent startEvent) {
+                    startEvent.getErrorEventDefinition().ifPresent(errorEventDef -> {
+                        addEventSubprocessToMap(eventMap, eventSubProcess, startEvent);
+                    });
```
souvikpanda99, line 324: In findCatchingEventSubprocesses, the second pass seems to add all error start events for any event subprocess that has a catch‑all. If an event subprocess contains both a specific error start and a catch‑all start, could the non‑matching specific start get re‑added and picked first (XML order), preventing the catch‑all from firing? Would it be safer to only append start events with errorRef == null in that second pass?

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Event.java
```diff
@@ -51,4 +52,11 @@ public boolean isLinkEvent() {
         }
         return false;
     }
+
+    public Optional<ErrorEventDefinition> getErrorEventDefinition() {
+        return eventDefinitions != null && !eventDefinitions.isEmpty() &&
+               eventDefinitions.getFirst() instanceof ErrorEventDefinition errorEventDefinition
```
souvikpanda99, line 58: Is the assumption that an event will always have exactly one event definition? If multiple definitions are possible, later ErrorEventDefinitions would be ignored

souvikpanda99 (commented):
Looks good to me, added some minor comments

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: eventSubprocessesWithoutErrorCode is only populated if the eventErrorCode is null. So a non-matching errorCode will not be part of the second pass.

activiti-core/activiti-bpmn-model/src/main/java/org/activiti/bpmn/model/Event.java
supratikdebnath: Yes, an endEvent or a boundaryEvent should have only one error reason associated to the event.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
souvikpanda99: Basically, here I'm talking about the case where a single EventSubProcess has multiple start events, not sure if that likely or not. When the EventSubProcess is added (because it HAS a catch-all StartEvent), the second pass re-iterates all StartEvents in that subprocess. This includes both the catch-all and any specific StartEvents that exist in the same subprocess.

souvikpanda99 (approved):

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/helper/ErrorPropagation.java
supratikdebnath: As per the documentation, event sub-process can have only one start event.

PR #5299, opened 2026-02-17
Update to Jakarta Mail

andashyland:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe:

## Description
Upgarded apache.commons mail. Cherry picked ACTIVITI-4935 onto develop.

> - [ ] Yes
> - [x] No

### DIFF
--- a/activiti-core/activiti-engine/pom.xml
+++ b/activiti-core/activiti-engine/pom.xml
@@ -46,12 +46,12 @@
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-email2-jakarta</artifactId>
-      <version>2.0.0-M1-20240118.130716-5</version>
+      <version>2.0.0-M1</version>
       </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-email2-core</artifactId>
-      <version>2.0.0-M1-20240118.130716-4</version>
+      <version>2.0.0-M1</version>
      </dependency>
 <!-- jakarta.mail is transitive dependency commons-email2-jakarta,
       Hence jakarta.mail dependency can be removed from here, when commons-email2-jakarta is available in maven central -->
--- a/activiti-core/activiti-engine/pom.xml
+++ b/activiti-core/activiti-engine/pom.xml
@@ -45,7 +45,20 @@
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
-      <artifactId>commons-email</artifactId>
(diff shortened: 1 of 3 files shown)


activiti-core/activiti-engine/pom.xml
```diff
@@ -45,7 +45,20 @@
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
-      <artifactId>commons-email</artifactId>
+      <artifactId>commons-email2-jakarta</artifactId>
+      <version>2.0.0-M1</version>
+      </dependency>
+    <dependency>
```
vrmoreira: I hate to do this, changing on a cherry-pick. Up to you to do the change or not.
My comment is we should not include "commons-email2-core" and "jakarta.mail" artifacts, since they are fetched transitively by maven. No need to declare them.

vrmoreira (changes requested):

activiti-core/activiti-engine/pom.xml
```diff
@@ -45,8 +45,14 @@
...
     <dependency>
       <groupId>org.apache.commons</groupId>
-      <artifactId>commons-email</artifactId>
-    </dependency>
+      <artifactId>commons-email2-jakarta</artifactId>
+      <version>2.0.0-M1</version>
+      </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-email2-core</artifactId>
```
vrmoreira: Think you can remove this "commons-email2-core" dependency as well.
it's already declared as a transitive dependency here: https://github.com/apache/commons-email/blob/rel/commons-email-2.0.0-M1/commons-email2-jakarta/pom.xml#L36

vrmoreira (changes requested):

activiti-core/activiti-engine/pom.xml
andashyland: ok. done. What do you think we should do to the original branch 8.1.x? Create a new PR for the original card or create a new PR on a new card or may be leave it for now. Next question will be should we hard-code the version here or actually put a maven property in the pom file of activiti-parent? I think we should do that. This was not done in original PR.

vrmoreira: > What do you think we should do to the original branch 8.1.x? 
That's why I hate to change on cherry-picks: all changes should be propagated to the original commit as well.

As a best-practice, you're right, by declaring the version as a property will help to maintain the software (and, possibly, migrate to a BOM approach)

PR #5450, opened 2026-06-23
MNT-24851: Add support for signal-triggered event sub-processes

andashyland:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description
Add support for signal-triggered event sub-processes
https://hyland.atlassian.net/wiki/spaces/~andas/pages/4152918213/MNT-24851+APS+-+Event+SubProcess+are+not+working

- Issue Link: https://hyland.atlassian.net/browse/MNT-24851

> - [ ] Yes
> - [x] No


activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/AbstractEventSubProcessStartEventActivityBehavior.java
andashyland, line 1: Creation of this class was not strictly necessary. This was only done to make sonarqube happy. More then 3% code repetition is not allowed and PR gets blocked by sonarqube. I still think the refactoring was not necessary and just copying from EventSubProcessMessageStartEventActivityBehavior class to create EventSubProcessSignalStartEventActivityBehavior with little tweaks was good enough. This has unnecessarily made review and testing difficult. People may feel like the existing subprocess message start event has been modified and needs to be tested thoroughly.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/AbstractEventSubProcessStartEventActivityBehavior.java
```diff
@@ -0,0 +1,161 @@
...
+
+        // initialize the template-defined data objects as variables
+        Map<String, Object> dataObjectVars = processDataObjects(eventSubProcess.getDataObjects());
+        if (dataObjectVars != null) {
+            execution.setVariablesLocal(dataObjectVars);
+        }
+    }
+
+    @Override
+    public void trigger(DelegateExecution execution, String triggerName, Object triggerData) {
```
astrologic7, line 65: should we take into account the **\<dataObject\>** tag here as `execute()` does? If that's the case we should also establish what to do in case of variables clash.
e.g.

```java
  private void initializeDataObjects(DelegateExecution scopeExecution, StartEvent startEvent) {
      EventSubProcess eventSubProcess = (EventSubProcess) startEvent.getSubProcess();
      Map<String, Object> dataObjectVars = processDataObjects(eventSubProcess.getDataObjects());
      if (dataObjectVars != null) {
          dataObjectVars.forEach((name, value) -> {
              if (!scopeExecution.hasVariable(name)) {
                  scopeExecution.setVariableLocal(name, value);
              }
          });
      }
  }

```

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/parser/handler/StartEventParseHandler.java
```diff
@@ -64,6 +66,17 @@ protected void executeParse(BpmnParse bpmnParse, StartEvent element) {
                             .getActivityBehaviorFactory()
                             .createEventSubProcessErrorStartEventActivityBehavior(element)
                     );
+                } else if (eventDefinition instanceof SignalEventDefinition signalDefinition) {
+                    BpmnModel bpmnModel = bpmnParse.getBpmnModel();
+                    Signal signal = null;
+                    if (bpmnModel.containsSignalId(signalDefinition.getSignalRef())) {
```
astrologic7, line 72: should we check that `getISignalRef()` doesn't return null? Looking at the code flow looks like we could get NPE

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/ProcessInstanceHelper.java
```diff
@@ -329,6 +331,25 @@ public void startProcessInstance(
                                 correlationKey.ifPresent(subscription::setConfiguration);
 
                                 messageEventSubscriptions.add(subscription);
+                            } else if (eventDefinition instanceof SignalEventDefinition signalEventDefinition) {
+                                BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(
+                                    processInstance.getProcessDefinitionId()
+                                );
+                                Signal signal = null;
+                                String signalName = signalEventDefinition.getSignalRef();
+                                if (bpmnModel.containsSignalId(signalEventDefinition.getSignalRef())) {
```
astrologic7, line 340: should we check that `getISignalRef()` doesn't return null? Looking at the code flow looks like we could get NPE

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/ProcessInstanceHelper.java
```diff
@@ -329,6 +331,25 @@ public void startProcessInstance(
...
+                                    signal = bpmnModel.getSignal(signalEventDefinition.getSignalRef());
+                                    signalName = signal.getName();
```
astrologic7, line 342: is `signalEventDefinition.getSignalExpression()` ignored on purpose here?

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/SignalEventSubprocessTest.java
```diff
@@ -0,0 +1,252 @@
...
+            .signalEventSubscriptionName(SIGNAL_NAME)
+            .singleResult();
+        assertThat(execution).isNotNull();
+        runtimeService.signalEventReceived(SIGNAL_NAME, execution.getId());
+
+        // interrupting => only the event sub-process task remains
+        Task task = taskService.createTaskQuery().singleResult();
+        assertThat(task.getTaskDefinitionKey()).isEqualTo("eventSubProcessTask");
+
+        /* TODO: execute() is never being called. Why? Is it a bug? Whatever this is, will deal in a separate card.
```
astrologic7: looks like here we call `trigger()` not `execute()` and there we don't evaluate **\<dataObject\>** see the related comment

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/SignalEventSubprocessTest.java
```diff
@@ -0,0 +1,252 @@
...
+
+    /**
+     * Verifies that a non-interrupting signal-triggered event sub-process can fire more than
+     * once for the same process instance, i.e. that the subscription is re-created after each
+     * trigger.
+     */
+    @Deployment(
+        resources = "org/activiti/engine/test/bpmn/event/signal/SignalEventSubprocessTest.testNonInterruptingUnderProcessDefinition.bpmn20.xml"
+    )
+    public void testNonInterruptingCanTriggerMultipleTimes() {
```
astrologic7, line 161: completing `eventSubProcessTask` (and the subprocess) itself breaks the subscription's renewal mechanism, is it expected?

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/ProcessInstanceHelper.java
andashyland: ok. Done.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/AbstractEventSubProcessStartEventActivityBehavior.java
andashyland: Ok. Done. Only minor side-effect is that changing this also changes the behaviour EventSubProcessMessageStartEventActivityBehavior. But I think that should be ok.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/util/ProcessInstanceHelper.java
andashyland: I wanted to keep the implementation simple with minimalistic change to the existing behaviour. Supporting EL for signalEventDefinition.getSignalExpression() makes it a bit more complex because of how I have implemented Event sub-process specificly non-interrupting signal event sub-processes. I am basically re-creating the subscription once consumed. If the name is expression-based and a referenced variable changed between fires, we must re-evaluate and the name could drift easily introducing a problem. So, I would have to cache the resolved name at first subscription and reuse it on re-arm, so a non-interrupting sub-process keeps listening for the same signal it originally subscribed to. One more thing is that I think a uniform signal-name resolver should be used across all signal classes (like IntermediateThrowSignalEventActivityBehavior, IntermediateCatchSignalEventActivityBehavior, BoundarySignalEventActivityBehavior, EventSubProcessSignalStartEventActivityBehavior, etc). Presently signal-name resolving is inconsistent, but it works probably because our customers are using id and name as same something like ```<signal id="X" name="X">``` . This was extra complexity and I have to touch many existing files, that I wanted to avoid.
I would try to add expression support now, if you feel strongly we need it in this PR.

activiti-core/activiti-engine/src/main/java/org/activiti/engine/impl/bpmn/parser/handler/StartEventParseHandler.java
andashyland: ok. Done.

activiti-core/activiti-engine/src/test/java/org/activiti/engine/test/bpmn/event/signal/SignalEventSubprocessTest.java
andashyland: I have updated EndExecutionOperation, but have commented it out, for you to review. If you think we should make that change then I will uncomment it out.

PR #5537, opened 2026-09-08
AAE-51108 Configure SetVariablesTask as asynchronous

wojciech-piotrowiak:
… coverage

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link:

> - [X] Yes
> - [ ] No


mauriziovitale (approved):

wojciech-piotrowiak:
@copilot Have a problem when this task fails and throw error in calculator. Then transaction is probably rollbacked which might cause situation when whole process instance is still runnng and task is NOT marked as failed. My Friend suggested me this, but not sure how to proceed: the behavior should not throw an exception, otherwise the transaction will be rollbacked
in addition of not throwing an exception, in order to have the task marked as error runtime bundle needs to send an integrationErrorReceived event to query service
From what I see on your changes you don't have any of these two points above. I think you'll need a new type of event instead of relying on existing incident

PR #5544, opened 2026-09-13
AAE-51778 Implement preview propagation mechanism

wojciech-piotrowiak:
Mechanism that allow to automaticaly test propagation between activiti and activiti-cloud

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

## Description

- Issue Link:

> - [ ] Yes
> - [x] No


.jx/updatebot-preview.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: updatebot.jenkins-x.io/v1alpha1
+kind: UpdateConfig
+spec:
+  rules:
+    - urls:
+        - https://github.com/Activiti/activiti-cloud
+      reusePullRequest: false
```
Giovanni007: is it correct to have reuse false?

wojciech-piotrowiak: I used that when preview version was dynamic, but once it is fixed I can use true, thanks!

.github/workflows/main_pull.yml
```diff
@@ -120,3 +123,55 @@ jobs:
...
+          version: ${{ env.VERSION }}
+          auto-merge: 'false'
+          labels: 'do-not-merge,preview-source-pr-${{ github.event.pull_request.number }}'
+          base-branch-name: ${{ env.DEVELOPMENT_BRANCH }}
+          git-username: ${{ vars.HXPS_GIT_USERNAME }}
+          git-token: ${{ secrets.BOT_GITHUB_TOKEN }}
+          git-author-name: ${{ vars.HXPS_GIT_USERNAME }}
+          git-author-email: ${{ vars.HXPS_GIT_EMAIL }}
+          gpg-private-key: ${{ secrets.HXPS_GIT_COMMIT_SIGNING_PRIVATE_KEY }}
+          gpg-private-key-fingerprint: ${{ secrets.HXPS_GIT_COMMIT_SIGNING_FINGERPRINT }}
```
Giovanni007, line 157: This is not strictly related to this PR, since the same credentials were already used by the existing propagation workflow. However, we should confirm that the token is limited to `activiti-cloud` if we don't need to access to other repositories

Giovanni007 (approved):

PR #5554, opened 2026-09-18
AAE-52162 Skip parsing huge Strings for expressions

andrea-ligios:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe:

## Description

The runtime-bundle service was experiencing OutOfMemoryError crashes during expression resolution when processing integration results with large string variables. The issue occurred in Activiti's `ExpressionResolver.resolveExpressionsString()` method at line 96.

> [!WARNING]
When a large string (e.g., 1MB+) contained many `$`, `{`, and `}` characters, the regex engine would exhaust memory trying to match the entire string, leading to OutOfMemoryError.

While big variables are legit, expressions in such big variables are likely not, therefore we're going to add the possibility to introduce a var-size limit, beyond which parsing expressions is not performed.

If the limit is not defined, it keeps working as is, to avoid breaking the retrocompatibility.

> [!IMPORTANT]
The limit is defined in bytes in the var `MAX_VAR_SIZE_FOR_EXPRESSION_PARSING`.

> [!NOTE]
Furthermore, Copilot has rewritten the parsing feature to use a linear parser rather than a regex, to get better performances overall.

- Issue Link: https://hyland.atlassian.net/browse/AAE-52162

> - [ ] Yes
> - [x] No


eromano (approved):

activiti-core/activiti-api-impl/activiti-api-process-runtime-impl/src/main/java/org/activiti/runtime/api/impl/ExpressionResolver.java
```diff
@@ -36,25 +36,84 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ExpressionResolver.class);
 
-    private static final String EXPRESSION_PATTERN_STRING = "([\\$]\\{([^\\}]*)\\})";
-    private static final Pattern EXPRESSION_PATTERN = Pattern.compile(EXPRESSION_PATTERN_STRING);
-    private static final int EXPRESSION_KEY_INDEX = 1;
+    private static final String EXPRESSION_PREFIX = "${";
+    private static final int DEFAULT_MAX_VAR_SIZE_FOR_EXPRESSION_PARSING = Integer.MAX_VALUE;
+    private static final int MAX_VAR_SIZE_FOR_EXPRESSION_PARSING = resolveMaxVarSizeForExpressionParsing(
+        System.getenv("MAX_VAR_SIZE_FOR_EXPRESSION_PARSING")
+    );
```
erdemedeiros, line 45: Can we bind this through Spring configuration instead of reading the environment variable directly here?

`ExpressionResolver` is already constructed as a bean in `ConnectorsAutoConfiguration`, so the limit can simply be a constructor argument and the class itself stays Spring-free:

```java
@Bean
@ConditionalOnMissingBean
public ExpressionResolver expressionResolver(
    ExpressionManager expressionManager,
    JsonMapper jsonMapper,
    @Value("${activiti.expression-resolver.max-var-size-for-parsing:" + DEFAULT_MAX_VAR_SIZE + "}")
    int maxVarSizeForExpressionParsing
) {
    return new ExpressionResolver(
        expressionManager,
        jsonMapper,
        new DefaultDelegateInterceptor(),
        maxVarSizeForExpressionParsing
    );
}
```

andrea-ligios:
Closing since the root cause has been found and fixed in https://github.com/Activiti/Activiti/pull/5558

